PR #3, opened 2023-05-15
design

### LINKED ISSUE #1: design document
Create a design document in this repository that describes what twoliter is, what its requirements are, and how we think it can be built.

webern:
Adds a design doc describing the Twoliter build tool.

*Issue #, if available:*

Closes #1

*Description of changes:*

Adds a document describing the design of twoliter.

*TODO*

- [x] Discuss Opt-out features.
- [x] `dnf list installed` output for `https://github.com/bottlerocket-os/twoliter/pull/3/files#r1195749281`
- [x] Add a diagram to the Twoliter Build section `https://github.com/bottlerocket-os/twoliter/pull/3/files#r1195553253`
- [x] Investigate and write about using Cargo to do dependency management of Kits and the required SDK. (see #4)
- [x] Kits need to declare the sdk they were built with
- [x] All the kits in a build need to declare the same sdk
  - [x] Use cargo or cargo-like mechanisms to enforce (see #4)
- [x] Use immutable tags for kit metadata
- [x] Fix references to bootconfig (it should be produced by the SDK)
- [x] Discuss kit target architectures. `https://github.com/bottlerocket-os/twoliter/pull/3/files/#r1195738646`
- [x] Solve the signing requirements (or backlog the issue) `https://github.com/bottlerocket-os/twoliter/pull/3/files/#r1195739587` (See #5)


docs/design/README.md
```diff
@@ -0,0 +1,551 @@
+# Twoliter Design Document
+
+## Introduction
+
+Twoliter is a command line tool for building, testing and deploying custom variants of [Bottlerocket](https://github.com/bottlerocket-os/bottlerocket).
```
stockholmux: nit. 
Use oxford command between 'testing' & 'and'

docs/design/README.md
```diff
@@ -0,0 +1,551 @@
...
+Integration with the new settings system is out of scope for this document, and will be integrated with settings as both projects mature.
+
+### Requirements
+
+The `twoliter` build tool should meet the following requirements.
+
+1. Binaries and metadata published for consumption by out-of-tree builds must be signed and automatically verified.
+2. Out-of-tree builds must support generating license attribution documents.
+3. All functionality necessary to build, test, and publish out-of-tree builds must be provided by a unified CLI.
+4. Maintainers must be able to define their own packages.
```
stockholmux, line 44: Can you define 'maintainers' somewhere? I'm a bit confused if this means maintainer of the out of tree build or maintainers of bottlerocket. 

Perhaps use a different term would be better? In the glossary you use 'builder' is that the same meaning?

docs/design/README.md
```diff
@@ -0,0 +1,551 @@
...
+11. Maintainers must be free to choose their own version numbering scheme.
+12. Maintainers must have a simple mechanism that can be used to update Bottlerocket dependencies to newer versions.
+13. Maintainers must be able to host their own copies of any required container images or source archives.
+14. Maintainers must be able to specify their own partition layouts (thank you @jfbette)
+
+### Glossary
+
+Here are a few of the terms we will use in this document:
+
+- **In Tree**: Anything that exists in, or is defined by, the [main Bottlerocket git repository]
```
stockholmux: nit: 'in tree' is used here but 'in-tree' (hyphenated) is used on line 14. I don't have grammatical opinion, but it should be consistent.

docs/design/README.md
```diff
@@ -0,0 +1,551 @@
...
+- **Out of Tree Build (OOTB)**: User-customized builds of Bottlerocket defined by an external git repository owned by the user.
```
stockholmux: nit: inconsistent hyphenation - no hypen here, but on line 30 it has a hyphen.

docs/design/README.md
```diff
@@ -0,0 +1,551 @@
...
+It is in the SDK environment that the package builds take place.
+
+Once all the software for a Bottlerocket image has been built into RPM images, a final image creation step occurs.
+A `docker build` command installs everything into a rooted image file.
+
+The toolchains necessary for each arch (x86_64 and Arm) are available as separate SDK containers to allow for cross-compilation.
+The build host has the following requirements: sufficient versions of Docker, buildx, Cargo, Cargo Make and a few tools like lz4 and jq.
+The root of the Bottlerocket git repo is mounted and artifacts are created in `.cargo` and `build` directories.
+
+Certain custom tools for building, testing and publishing Bottlerocket (buildsys, testsys, and pubsys) are compiled from source during a Bottlerocket build.
```
stockholmux: nit: should you link buildsys, testsys and pubsys ?

docs/design/README.md
```diff
@@ -0,0 +1,551 @@
...
+* Kit: The `hello-dev-kit` depends on the `hello-agent` package by placing it in the `Cargo.toml` `build-dependencies` section.
+* Package: The `hello-agent` package.
+
+The user kicks things off with:
+
+```sh
+twoliter build --variant my-variant --arch aarch64
+```
+
+First Twoliter learns the external kit dependency graph using the traversal described for `twoliter update`.
```
stockholmux, line 475: The sequence described below is probably as clear as you can make it, but I think writing out the sequence is just challenging to follow. Adding simple diagram here would make it much easier to follow.

docs/design/README.md
```diff
@@ -0,0 +1,551 @@
...
+If a package is available in more than one kit, it will be taken from the higher priority kit by dnf.
+The list of desired packages, taken from the variant's `Cargo.toml` metadata (as it is now), will be installed along with any of their dependencies.
+
+After installing packages, the rest of the variant image creation steps proceed the same way that they do now.
+
+## Bottlerocket Forks
+
+Iterating on the build tools (like buildsys), `Makefile.toml`, docker files and scripts will be difficult unless Twoliter can take these things from source.
+Additionally, some builders will need to make changes in the Bottlerocket sourcetree.
+These use-cases lead to requirement 9.
```
stockholmux: nit: parenthetically add requirement 9 here or link somehow. Making the reader go back and hunt for requirement 9 is not very friendly.

docs/design/README.md
```diff
@@ -0,0 +1,551 @@
...
+# or an empty string.
+[sdk-image]
+registry = "public.ecr.aws/bottlerocket"
+name = "bottlerocket-sdk"
+version = "v0.30.0"
+
+# This is the container that Twoliter uses to invoke `cargo make` and other build
+# commands. It will be available for both x86_64 and Arm hosts, but it is agnostic
+# as to which architecture the build is targeting.
+[sodapop-image]
```
stockholmux: 'sodapop-image'?

docs/design/README.md
```diff
@@ -0,0 +1,551 @@
...
+
+An additional problem to solve is that the paths to mounted directories need to be absolutely the same on the host as they are in the Twoliter container.
+This will not be problem unless the host build or sourcecode directory is in some wildly unusual place, like `/etc/docker` or `/usr/bin`.
+We can prevent this by making a certain set of top-level directories off-limits on the host for the directories we need to mount.
+An error message will describe the problem instead of running the container and failing due to occluded directories.
+
+## Twoliter Projects
+
+Twoliter will have the concept of a project.
+A project can contain multiple variant definitions, multiple kit definitions, and multiple packages.
```
stockholmux, line 319: can you give an simple example of what a twoliter project might be used for? Multiple variants in a single project has thrown me a bit. I'm guessing you might want to have, say, two different architectures in a project, but it's not super clear.

stockholmux (commented):
This is really great. Twoliter's purpose is nice and crisp. I have a few minor nits and comments inline. 

General thoughts:
- The document switches between being written as a narrative and other times it uses specification wording ("Maintainers must...", "Twoliter will..."). Personally, I don't know if the specification wording is helpful, regardless, it might be good to go in a single direction for the entire document or at least make a section internally consistent.
- A section that describes an end-to-end concrete example would be helpful. Describe a kit, show how it's used in a twoliter project, how you build it, etc.

stockholmux (commented):
This is really great. Twoliter's purpose is nice and crisp. I have a few minor nits and comments inline. 

General thoughts:
- The document switches between being written as a narrative and other times it uses specification wording ("Maintainers must...", "Twoliter will..."). Personally, I don't know if the specification wording is helpful, regardless, it might be good to go in a single direction for the entire document or at least make a section internally consistent.
- A section that describes an end-to-end concrete example would be helpful. Describe a kit, show how it's used in a twoliter project, how you build it, etc.

docs/design/README.md
webern: I fixed this by changing all `builders` to `maintainers` and adding `maintainer` to the glossary.

docs/design/README.md
webern: attempted to fix, though unsure of capitalization around hyphenation.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
+# Twoliter Design Document
+
+## Introduction
+
+Twoliter is a command line tool for building, testing and deploying custom variants of [Bottlerocket](https://github.com/bottlerocket-os/bottlerocket).
+
+### Background
+
+One of Bottlerocket's security features is its immutable root filesystem and lack of package manager.
+Because software cannot be installed on Bottlerocket after the image has been created, the build system supports creating different Bottlerocket *variants*.
```
bcressey:
```suggestion
Because software cannot be installed on Bottlerocket after the image has been created, it lacks a traditional package manager.
Instead, the build system supports creating different Bottlerocket *variants*.
```

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
+# Twoliter Design Document
+
+## Introduction
+
+Twoliter is a command line tool for building, testing and deploying custom variants of [Bottlerocket](https://github.com/bottlerocket-os/bottlerocket).
+
+### Background
+
+One of Bottlerocket's security features is its immutable root filesystem and lack of package manager.
```
bcressey:
```suggestion
One of Bottlerocket's security features is its immutable root filesystem.
```

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+
+### Requirements
+
+The `twoliter` build tool should meet the following requirements.
+
+1. Binaries and metadata published for consumption by out-of-tree builds must be signed and automatically verified.
+2. Out-of-tree builds must support generating license attribution documents.
+3. All functionality necessary to build, test, and publish out-of-tree builds must be provided by a unified CLI.
+4. Maintainers must be able to define their own packages.
+5. Maintainers must be able to mix and match core project packages with their own.
```
jpmcb, line 45: Does "mix and match" here mean that someone could modify the core components of a upstream, in-tree package? I.e., add their own git patches, modify the patches we have, re-define how the rpm works?

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+│   ├── Cargo.toml
+│   └── lib.rs
+└── vmware
+    ├── build.rs
+    ├── Cargo.toml
+    └── lib.rs
+```
+
+A variant such as `metal-k8s-1.22` would consume the necessary kits.
+For example, it would reference `core-kit`, `k8s-kit` and `metal-kit`.
```
ecpullen, line 172: does `x-kit` just mean `kits/x`? Or are these separate kits?

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+Integration with the new settings system is out of scope for this document, and will be integrated with settings as both projects mature.
+
+### Glossary
+
+Here are a few of the terms we will use in this document:
+
+- **In-Tree**: Anything that exists in, or is defined by, the [main Bottlerocket git repository]
+- **Out-of-Tree Build (OOTB)**: User-customized builds of Bottlerocket defined by an external git repository owned by the user.
+- **Out-of-Tree Variants**: Synonymous with *Out of Tree Builds*.
+- **Builder**: A person who creates Bottlerocket variants using Twoliter.
```
webern: My find and replace was not comprehensive. Anywhere we see the word Builder it needs to be changed to Maintainer.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+
+One of Bottlerocket's security features is its immutable root filesystem and lack of package manager.
+Because software cannot be installed on Bottlerocket after the image has been created, the build system supports creating different Bottlerocket *variants*.
+Variants allow for different software and settings to be defined in Bottlerocket at build time.
+
+Currently, in order to define a variant, it needs to be added to the `variants` directory of the Bottlerocket git repository.
+We call this *in-tree*, meaning that a variant can only be defined within Bottlerocket's git tree.
+Similarly, to add software to Bottlerocket, you need add it to the in-tree `packages` directory.
+
+A fundamental intention with variants is to allow for users to customize Bottlerocket for their unique use cases.
```
bcressey:
```suggestion
The fundamental goal of variants is to support customizing Bottlerocket for different use cases.
```

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+
+## Introduction
+
+Twoliter is a command line tool for building, testing and deploying custom variants of [Bottlerocket](https://github.com/bottlerocket-os/bottlerocket).
+
+### Background
+
+One of Bottlerocket's security features is its immutable root filesystem and lack of package manager.
+Because software cannot be installed on Bottlerocket after the image has been created, the build system supports creating different Bottlerocket *variants*.
+Variants allow for different software and settings to be defined in Bottlerocket at build time.
```
bcressey:
```suggestion
Variants allow for different software and settings to be defined in Bottlerocket at build time.
This improves security by ensuring that only the minimum software is installed to support a specific use case.
```

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+
+Currently, in order to define a variant, it needs to be added to the `variants` directory of the Bottlerocket git repository.
+We call this *in-tree*, meaning that a variant can only be defined within Bottlerocket's git tree.
+Similarly, to add software to Bottlerocket, you need add it to the in-tree `packages` directory.
+
+A fundamental intention with variants is to allow for users to customize Bottlerocket for their unique use cases.
+To do so currently, however, requires forking and altering the Bottlerocket git repo.
+To stay up-to-date with core project would require frequent rebasing with conflicting changes in the core project.
```
bcressey:
```suggestion
To stay up-to-date with core project would require frequent rebasing with high potential for conflicting changes in the core project.
```

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+One of Bottlerocket's security features is its immutable root filesystem and lack of package manager.
+Because software cannot be installed on Bottlerocket after the image has been created, the build system supports creating different Bottlerocket *variants*.
+Variants allow for different software and settings to be defined in Bottlerocket at build time.
+
+Currently, in order to define a variant, it needs to be added to the `variants` directory of the Bottlerocket git repository.
+We call this *in-tree*, meaning that a variant can only be defined within Bottlerocket's git tree.
+Similarly, to add software to Bottlerocket, you need add it to the in-tree `packages` directory.
+
+A fundamental intention with variants is to allow for users to customize Bottlerocket for their unique use cases.
+To do so currently, however, requires forking and altering the Bottlerocket git repo.
```
bcressey: nit:
```suggestion
However, doing so currently requires forking and altering the Bottlerocket git repo.
```

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+        ├── bottlerocket-x86_64-apiclient-0.0-0.x86_64.rpm
+        ├── ...
+        └── bottlerocket-x86_64-wicked-0.6.68-1.x86_64.rpm
+```
+
+The file at `/local/etc/yum.repos.d` can be placed at `/etc/yum.repos.d` later when multiple kits are being aggregated.
+This will alow `dnf` to treat each kit as a yum repository.
+When aggregating kits, the yum repos will be prioritized according to the maintainer's requirements.
+
+The presence of `bootconfig` is questionable.
```
jpmcb: Will different out of tree builds have different boot configs? Maybe it make sense to have the bootconfig as it's own kit if we anticipate out of tree builders to modify/make their own bootconfigs.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+
+## Containers
+
+In order to minimize the amount of software required on the maintainer's system, Twoliter will use containers to drive the builds.
+There are two containers that will be published to facilitate the build system.
+
+### Twoliter Container
+
+Twoliter will use this container to bootstrap the build process.
+It needs to contain items the existing Bottlerocket build system expects to be available on the host.
```
webern: Broken sentence.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+The file at `/local/etc/yum.repos.d` can be placed at `/etc/yum.repos.d` later when multiple kits are being aggregated.
+This will alow `dnf` to treat each kit as a yum repository.
+When aggregating kits, the yum repos will be prioritized according to the maintainer's requirements.
+
+The presence of `bootconfig` is questionable.
+These files may need to be moved to the SDK or somewhere else.
+
+The presence of the `migrations` directory is temporary and controversial.
+Only the core project can have migrations, and we expect the migration system to go away.
+Ideally the presence of these will be completely eliminated as OOTB reaches maturity.
```
jpmcb: I'd like to hear more on this: are migrations going away because of the new settings mechanism that are coming (mentioned earlier)?

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+5. Maintainers must be able to mix and match core project packages with their own.
+6. Maintainers must be able to replace a core project package with their own.
+7. Maintainers must be able to use core project packages without building them.
+8. Maintainers must be able to take and use packages from multiple sources.
+9. Maintainers must be able to fork and patch the core project in lieu of using the published binary artifacts.
+10. Maintainers must bring their own keys for all signing operations.
+11. Maintainers must be free to choose their own version numbering scheme.
+12. Maintainers must have a simple mechanism that can be used to update Bottlerocket dependencies to newer versions.
+13. Maintainers must be able to host their own copies of any required container images or source archives.
+14. Maintainers must be able to specify their own partition layouts (thank you @jfbette)
```
bcressey: It breaks up the flow a little to credit individuals in-line, but adding a "credits" section to the end for posterity seems fine.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+Packages are defined and built using RPM.
+
+These package builds take place in a docker environment using `docker build` commands and `buildx`.
+The toolchains used to compile Rust, C, C++ and Go software are distributed in a container image called the [bottlerocket-sdk].
+It is in the SDK environment that the package builds take place.
+
+Once all the software for a Bottlerocket image has been built into RPM images, a final image creation step occurs.
+A `docker build` command installs everything into a rooted image file.
+
+The toolchains necessary for each arch (x86_64 and Arm) are available as separate SDK containers to allow for cross-compilation.
```
bcressey:
```suggestion
The toolchains necessary for each arch (x86_64 and aarch64) are available as separate SDK containers to allow for cross-compilation.
```

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+      ],
+      "v0.2.0": [
+        "public.ecr.aws/bottlerocket/bottlerocket-core-kit:v1.16.0"
+      ]
+    }
+  }
+}
+```
+
+This metadata can be stored and updated in the kit container registry with a mutable tag so that `twoliter` can easily find it.
```
jpmcb: > This metadata can be stored and updated in the kit container registry with a mutable tag

An image in a mutable tag could easily be modified by a malicious actor and force those malicious bits onto any registry puller who is targeting `metadata`. This is why `latest` is strongly discouraged from being used for kubernetes images.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+
+These package builds take place in a docker environment using `docker build` commands and `buildx`.
+The toolchains used to compile Rust, C, C++ and Go software are distributed in a container image called the [bottlerocket-sdk].
+It is in the SDK environment that the package builds take place.
+
+Once all the software for a Bottlerocket image has been built into RPM images, a final image creation step occurs.
+A `docker build` command installs everything into a rooted image file.
+
+The toolchains necessary for each arch (x86_64 and Arm) are available as separate SDK containers to allow for cross-compilation.
+The build host has the following requirements: sufficient versions of Docker, buildx, Cargo, Cargo Make and a few tools like lz4 and jq.
```
bcressey: Do we really require `jq`?

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+        │   ├── fcd4dadc59c6d6512410b103a38762bf266389e01a60add12319191c4cc16ec4-other.xml.gz
+        │   └── repomd.xml
+        ├── bottlerocket-x86_64-acpid-2.0.34-1.x86_64.rpm
+        ├── bottlerocket-x86_64-apiclient-0.0-0.x86_64.rpm
+        ├── ...
+        └── bottlerocket-x86_64-wicked-0.6.68-1.x86_64.rpm
+```
+
+The file at `/local/etc/yum.repos.d` can be placed at `/etc/yum.repos.d` later when multiple kits are being aggregated.
+This will alow `dnf` to treat each kit as a yum repository.
```
arnaldo2792: Nit: typo `alow`

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+When `twoliter` commands are invoked, the tool will search for the `Twoliter.toml` file in `.`.
+If it doesn't find it, it will search in `..`, then `../..`, etc.
+This will allow the tool to work from any subdirectory within the project (and is inspired by `cargo`).
+
+`Twoliter.toml` will contain global project settings. Here is a sample `Twoliter.toml` file.
+
+```toml
+# This let's twoliter know if it is reading an older version of the Twoliter.toml
+# schema, or if the directory structure of the project is from an older version.
+twoliter-project-version = 1
```
webern: I'm struggling with what to call this as it conflates with `project-version` below.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+[bottlerocket-sdk]: https://github.com/bottlerocket-os/bottlerocket-sdk
+
+### Enhancing the Build System
+
+The existing build system will remain intact with certain features and abstractions added to it.
+Twoliter will serve as a facade providing a new interface over the existing build system.
+Twoliter will understand how to start a Bottlerocket build using the OOTB user's sources, packages etc.
+It will inject these directories into the existing build system by setting existing and new `Makefile.toml` variables.
+
+Twoliter will eliminate the need for all build-host software other than `docker.`
```
bcressey: nit:
```suggestion
Twoliter will eliminate the need for all build-host software other than `docker`.
```

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+The new project will include a minimal variant with a simple custom package and settings.
+The command will take a few arguments, such as `--project name` and `--variant-name`.
+
+Similarly, a command will be available to add a new variant to an existing project.
+For example, `twoliter new-variant` which will add a variant.
+
+```
+❯ twoliter new project --name ootb-variants --variant-name example-dev
+❯ cd ootb-variants
+❯ twoliter build --variant hello-dev
```
ecpullen: nit: Should `hello-dev` be `example-dev` or was that intentional?

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+    └── lib.rs
+```
+
+A variant such as `metal-k8s-1.22` would consume the necessary kits.
+For example, it would reference `core-kit`, `k8s-kit` and `metal-kit`.
+To add `iam-authenticator`, `aws-kit` would be added.
+
+It may be convenient for certain build steps to have a single package representing a kit, such as `bottlerocket-aarch64-core-kit-0.1-1.aarch64.rpm`
+If so, these spec files could be added to the above directory tree.
+It is also possible to generate these spec files as they would be simple aggregations of all the packages in the kit.
```
webern: This section is missing a discussion of target architectures.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+Twoliter will eliminate the need for all build-host software other than `docker.`
+A bootstrapping container will be published that contains any software needed to kick off and sustain a `cargo make` build command.
+
+## Kits
+
+A central idea in this design is that the build system will be able to produce and consume Docker images that contain RPM packages.
+We call these images *kits*.
+A kit is a Yum repo distributed in its entirety, atomically, as a container image.
+Kits will satisfy requirements 4-8 above.
+Existing container registry mechanisms, such as ECR, will automatically sign these images satisfying requirement 1.
```
bcressey: Two points here:
* container signing isn't automatic, client tooling is needed
* it would be nice if RPMs were GPG-signed and if `dnf` enforced signature checking for the non-local kit repos

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+
+It should not be a requirement for a maintainer to push a kit to a container registry.
+In other words, it should not be necessary to publish a kit in order to create a Bottlerocket variant.
+This is where *local kits* come in.
+Kits defined in the project tree need not be pre-built before a `twoliter build` command can succeed.
+Twoliter will understand that the local kits need to be built before the rest of the variant build can proceed.
+
+### Kits Project Directory
+
+Currently, there are three directories used for sourcecode when building Bottlerocket:
```
bcressey: nit:
```suggestion
Currently, there are three directories used for source code when building Bottlerocket:
```

docs/design/README.md
jpmcb: I'd strongly recommend we consider uniquely tagging metadata images and not using a mutable tag.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+│   └── lib.rs
+└── vmware
+    ├── build.rs
+    ├── Cargo.toml
+    └── lib.rs
+```
+
+A variant such as `metal-k8s-1.22` would consume the necessary kits.
+For example, it would reference `core-kit`, `k8s-kit` and `metal-kit`.
+To add `iam-authenticator`, `aws-kit` would be added.
```
bcressey:
```suggestion
To add `aws-iam-authenticator`, `aws-kit` would be added.
```

As an inaugural bit of bikeshedding: `aws-iam-authenticator` might belong in the `k8s-kit` rather than `aws-kit`, since it is Kubernetes specific and can be used outside AWS - sort of like `ecr-credential-provider`.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+- Scripts accessed by cargo make, such as `docker-go`.
+- `Makefile.toml`
+
+It is within this container that Twoliter will invoke `cargo make` to initiate a build.
+
+### SDK Changes
+
+The existing build system expects access to the mounted core project git tree for certain things during the build.
+These will not be available and need to be added to the Bottlerocket SDK:
+- Scripts used by `docker build` commands such as `rpm2img`, `rpm2kmodkit` and `partplanner`.
```
zmrow: Typo: `partyplanner`

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+        │   └── bottlerocket-x86_64-kernel-5.15-archive-5.15.102-1.x86_64.rpm
+        ├── bootconfig
+        │   ├── empty-bootconfig.data
+        │   └── qemu-x86-console-bootconfig.data
+        ├── licenses
+        │   ├── COPYRIGHT
+        │   ├── LICENSE-APACHE
+        │   └── LICENSE-MIT
+        ├── migrations
+        │   └── bottlerocket-x86_64-migrations-0.0-0.x86_64.rpm
```
bcressey: Can we strip out all the non-RPM and non-repo content here?

This stuff shouldn't be published along with the kit.

docs/design/README.md
cbgbt: Why store the metadata in a separate image?

My assumption is that this is to allow dependency resolution to take place (possibly erroring) without needing to drag down a plethora of RPMs before doing so.

docs/design/README.md
bcressey: IMO - we should eliminate these from the design doc, and just backlog issues to clean them up from the initial implementation.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+        │   ├── d61fe62ca166eab1cbd502cf21cd977798ef348de49fb35e38a93ebf696eb0a5-filelists.sqlite.bz2
+        │   ├── fcd4dadc59c6d6512410b103a38762bf266389e01a60add12319191c4cc16ec4-other.xml.gz
+        │   └── repomd.xml
+        ├── bottlerocket-x86_64-acpid-2.0.34-1.x86_64.rpm
+        ├── bottlerocket-x86_64-apiclient-0.0-0.x86_64.rpm
+        ├── ...
+        └── bottlerocket-x86_64-wicked-0.6.68-1.x86_64.rpm
+```
+
+The file at `/local/etc/yum.repos.d` can be placed at `/etc/yum.repos.d` later when multiple kits are being aggregated.
```
bcressey: I don't follow the need for `/local/etc`.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+These files may need to be moved to the SDK or somewhere else.
+
+The presence of the `migrations` directory is temporary and controversial.
+Only the core project can have migrations, and we expect the migration system to go away.
+Ideally the presence of these will be completely eliminated as OOTB reaches maturity.
+
+### Kit Dependencies
+
+One kit may depend on a package from another kit.
+For example, a maintainer's out-of-tree package may depend on `glibc` from `bottlerocket-core-kit`.
```
bcressey:
```suggestion
```

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+```
+
+First Twoliter learns the external kit dependency graph using the traversal described for `twoliter update`.
+It pulls all kit container images that will be needed for the build.
+
+Twoliter starts the twoliter container environment and mounts the project directory.
+The container has a copy of Bottlerocket's `Makefile.toml` build tools in it.
+Twoliter invokes the `cargo make build` task.
+
+The `cargo make build` task is an alias of `cargo make variant` and its entrypoint is the variant's `Cargo.toml` file.
```
arnaldo2792, line 482: Will `twoliter` allow users to use the exact same variables used to configure a build with the existing `Makefile.toml`?

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+This may be published to a container registry and tagged with versions by the publisher:
+
+```
+registry.com/my-awesome-kit:v0.1.0
+registry.com/my-awesome-kit:v0.2.0
+etc.
+```
+
+Consider that a package within this kit depends on Bottlerocket's `glibc`, and that the required version has changed between this kit's `v0.1.0` and `v0.2.0` versions.
+Our JSON metadata for `my-awsome-kit` will let us know:
```
bcressey:
```suggestion
Our JSON metadata for `my-awesome-kit` will let us know:
```

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+      ]
+    }
+  }
+}
+```
+
+This metadata can be stored and updated in the kit container registry with a mutable tag so that `twoliter` can easily find it.
+
+```
+registry.com/my-awesome-kit:metadata
```
bcressey: Maybe something twoliter specific, like:
```suggestion
registry.com/my-awesome-kit:metadata.twoliter
```

docs/design/README.md
cbgbt: One possible use-case is that we could move all `aws-k8s-*` variants out-of-tree to the same twoliter project, thereby sharing the code they can, apart from kubelet builds.

Another example could be that I'm publishing a Bottlerocket appliance and I need a build which works on AWS and another which works on e.g. Raspberry Pi. The installed software would largely intersect, so it would be convenient to define them in the same source tree.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+registry.com/my-awesome-kit:v0.1.0
+registry.com/my-awesome-kit:v0.2.0
+etc.
+```
+
+These metadata files will give twoliter the information it needs to resolve, pull, and aggregate all dependency kits needed for a build.
+This dependency resolution may be kept in a lock file inspired by `Cargo.lock` if we find it helpful.
```
bcressey: If we don't have some sort of stable, Git-committable artifact that represents the results of the dependency resolution process, it will create challenges for reproducible builds.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+### Requirements
+
+The `twoliter` build tool should meet the following requirements.
+
+1. Binaries and metadata published for consumption by out-of-tree builds must be signed and automatically verified.
+2. Out-of-tree builds must support generating license attribution documents.
+3. All functionality necessary to build, test, and publish out-of-tree builds must be provided by a unified CLI.
+4. Maintainers must be able to define their own packages.
+5. Maintainers must be able to mix and match core project packages with their own.
+6. Maintainers must be able to replace a core project package with their own.
```
rpkelly, line 46: Does the combination of 4, 5, and 6 mean that they could choose to drop a core package?

docs/design/README.md
bcressey:
```suggestion
- Scripts used by `docker build` commands such as `rpm2img`, `rpm2kmodkit` and `partyplanner`.
```

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+        └── ...rpms
+```
+
+The kit priority order will be provided in the variant's `Cargo.toml` metadata section.
+Under the hood, Twoliter will composite the kits with dnf configurations for repo priority.
+
+After compositing the kits, the variant image build will proceed similarly to the way that it used to.
+Before installing the RPM packages, the Bottlerocket SDK's Fedora-provided `/etc/yum.repos.d/` directory will be occluded by the kits composite.
+This way, dnf install commands will pull from the kit-provided RPM repositories.
+If a package is available in more than one kit, it will be taken from the higher priority kit by dnf.
```
arnaldo2792, line 526: Will the user get any notification when a package is present in two different kits? What if one of kits used by the users introduced a package present in an existing kit and the user doesn't realize that until they run their image?

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+The list of desired packages, taken from the variant's `Cargo.toml` metadata (as it is now), will be installed along with any of their dependencies.
+
+After installing packages, the rest of the variant image creation steps proceed the same way that they do now.
+
+## Bottlerocket Forks
+
+Iterating on the build tools (like buildsys), `Makefile.toml`, docker files and scripts will be difficult unless Twoliter can take these things from source.
```
arnaldo2792: nit: `s/docker files/Docker files`.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+
+A project will have a top level `Twoliter.toml` file (and possibly a `Twoliter.lock` file).
+When `twoliter` commands are invoked, the tool will search for the `Twoliter.toml` file in `.`.
+If it doesn't find it, it will search in `..`, then `../..`, etc.
+This will allow the tool to work from any subdirectory within the project (and is inspired by `cargo`).
+
+`Twoliter.toml` will contain global project settings. Here is a sample `Twoliter.toml` file.
+
+```toml
+# This let's twoliter know if it is reading an older version of the Twoliter.toml
```
bcressey:
```suggestion
# This lets twoliter know if it is reading an older version of the Twoliter.toml
```

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+`twoliter update` is another inspiration from `cargo`.
+This command makes it easy for maintainers to update the packages (kits actually) they are using to the latest versions.
+
+For simplicity, no two variants in a project can use differing versions of a single kit.
+Twoliter will have the ability to create a kit dependency tree with the following procedure.
+First Twoliter will read the kit dependencies from each variant and add any external kits to its graph.
+Next Twoliter will read the kit definitions in the `kits` directory, and add all of their external dependencies to its graph.
+
+Once it has this list, Twoliter will begin a traversal.
+For each kit it will pull the `metdata` container for each external kit and store the kit.json files locally.
```
cbgbt: nit
```suggestion
For each kit it will pull the `metadata` container for each external kit and store the kit.json files locally.
```

docs/design/README.md
bcressey: `twoliter` prefix seems redundant in a `Twoliter.toml` file:
```suggestion
# schema, or if the directory structure of the project is from an older version.
schema-version = 1
```

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+Each time it finds a new external kit (i.e. an external kit that is depended on by an external kit), it will add this new dependency to the graph.
+(TODO: are cycles OK? They might be.)
```
cbgbt: I believe they're fine so long as the versioning is satisfied.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+# with names like `bottlerocket-aws-ecs-1` where the "project" name is a prefix
+# to the variant name. Here we would produce images with names like
+# `ootb-variants-example-dev` where `ootb-variants` is the project name and
+# `example-dev` is the variant name.
+project-name = "ootb-variants"
+
+# This allows the maintainer to rev a version globally for all variants defined in
+# the project. We may choose to enforce an opinionated versioning scheme here or
+# allow any arbitrary string, TBD.
+project-version = "1.0.0"
```
bcressey: In-place updates won't correctly with `updog` without semver versions.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+version = "v0.30.0"
+
+# This is the container that Twoliter uses to invoke `cargo make` and other build
+# commands. It will be available for both x86_64 and Arm hosts, but it is agnostic
+# as to which architecture the build is targeting.
+[twoliter-image]
+registry = "public.ecr.aws/bottlerocket"
+name = "twoliter"
+version = "v0.1.0"
+```
```
bcressey: Does this have a dependency on the `sdk-image`?

```suggestion
# This is the container that Twoliter uses to invoke `cargo make` and other build
# commands. It will be available for both x86_64 and Arm hosts, but it is agnostic
# as to which architecture the build is targeting.
[build-image]
registry = "public.ecr.aws/bottlerocket"
name = "twoliter"
version = "v0.1.0"
```
```

File: docs/design/README.md (hunk 42 above)
Comment (jpmcb) [reply docs/design/README.md:46]: I was just wondering about this: does this give an out of tree builder the capability to make a _less_ secure Bottlerocket? Should we let them? What if someone tries to masquerade their "evil" Bottlerocket variant as the real one?

File: docs/design/README.md [hunk 50]
```diff
@@ -0,0 +1,552 @@
...
+├── variants
+│   ├── example-dev
+│   │   ├── build.rs
+│   │   ├── Cargo.toml
+│   │   ├── pkg.rs
+│   │   └── variant.spec
+│   └── Cargo.toml
+├── COPYRIGHT
+├── LICENSE-APACHE
+└── LICENSE-MIT
```
bcressey: nit: maybe elide license files, since not all projects will have these licenses

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+First Twoliter will read the kit dependencies from each variant and add any external kits to its graph.
+Next Twoliter will read the kit definitions in the `kits` directory, and add all of their external dependencies to its graph.
+
+Once it has this list, Twoliter will begin a traversal.
+For each kit it will pull the `metdata` container for each external kit and store the kit.json files locally.
+Each time it finds a new external kit (i.e. an external kit that is depended on by an external kit), it will add this new dependency to the graph.
+(TODO: are cycles OK? They might be.)
+
+At this point Twoliter will determine the maximum version of each kit that is possible without introducing multiple versions of the same kit.
+TODO: this is probably hard, but there is prior art to look at.
```
cbgbt: You may actually need to store things as some sort of reference tree, since the dependency closure can change based on which minor version of a package you select. Tricky.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+
+Similarly, a command will be available to add a new variant to an existing project.
+For example, `twoliter new-variant` which will add a variant.
+
+```
+❯ twoliter new project --name ootb-variants --variant-name example-dev
+❯ cd ootb-variants
+❯ twoliter build --variant hello-dev
+  ...
+❯ twoliter new-variant my-new-variant
```
bcressey: Can we avoid options and use sub-commands / sub-sub-commands? E.g.
* `twoliter build variant ...`
* `twoliter new project --name ... --variant ...`
* `twoliter new variant ...`

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+The `twoliter` build tool should meet the following requirements.
+
+1. Binaries and metadata published for consumption by out-of-tree builds must be signed and automatically verified.
+2. Out-of-tree builds must support generating license attribution documents.
+3. All functionality necessary to build, test, and publish out-of-tree builds must be provided by a unified CLI.
+4. Maintainers must be able to define their own packages.
+5. Maintainers must be able to mix and match core project packages with their own.
+6. Maintainers must be able to replace a core project package with their own.
+7. Maintainers must be able to use core project packages without building them.
+8. Maintainers must be able to take and use packages from multiple sources.
```
arnaldo2792: There wasn't discussion about "opt-out" features, will that be covered on a separate revision?

docs/design/README.md
cbgbt: Can we take advantage of an existing package manager to do this somehow?

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+# The container image with the SDK, buildsys, etc. This is provided as a structure
+# so that we can switch on the target architecture. For example, when building
+# Bottlerocket images for x86_64 machines, the following would be produced:
+# public.ecr.aws/bottlerocket/bottlerocket-build-tools-x86_64:v1.14.1. Twoliter
+# will assume the container is available locally if the registry field is null
+# or an empty string.
+[sdk-image]
+registry = "public.ecr.aws/bottlerocket"
+name = "bottlerocket-sdk"
+version = "v0.30.0"
```
bcressey: How does this intersect with `twoliter update` and figuring out the right SDK to use?

docs/design/README.md
webern: > Can we strip out all the non-RPM and non-repo content here?

Note: this is about the licenses.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+    └── bottlerocket-core-kit
+        ├── archives
+        │   └── bottlerocket-x86_64-kernel-5.15-archive-5.15.102-1.x86_64.rpm
+        ├── bootconfig
+        │   ├── empty-bootconfig.data
+        │   └── qemu-x86-console-bootconfig.data
+        ├── licenses
+        │   ├── COPYRIGHT
+        │   ├── LICENSE-APACHE
+        │   └── LICENSE-MIT
```
webern:
```suggestion
```

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+When aggregating kits, the yum repos will be prioritized according to the maintainer's requirements.
+
+The presence of `bootconfig` is questionable.
+These files may need to be moved to the SDK or somewhere else.
+
+The presence of the `migrations` directory is temporary and controversial.
+Only the core project can have migrations, and we expect the migration system to go away.
+Ideally the presence of these will be completely eliminated as OOTB reaches maturity.
+
+### Kit Dependencies
```
arnaldo2792, line 209: Can kits be built from other kits? Say we have the `core-kit`, that depends on the `bootstrap-containers-kit`, but in my variant, I don't want bootstrap-containers, can I build my `core-kit` without it?

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+# tree /local --dirsfirst
+/local
+├── etc
+│   └── yum.repos.d
+│       └── bottlerocket-core-kit.repo
+└── kits
+    └── bottlerocket-core-kit
+        ├── archives
+        │   └── bottlerocket-x86_64-kernel-5.15-archive-5.15.102-1.x86_64.rpm
+        ├── bootconfig
```
webern: Generate these on the fly using tooling in the SDK.

docs/design/README.md
```diff
@@ -0,0 +1,552 @@
...
+# allow any arbitrary string, TBD.
+project-version = "1.0.0"
+
+# The container image with the SDK, buildsys, etc. This is provided as a structure
+# so that we can switch on the target architecture. For example, when building
+# Bottlerocket images for x86_64 machines, the following would be produced:
+# public.ecr.aws/bottlerocket/bottlerocket-build-tools-x86_64:v1.14.1. Twoliter
+# will assume the container is available locally if the registry field is null
+# or an empty string.
+[sdk-image]
```
webern: kits need to declare the sdk they built with
all the kits in a build need to declare the same sdk
this could be a wildcard, i.e. the SDK could be `*` and we could take what our dependencies are using.

docs/design/README.md
webern: Note: remove migrations and other tech debt discussions from the doc.

docs/design/README.md
webern: > Will the user get any notification when a package is present in two different kits? What if one of kits used by the users introduced a package present in an existing kit and the user doesn't realize that until they run their image?

We could use `dnf list installed` at the end of variant image build and output that information somewhere. I'll add it to the doc.

docs/design/README.md
webern: In principal yes. In practice it may be a subset, and over time the interfaces of `twoliter` and `Makefile.toml` may (intentionally) diverge.

docs/design/README.md
webern: > Does this have a dependency on the sdk-image?

It does not. It's more like what your "host" is now. It just needs cargo, docker and the Makefile. I'm not crazy about `build-image` as the name.

docs/design/README.md
webern: > How does this intersect with twoliter update and figuring out the right SDK to use?

This is an interesting question. I'm going to comment out these lines in the file for now and add an explanation that says, basically you would rarely use this and would need to build all kits, including external kits, from source using your preferred SDK.

docs/design/README.md
webern: > Why store the metadata in a separate image?
>
> My assumption is that this is to allow dependency resolution to take place (possibly erroring) without needing to drag down a plethora of RPMs before doing so.

Correct. So that dependency resolution can be lightweight before pulling the bigger images.

docs/design/README.md
webern: > Can kits be built from other kits? Say we have the core-kit, that depends on the bootstrap-containers-kit, but in my variant, I don't want bootstrap-containers, can I build my core-kit without it?

When building the kit, the dependency would need to be available because that is how RPM works. There is not a concept of *your* core kit. Here `core-kit` refers to a certain set of packages in one of the kits produced by the Bottlerocket team. *Your* kit would not need to have the same packages as the `core-kit`, you could just install them from the `core-kit`. I hope this is helping.

docs/design/README.md
webern: > I'd like to hear more on this: are migrations going away because of the new settings mechanism that are coming (mentioned earlier)?

Yes that's correct. We anticipate the settings SDK to do away with the need for settings migrations. If we need to retain functionality for OS-level migrations (TBD), we will need a design for that.

docs/design/README.md
webern: > Will different out of tree builds have different boot configs? Maybe it make sense to have the bootconfig as it's own kit if we anticipate out of tree builders to modify/make their own bootconfigs.

Custom bootconfigs are not expected to be a requirement.

docs/design/README.md
webern: > does x-kit just mean kits/x?

Yes. I like this naming convention. `myname` produces a kit container named `myname-kit`. TBD as to whether we will enforce this convention.

docs/design/README.md
webern: > Does "mix and match" here mean that someone could modify the core components of a upstream, in-tree package?

Not exactly. In order to alter an upstream package, the maintainer would need to replace it. I don't think there is a notion of patching an upstream RPM package (though RPM packages themselves have a notion of patching upstream sources)

webern:
https://github.com/bottlerocket-os/twoliter/compare/36fe14f60948559e904c89ef3c662655ba1c8b26..2c6519d095a2ce9cf5a307e3dcaea185363a61c4

This includes all the easy fixes, most of the related comments have been marked as resolved.

webern:
https://github.com/bottlerocket-os/twoliter/compare/2c6519d095a2ce9cf5a307e3dcaea185363a61c4..54524c5b36b2a3b32a1fc912eb633713d8574280

I had missed a few of Ben's quick fixes.

webern:
https://github.com/bottlerocket-os/twoliter/compare/54524c5b36b2a3b32a1fc912eb633713d8574280..9b89f03de9e5c2e359822e097200b7a677a829a2

Add a requirement that makes it more clear that maintainers can opt out of certain Bottlerocket features by choosing not to install the relevant packages. No change was needed to the design.

webern:
https://github.com/bottlerocket-os/twoliter/compare/9b89f03de9e5c2e359822e097200b7a677a829a2..1508c19846f3326e780ca58197dcecbbccac8c2f

Print `dnf list installed` for auditing.

docs/design/README.md
webern: I added a specific requirement that makes it clear dropping a core package is a supported use case.

webern:
https://github.com/bottlerocket-os/twoliter/compare/1508c19846f3326e780ca58197dcecbbccac8c2f..dad8a9bb5476f15803e9ac9d64af73cdf812a938

- Adds the idea that the kits need to all be built with the same SDK.
- Adds x86_64 suffixes to kit docker repo names to imply how arches will be handled (same as SDK arches)
- Adds a sequence diagram to the best of my ability

webern:
https://github.com/bottlerocket-os/twoliter/compare/dad8a9bb5476f15803e9ac9d64af73cdf812a938..d62870ea1907b2555040cc421cb8c609e5c3f9ea

- put in some placeholder language about using cargo for dependency resolution
- placeholder language about signing artifacts
- improve diagram
- add bootconfig to the things the SDK will need

webern:
This is pretty serviceable and I would like to get it merged and get going on issue creation. I think we can safely assume that the design will evolve but we have a pretty good picture of what we are building.

The diff of everything I've done since everyone reviewed it is: https://github.com/bottlerocket-os/twoliter/compare/da356b0..d62870ea1907b2555040cc421cb8c609e5c3f9ea

docs/design/README.md
```diff
@@ -0,0 +1,595 @@
...
+
+It will be helpful to understand how the existing build system works.
+The best way to learn this is to read the blog post *[How the Bottlerocket build system works]*.
+
+Briefly, the entry point of the existing build system is Rust's build system, `cargo`.
+A plugin called [cargo make] gets its instructions from `Makefile.toml`.
+We use `cargo` to resolve dependencies between variants and packages, and we use custom build scripts to launch docker-hosted build environments.
+Packages are defined and built using RPM.
+
+These package builds take place in a docker environment using `docker build` commands and `buildx`.
```
bcressey: Do we use `buildx` anywhere today? Maybe this:
```suggestion
These package builds take place in a docker environment using `docker build` commands with the `BuildKit` backend.
```

docs/design/README.md
```diff
@@ -0,0 +1,595 @@
...
+The toolchains used to compile Rust, C, C++ and Go software are distributed in a container image called the [bottlerocket-sdk].
+It is in the SDK environment that the package builds take place.
+
+Once all the software for a Bottlerocket image has been built into RPM images, a final image creation step occurs.
+A `docker build` command installs everything into a rooted image file.
```
bcressey: Not sure what you mean by "rooted", which can have some unfortunate connotations:

```suggestion
A `docker build` command installs everything into an image file.
```

docs/design/README.md
```diff
@@ -0,0 +1,595 @@
...
+To add `aws-iam-authenticator`, `aws-kit` would be added.
+
+It may be convenient for certain build steps to have a single package representing a kit, such as `bottlerocket-aarch64-core-kit-0.1-1.aarch64.rpm`
+If so, these spec files could be added to the above directory tree.
+It is also possible to generate these spec files as they would be simple aggregations of all the packages in the kit.
+
+### Kit Structure
+
+A kit container will be built using `FROM scratch` because it does not need a runtime environment.
+It's directory structure will look like this:
```
bcressey: nit:
```suggestion
Its directory structure will look like this:
```

docs/design/README.md
```diff
@@ -0,0 +1,595 @@
...
+```sh
+# tree /local --dirsfirst
+/local
+├── etc
+│   └── yum.repos.d
+│       └── bottlerocket-core-kit.repo
+└── kits
+    └── bottlerocket-core-kit
+        ├── archives
+        │   └── bottlerocket-x86_64-kernel-5.15-archive-5.15.102-1.x86_64.rpm
```
bcressey: `archives` is another bit that isn't really core to the structure:
```suggestion
```

docs/design/README.md
```diff
@@ -0,0 +1,595 @@
...
+    "arch": "x86_64",
+    "sdk": "public.ecr.aws/bottlerocket/bottlerocket-sdk-x86_64:v0.50.0",
+    "dependencies": [
+        "public.ecr.aws/bottlerocket/bottlerocket-core-kit-x86_64:v1.15.1"
+    ]
+  }
+}
+```
+
+In this example, it would be an error if `bottlerocket-core-kit-x86_64:v1.15.1` declared that it had been built with anything other `public.ecr.aws/bottlerocket/bottlerocket-core-kit-x86_64:v1.15.1`.
```
bcressey: A fix, assuming I understand this example correctly:

```suggestion
In this example, it would be an error if `bottlerocket-core-kit-x86_64:v1.15.1` declared that it had been built with anything other `public.ecr.aws/bottlerocket/bottlerocket-sdk-x86_64:v0.50.0`.
```

docs/design/README.md
```diff
@@ -0,0 +1,595 @@
...
+- **Maintainer**: A person who creates Bottlerocket variants using Twoliter.
+- **Kit**: A container image with RPM packages in it.
+
+[main Bottlerocket git repository]: https://github.com/bottlerocket-os/bottlerocket
+
+### Requirements
+
+The `twoliter` build tool should meet the following requirements.
+
+1. Binaries and metadata published for consumption by out-of-tree builds must be signed and automatically verified.
```
strategicpause, line 40: Does this require work by the team who would be using OOTB? ie: would we need to ensure binaries we provide are signed?

docs/design/README.md
```diff
@@ -0,0 +1,595 @@
...
+2. Out-of-tree builds must support generating license attribution documents.
+3. All functionality necessary to build, test, and publish out-of-tree builds must be provided by a unified CLI.
+4. Maintainers must be able to define their own packages.
+5. Maintainers must be able to mix and match core project packages with their own.
+6. Maintainers must be able to replace a core project package with their own.
+7. Maintainers must be able to opt out of certain Bottlerocket subsystems by choosing not to install them.
+8. Maintainers must be able to use core project packages without building them.
```
strategicpause, line 49: ie: host container runtime?

docs/design/README.md
```diff
@@ -0,0 +1,595 @@
...
+The `/etc/yum.repos.d` file can be used later to configure `dnf`.
+This will allow `dnf` to treat each kit as a yum repository.
+When aggregating kits, the yum repos will be prioritized according to the maintainer's requirements.
+
+### Kit Dependencies
+
+One kit may depend on a package from another kit.
+All kits used together, need to have been built by the same SDK.
+Another way to say this is that all kits must depend on the same SDK.
+These dependencies need to be specified at the kit level so that tooling understands which kits are required for a build.
```
bcressey: The SDK is the starting point for this problem, given that Rust binaries are aggressively linked to the `libstd-*.so` shared library.

However, it can arise for kit:kit dependencies as well, for essentially the same reason: any shared library that one kit contributes can become a hard dependency for its downstream kits.

docs/design/README.md
```diff
@@ -0,0 +1,595 @@
...
+Twoliter will understand how to start a Bottlerocket build using the OOTB user's sources, packages etc.
+It will inject these directories into the existing build system by setting existing and new `Makefile.toml` variables.
+
+Twoliter will eliminate the need for all build-host software other than `docker`.
+A bootstrapping container will be published that contains any software needed to kick off and sustain a `cargo make` build command.
+
+## Kits
+
+A central idea in this design is that the build system will be able to produce and consume Docker images that contain RPM packages.
+We call these images *kits*.
```
strategicpause, line 106: Would a variant maintainer expected to be brining their own container image? For example, a container image which may contain precompiled binaries or build tools for adding binaries to the Bottlerocket rootfs.

docs/design/README.md
webern:
```suggestion
In this example, it would be an error if `bottlerocket-core-kit-x86_64:v1.15.1` declared that it had been built with anything other `public.ecr.aws/bottlerocket/bottlerocket-sdk-x86_64:v0.50.0`.
```

docs/design/README.md
```diff
@@ -0,0 +1,595 @@
...
+
+In order to satisfy self-signing requirements, Kit RPMs will need to be GPG-signed.
+The kit will be distributed with the GPG public key, which means we need the container to also be signed.
+Existing container-signing mechanisms will be made available.
+(**TODO** - refine the design for signing artifacts, see https://github.com/bottlerocket-os/twoliter/issues/5)
+
+### Local and External Kits
+
+During the build process, there will be a distinction between *local kits* and *external kits*.
+A *local kit* is any kit where the packages are defined in the project tree and the maintainer wants to build them from source as part of the Bottlerocket build.
```
strategicpause, line 124: If I understand correctly, I could have a private package which is responsible or building this kit? This would prevent me from having to have a dependency on an external container registry for the build of my variant?

docs/design/README.md
```diff
@@ -0,0 +1,595 @@
...
+Twoliter will understand that the local kits need to be built before the rest of the variant build can proceed.
+
+### Kits Project Directory
+
+Currently, there are three directories used for source code when building Bottlerocket:
+- `sources`: A Cargo workspace for first-party Rust code.
+- `packages`: A collection of Rust packages that build RPM packages.
+- `variants`: A Cargo workspace for defining what Bottlerocket variants.
+
+A new directory will be added: `kits`.
```
strategicpause, line 140: Would kits be a replacement for some packages which are a pain for in-tree builds? xfsprogs comes to mind.

docs/design/README.md
```diff
@@ -0,0 +1,595 @@
...
+See https://github.com/bottlerocket-os/twoliter/issues/4
+
+`twoliter update` is another inspiration from `cargo`, this time inspired by `cargo update`.
+This command makes it easy for maintainers to update the packages (kits actually) they are using to the latest versions.
+
+In the original Bottlerocket build system, we used Cargo crates to model RPM dependencies.
+In Twoliter, we will again lean on Cargo to manage dependencies for us.
+We will create a local Cargo registry to represent kit and SDK versions, then use `cargo update` to resolve the latest dependencies.
+
+Before describing this procedure, we should consider that no two versions of the same kit can be in the dependency tree.
```
bcressey, line 417: This is potentially solved by introducing the same tight version binding between kits as exists between SDK and kit.

docs/design/README.md
```diff
@@ -0,0 +1,595 @@
...
+# schema, or if the directory structure of the project is from an older version.
+schema-version = 1
+
+# This is the name of this project, which can have multiple variants in it. For
+# in-tree builds the project name is "bottlerocket", and thus we produce images
+# with names like `bottlerocket-aws-ecs-1` where the "project" name is a prefix
+# to the variant name. Here we would produce images with names like
+# `ootb-variants-example-dev` where `ootb-variants` is the project name and
+# `example-dev` is the variant name.
+project-name = "ootb-variants"
```
bcressey: I'm not entirely sure about the project name prefix - it seems simpler to just have kits and variants define their own names.

This implicit behavior feels like it complicates the dependency resolution process, because you could copy the same project tree into a different repository and suddenly need to resolve "ootb-variants-core-kit" instead of "bottlerocket-core-kit" because this prefix changed.

docs/design/README.md
```diff
@@ -0,0 +1,595 @@
...
+
+# This allows the maintainer to rev a version globally for all variants defined in
+# the project. Twoliter will enforce semver because the update system requires it.
+project-version = "1.0.0"
```
bcressey, line 340: Just to say, despite my `project-name` skepticism, I'm on board with this top-level `project-version` field.

docs/design/README.md
```diff
@@ -0,0 +1,595 @@
...
+Each kit will be represented by an empty Cargo `lib.rs` project with the same name as the kit.
+The Cargo package version will be given by deterministically adding the kit's version to 1.x (**TODO** - hand-waved).
+
+The SDK will be a special Cargo package.
+Two SDKs may have the same version, but come from different container registries.
+To ensure a single SDK is used, it will be given a version in Cargo like this (**TODO** - or some hash of the necessary unique information):
+
+```toml
+[dependencies]
+sdk = "=0.50.0-public.ecr.aws_bottlerocket_bottlerocket-sdk-x86_64"
```
bcressey: I definitely don't want "arch" to appear here. I'd like that to be something the system resolves automatically to reduce the friction for adding a new architecture - going from `x86_64` only to also supporting `aarch64`, for example.

I'd also suggest having some sort of default here where everything is assumed to be rooted under `public.ecr.aws/bottlerocket` unless otherwise specified. So:

```
[dependencies]
# implicitly gets modified to append the target arch during dependency resolution
# results in public.ecr.aws/bottlerocket/bottlerocket-sdk-x86_64:v0.50.0
bottlerocket-sdk = "0.50.0"

# explicitly set to some other prefix
my-kit = { source = "my.registry/prefix", version = "0.10.0" }
```

webern:
> https://github.com/bottlerocket-os/twoliter/compare/d62870ea1907b2555040cc421cb8c609e5c3f9ea..5f335c43194a5127f8ed6a8630d178796e91affa

Use plantuml instead of drawio

docs/design/README.md
webern: > I'm not entirely sure about the project name prefix

This was a remnant from earlier thinking, I will delete the explanation and leave it a mystery as to why one might want a name for their project (aesthetic bike shedding?)

docs/design/README.md
webern: > archives is another bit that isn't really core to the structure:

I'll delete it from the design doc, but it needs to get into the build somehow. Basically by deleting it from the design doc we are saying "figure out kmod-kits later". Bottom line is rpm2kmodkits uses these and our cargo makefile  creates them.

docs/design/README.md
webern: > Would kits be a replacement for some packages which are a pain for in-tree builds? xfsprogs comes to mind.

That's sort of the idea. If you wanted to vend an `xfsprogs` package for multiple out-of-tree projects to use, you could create a kit with it and any other packages it needs.

If you just need to use it locally in a single out-of-tree build project, you can sort of forget about "kits" since you won't be publishing any.

Either way the rpm spec for `xfsprogs` would exist in your out-of-tree twoliter project's `packages` directory.

docs/design/README.md
webern: > If I understand correctly, I could have a private package which is responsible or building this kit? This would prevent me from having to have a dependency on an external container registry for the build of my variant?

You will need to depend on a container registry in order to depend on kits (and the packages they contain) from the Bottlerocket core project. Mirroring of these will be necessary for Enterprises that cannot reach out to the Internet. For the moment we are assuming the mirrors can be configured with containerd or docker.

docs/design/README.md
webern: > Would a variant maintainer expected to be brining their own container image? For example, a container image which may contain precompiled binaries or build tools for adding binaries to the Bottlerocket rootfs.

Unless you want to publish your RPM packages for other Bottlerocket out-of-tree projects to consume, then you don't need to think or worry about "kits". If you aren't publishing RPMs to other teams or to the outside world, and if they are only used as part of your own image builds, then kits are an implementation detail you don't need to worry about too much.

docs/design/README.md
webern: > ie: host container runtime?

Yes, I will add a note so that we understand the purpose behind this requirement.

docs/design/README.md
webern: > Does this require work by the team who would be using OOTB? ie: would we need to ensure binaries we provide are signed?

The only work will be to provide the signing keys during a build. Verifying the signatures of depended-upon binaries (i.e. core project RPMs) will be automatic.

webern (commented):
Answered some questions.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/5f335c43194a5127f8ed6a8630d178796e91affa..f6a6503285ea31720ed812bb0150360963231974

Addresses all known edit requests.

docs/design/README.md
```diff
@@ -0,0 +1,592 @@
...
+
+### Twoliter Container
+
+Twoliter will use this container to bootstrap the build process.
+It needs to contain items the Bottlerocket build system requires.
+Fedora will be its base as it is for the SDK.
+
+The image will include the following:
+- the docker CLI
+- cargo make
```
bcressey: nit: these will be in the SDK
```suggestion
```

docs/design/README.md
```diff
@@ -0,0 +1,592 @@
...
+The rest of the variant image creation steps proceed the same way that they do now.
+
+## Bottlerocket Forks
+
+Iterating on the build tools (like buildsys), `Makefile.toml`, Docker files and scripts will be difficult unless Twoliter can take these things from source.
+Additionally, some maintainers will need to make changes in the Bottlerocket sourcetree.
+These use-cases lead to requirement 10 (*Maintainers must be able to fork and patch the core project in lieu of using the published binary artifacts*).
+
+In order to facilitate these use-cases Twoliter and build system development cycles, Twoliter will have the ability to be pointed at a local checkout of such sourcecode.
+For this workflow it will be able to build Bottlerocket kits, the Twoliter container, and overwrite files in the Bottlerocket SDK.
```
bcressey: I thought our plan was to have the `twoliter` build itself bundle all of this together somehow. So if you wanted to iterate on it, you could check out this project, do a local release build, and use that `twoliter` binary instead.

docs/design/README.md
```diff
@@ -0,0 +1,592 @@
...
+# loading. If you are working on the SDK or need to use a special SDK, then it
+# will be necessary to build all kits from source, including external kits.
+# Twoliter will enforce that all kits have been build with the same SDK.
+#
+# How this override may be used is TBD.
+#
+# [sdk-image]
+# registry = "public.ecr.aws/bottlerocket"
+# name = "bottlerocket-sdk"
+# version = "v0.30.0"
```
bcressey: I am still unclear on where the initial SDK comes from.

If the default version is baked into `twoliter`, then we would have to do a new release of `twoliter` to change the default, and updating the `twoliter` binary would tend to break existing projects unless their kit dependencies had already been rebuilt.

If it's not baked in, then it's not obvious how `twoliter new` should set up a new project that doesn't have kit dependencies yet.

I suppose we could do a combination:
* twoliter binaries are pinned to a default SDK (latest at time of build). This SDK is used (if needed) for any pre-build steps required. (If we can put all the logic in `twoliter` then ideally we wouldn't need this, which would be great because we **really** don't want to pull two SDKs.)
* `twoliter new` will automatically add a dependency on the most recent core-kit release, as determined by querying container image tags and picking the latest per semver.
* `twoliter add kit` will pick the latest version of the kit (or the specified one), potentially downgrading any other kit (like core) or the SDK in order to satisfy dependencies.
* `twoliter new --freestanding` (or similar option) would skip the core kit query and use a "latest SDK" query instead. `twoliter new --freestanding --sdk-image=...` would also skip that query and just use the specified image.

docs/design/README.md
```diff
@@ -0,0 +1,592 @@
...
+
+# This is the container that Twoliter uses to invoke `cargo make` and other build
+# commands. It will be available for both x86_64 and Arm hosts, but it is agnostic
+# as to which architecture the build is targeting.
+[twoliter-image]
+registry = "public.ecr.aws/bottlerocket"
+name = "twoliter"
+version = "v0.1.0"
```
bcressey: I am still excited about the idea of baking this into `twoliter` itself.

docs/design/README.md
```diff
@@ -0,0 +1,592 @@
...
+
+At this point Twoliter will build a package that represents the maintainers top level dependencies.
+
+```toml
+[dependencies]
+bottlerocket-sdk = "=0.50.0"
+bottlerocket-core-kit = "=1.1.15"
+
+# explicitly set to some other prefix
+third-party-kit = { source = "some.registry/prefix", version = "=0.10.0" }
```
bcressey, line 448: What file does this go into? The top-level `Twoliter.toml`?

One idea is for that file to just have an unversioned list of kits (and maybe sources), like:
```
kits = [
    "bottlerocket-core-kit",
    "bottlerocket-aws-kit",
    "bottlerocket-k8s-kit",
    { name = "my-other-kit", source = "some.registry/prefix" },
]
```

And then `twoliter add` / `twoliter update` would rewrite `Twoliter.lock` with the latest versions that satisfy their mutual dependencies.

```
[sdk]
name = "bottlerocket-sdk"
version = "0.100.0"
source = "public.ecr.aws/bottlerocket"

[[kit]]
name = "bottlerocket-core-kit"
version = "0.50.0"
source = "public.ecr.aws/bottlerocket"

[[kit]]
name = "bottlerocket-aws-kit"
version = "0.1.0"
source = "public.ecr.aws/bottlerocket"

[[kit]]
name = "bottlerocket-k8s-kit"
version = "0.2.0"
source = "public.ecr.aws/bottlerocket"

[[kit]]
name = "my-other-kit"
version = "0.0.1"
source = "some.registry/prefix"
```

A project like this might end up seeing many, relatively independent builds of each kit if their only common dependency is the SDK and if the kit projects move faster than the SDK (which is likely). Trying to keep track of that by hand would be maddening.

I'm OK with not knowing how the dependency solver will be implemented, but I want to make sure the design reflects the intended maintainer experience.

docs/design/README.md
webern: > I thought our plan was to have the twoliter build itself bundle all of this together somehow. So if you wanted to iterate on it, you could check out this project, do a local release build, and use that twoliter binary instead.

You're right. I removed this section and requirement as it is no longer clear there will be any benefit from pointing at a fork. If a maintainer needs to patch Bottlerocket, they will publish their own modified core project kits as are affected by any patches then consume them in their project.

docs/design/README.md
webern: > What file does this go into? The top-level Twoliter.toml?

No, I think kit dependencies are done in variant/Cargo.toml.  What I was showing here is an example of how the resolver might use cargo to generate crates that will enforce the versioning of kits.

docs/design/README.md
webern: > I am still excited about the idea of baking this into twoliter itself.

A remnant of previous design iterations. Removed. Some combination of the following will describe the new twoliter container paradigm (and I have now alluded to it in the containers section)
- #6
- #4
- #11

docs/design/README.md
webern: > I am still unclear on where the initial SDK comes from.

So am I. I've deleted this from Twoliter.toml for now and will figure this out in:
- https://github.com/bottlerocket-os/twoliter/issues/11
- https://github.com/bottlerocket-os/twoliter/issues/4

webern:
> https://github.com/bottlerocket-os/twoliter/compare/f6a6503285ea31720ed812bb0150360963231974..54f3af6c9b0e9fa44046450aceab83003557bebf

Capture some of the things that have changed about the design during discussions. Ben's comments.

webern (commented):
Responses

webern:
> https://github.com/bottlerocket-os/twoliter/compare/54f3af6c9b0e9fa44046450aceab83003557bebf..bd0c97ed64f3d54e29813d2594a06a84f448cbad

Small fixes. @bcressey this is ready for review again.

bcressey (approved):

docs/design/README.md
```diff
@@ -0,0 +1,561 @@
...
+        └── bottlerocket-x86_64-wicked-0.6.68-1.x86_64.rpm
+```
+
+The `/etc/yum.repos.d` file can be used later to configure `dnf`.
+This will allow `dnf` to treat each kit as a yum repository.
+When aggregating kits, the yum repos will be prioritized according to the maintainer's requirements.
+
+### Kit Dependencies
+
+A kit may depend on packages other kits.
```
cbgbt: nit:
```suggestion
A kit may depend on packages in other kits.
```

docs/design/README.md
```diff
@@ -0,0 +1,561 @@
...
+These dependencies need to be specified at the kit level so that tooling understands which kits are required for a build.
+
+All kits used together, need to have been built by the same SDK.
```
cbgbt:
```suggestion
All kits used together need to have been built by the same SDK.
```

cbgbt (approved):
🚀

webern:
> https://github.com/bottlerocket-os/twoliter/compare/bd0c97ed64f3d54e29813d2594a06a84f448cbad..62c2771556f13cbca8308275d16c68057e07df89

Small fixes.

PR #22, opened 2023-07-07
github actions

webern:
Add a simple GitHub actions file to build, test and lint the project.

*Issue #, if available:*

N/A

*Description of changes:*

Enable GitHub actions.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,25 @@
+name: Rust
+on:
+  # triggers when a PR is posted
+  pull_request:
+    branches: [develop]
+    paths-ignore:
+      - "**.md"
+      - "**.plantuml"
+      - "**.svg"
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
Looks good. Two things we may want to consider is [making sure this doesn't run on forks](https://github.com/bottlerocket-os/bottlerocket/blob/develop/.github/workflows/build.yml#L29), and maybe adding some settings so that if someone uploads a PR update before the checks are done, it will [cancel the running check](https://github.com/bottlerocket-os/bottlerocket/blob/develop/.github/workflows/build.yml#L21-L23) since it will no longer be relevant.

But approving as I think those are just nice to haves and not critical.

jpculp (approved):
Not entirely sure why it's not running in this PR, but it worked when I copied it to my fork.

Issue #14, opened 2023-06-05
twoliter make can build bottlerocket


webern:
### Description

Get the `twoliter make` command working for all `Makefile.toml` targets.
This will require mounting the source repo into the Twoliter execution environment.

### Definition of Done

When all of the `cargo make` targets in the Bottlerocket `Makefile.toml` can be executed through `twoliter make`.

#### Depends on:

- [x] https://github.com/bottlerocket-os/twoliter/issues/13
- [x] https://github.com/bottlerocket-os/twoliter/issues/7
- [x] https://github.com/bottlerocket-os/twoliter/issues/9
- [x] https://github.com/bottlerocket-os/twoliter/issues/32
- [x] https://github.com/bottlerocket-os/bottlerocket/issues/3294

```
id: texec_command_works
epic: twoliter_exec
```

Related to https://github.com/bottlerocket-os/bottlerocket/issues/2669

PR #29, opened 2023-08-11
twoliter make

### LINKED ISSUE #13: basic twoliter make command
### Description

Create a subcommand in Twoliter named `make` which passes through arguments to `cargo make`.
This will be used to execute `Makefile.toml` targets during the Bottlerocket monorepo transition to Twoliter.

For example, if a user wants to run `cargo make -e FOO=bar variant`, then the following command will run (as an alias in the Bottlerocket monorepo):

```sh
twoliter make -e FOO=bar variant
```

The arguments can be taken exactly as the come and be passed directly to `cargo make`.

### Definition of Done

When `twoliter make` ~spins up its bootstraping container, and executes~ `cargo make` ~inside it~ with the same arguments it received.

It is not important for this issue that proper directories be mounted, and the `cargo make` target will certainly fail. This is fine.

#### Depends on:

- [x] https://github.com/bottlerocket-os/bottlerocket-sdk/issues/108

```
id: texec_command
epic: twoliter_exec
```

webern:
*Issue #, if available:*

Closes #13
Progress on #14

*Description of changes:*

```
    twoliter make

    Adds a command that can be used to build the Bottlerocket core project. This command provides a
    facade over the cargo make interface that the core project currently uses.
```

*Testing and TODO*

- [x] Check-in `Makefile.toml` with all of its git history from the Monorepo
- [x] Build a variant
- [x] Build a variant in our automated build environment


twoliter/build.rs
```diff
@@ -0,0 +1,87 @@
+/*!
+
+Prepare and package embedded tools in a tarball to be included with Twoliter.
```
cbgbt, line 3: Clever approach. Will this work correctly with cross-compilation? I think it depends on whether or not cargo's `OUT_DIR` is adjusted based on the target arch.

twoliter/src/tools.rs
```diff
@@ -0,0 +1,84 @@
+use crate::common::fs;
```
cbgbt: It would be a good idea to write some tests for this, e.g. set up a tmp directory with the sentinel and check that `should_install` does the right thing. Possibly installing as well, but I'm not sure how you'd verify that it worked correctly.

twoliter/src/cmd/exec.rs
```diff
@@ -0,0 +1,107 @@
...
+        for cargo_make_arg in &self.additional_args {
+            args.push(cargo_make_arg.clone());
+        }
+
+        exec(Command::new("cargo").args(args)).await
+    }
+}
+
+/// A list of environment variables that don't conform to naming conventions but need to be passed
+/// through to the `cargo make` invocation.
```
cbgbt: :cry:

cbgbt (approved):
Nice work! :rocket:

twoliter/build.rs
webern: I expect OUT_DIR is architecture specific. For the moment it's just Makefile.toml but I will make sure this is working when I add the binaries (highly confident it will work as expected).

webern:
> https://github.com/bottlerocket-os/twoliter/compare/4478b3193a2ff6587f324d5bde829a760d0b7472..e8aa220e3ce0643cc494c12ad4998f2b76f26a3a

Add `install_tools` tests.

twoliter/src/cmd/exec.rs
```diff
@@ -0,0 +1,107 @@
...
+use crate::project;
+use crate::tools::install_tools;
+use anyhow::Result;
+use clap::Parser;
+use log::trace;
+use std::path::PathBuf;
+use tokio::process::Command;
+
+/// Run a cargo make command in Twoliter's build environment. Known Makefile.toml environment
+/// variables will be passed-through to the cargo make invocation.
```
bcressey: Should this command be called `twoliter make`? Since it ends up being pretty specific to `cargo-make` and `Makefile.toml`.

twoliter/src/cmd/exec.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+/// A list of environment variables that don't conform to naming conventions but need to be passed
+/// through to the `cargo make` invocation.
+const ENV_VARS: [&str; 13] = [
+    "ALLOW_MISSING_KEY",
+    "AMI_DATA_FILE_SUFFIX",
+    "BOOT_CONFIG",
+    "BOOT_CONFIG_INPUT",
```
bcressey: nit: I think these would be matched by this below:
```
key.starts_with("BOOT_CONFIG")
```

(I'm also fine with retracting the related feature altogether, but that's out of scope here.)

twoliter/src/cmd/exec.rs
```diff
@@ -0,0 +1,107 @@
...
+    "CARGO_MAKE_CARGO_ARGS",
+    "CARGO_MAKE_DEFAULT_TESTSYS_KUBECONFIG_PATH",
+    "CARGO_MAKE_TESTSYS_ARGS",
+    "CARGO_MAKE_TESTSYS_KUBECONFIG_ARG",
```
bcressey: Should we add a filter like this below?
```
key.starts_with("CARGO_MAKE")
```

twoliter/src/tools.rs
```diff
@@ -0,0 +1,138 @@
...
+use crate::tools_hash::TOOLS_HASH;
+use anyhow::{Context, Result};
+use flate2::read::GzDecoder;
+use log::{debug, trace, warn};
+use std::path::Path;
+use tar::Archive;
+
+const TARBALL_DATA: &[u8] = include_bytes!(concat!(env!("OUT_DIR"), "/tools.tar.gz"));
```
bcressey: If there's a decent crate for zstd, that would be better than gzip in terms of compressed size and decompression speed.

twoliter/src/tools.rs
```diff
@@ -0,0 +1,138 @@
...
+
+pub(crate) async fn install_tools(tools_dir: impl AsRef<Path>, force: bool) -> Result<()> {
+    let tools_dir = tools_dir.as_ref();
+    let sentinel_filepath = tools_dir.join("installed");
```
bcressey: nit: could be a hidden file, which tends to imply the "sentinel" semantic:
```suggestion
    let sentinel_filepath = tools_dir.join(".installed");
```

twoliter/src/tools.rs
```diff
@@ -0,0 +1,138 @@
...
+        .await
+        .context("Unable to install tools")?;
+
+    // Write out a file that can be used to detect what version of the tools has been installed.
+    fs::write(&sentinel_filepath, &TOOLS_HASH)
+        .await
+        .context(format!(
+            "Unable to write the tools hash to '{}'",
+            sentinel_filepath.display(),
+        ))?;
```
bcressey: nit: I would tend to make this part of the `unpack_tarball()` contract:
```
unpack_tarball(&tools_dir, &sentinel_filepath)
```

as the unpack function should "know" the hashes of what it's unpacking.

twoliter/src/tools.rs
```diff
@@ -0,0 +1,138 @@
...
+    };
+    trace!("installed hash is '{}'", installed);
+    let do_install = installed != TOOLS_HASH;
+    if do_install {
+        debug!(
+            "Previously installed hash is '{}', our hash '{}'",
+            installed, TOOLS_HASH
+        )
+    }
+    do_install
```
bcressey: I am a bit skeptical of this - I'd prefer to check the hashes of each file unpacked from the tarball to determine whether we need to replace it, rather than relying on the hash-of-hashes in the sentinel file.

It might also be an unnecessary optimization - can we just always unpack the archive? Does it save much time?

twoliter/src/tools.rs
```diff
@@ -0,0 +1,138 @@
...
+    std::fs::remove_file(&tempdir.path().join("Makefile.toml")).unwrap();
+    install_tools(tempdir.path(), false).await.unwrap();
+    // The removed file should still *not* be there.
+    assert!(!tempdir.path().join("Makefile.toml").is_file());
+
+    // Call install_tools with force=true and show that it worked by checking that the removed file
+    // is once again found.
+    install_tools(tempdir.path(), true).await.unwrap();
+    // The removed file is there again because we called install_tools with force=true.
+    assert!(tempdir.path().join("Makefile.toml").is_file());
```
bcressey: Can we get rid of the modality here? It seems like running `install_tools()` should always bring the tools into the expected state.

twoliter/build.rs
```diff
@@ -0,0 +1,87 @@
...
+    let mut buf_writer = Vec::new().writer();
+    let enc = GzEncoder::new(&mut buf_writer, Compression::default());
+    let mut tar = tar::Builder::new(enc);
+    tar.append_dir_all("", &tools_dir).unwrap();
+
+    // Drop tar object to ensure any finalizing steps are done.
+    drop(tar);
+
+    // Get a reference to the tarball bytes.
+    let tar_data = buf_writer.get_ref();
```
bcressey: nit: to indicate that it's been compressed already -
```suggestion
    // Get a reference to the tarball bytes.
    let tar_gz_data = buf_writer.get_ref();
```

twoliter/src/cmd/exec.rs
webern: No, I don't think so. There are a lot of `CARGO_MAKE_*` variables that `cargo make` sets in the environment, possibly for it's own purposes, and I don't think we want to mess with those. I'm not exactly sure if it would behave differently in the presence of the user setting those. In fact I wonder if I should delete `CARGO_MAKE_CARGO_ARGS` but it seems to be working fine.

twoliter/src/tools.rs
webern: It raises a question I have of whether a call to `twoliter install tools` should be required of the user or not. If not, then my goal is to have a one-liner that can be pasted at the top of most commands `install_tools(path, false).await?`.

Perhaps the function could be named better `check_and_install_tools` or something.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/e8aa220e3ce0643cc494c12ad4998f2b76f26a3a..eb2f31c70ca3dd09a81a2a1d63f532b073cbaf7b

- always install tools, do not check for a hash
- use zstd

webern:
> https://github.com/bottlerocket-os/twoliter/compare/eb2f31c70ca3dd09a81a2a1d63f532b073cbaf7b..8b749f085bd581ecf3e50f2b5ad9e235679c3c2c

Remove unused dependencies.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/8b749f085bd581ecf3e50f2b5ad9e235679c3c2c..1581ebdde430623188fd6a4e93e2153a936842b9

Rename `twoliter exec` to `twoliter make`.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/1581ebdde430623188fd6a4e93e2153a936842b9..10cfa41d941ff69692b2e89a4380a5a0abbd1d07

Remove `BOOT_CONFIG` vars from the array of non-conforming vars.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/10cfa41d941ff69692b2e89a4380a5a0abbd1d07..e2b025b1a8c56154d1a5b48b970ac3708ac75802

Rebase to pull in the actual makefile

webern:
> https://github.com/bottlerocket-os/twoliter/compare/e2b025b1a8c56154d1a5b48b970ac3708ac75802..b3b519a24daeb7d88d18db708eb51054558c50e1

Clean up a mistake in the commits.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/b3b519a24daeb7d88d18db708eb51054558c50e1..9415b10064e3b29c31593a42c7ee604166cb7a5a

Use the real, checked-in Makefile.toml

webern:
> https://github.com/bottlerocket-os/twoliter/compare/4c47e441c3b8a194f35bba3fc82c979d1ff1a79d..1666a14d9948be2432fda3b267c4d58bb9d7508a

Use a tempdir for twoliter's tools.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/1666a14d9948be2432fda3b267c4d58bb9d7508a..7f21e5c264ab24a9d61c8a560088a2891f7e943b

Further cleanup now that we use a tempdir.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/7f21e5c264ab24a9d61c8a560088a2891f7e943b..2b687e518da1d9ead560f346df9551dffbc12add

Capture a newly added Makefile variable.

cbgbt (approved):
> Capture a newly added Makefile variable.

I forgot to mention this. Thanks!

twoliter/src/project.rs
```diff
@@ -25,16 +43,26 @@ impl Project {
...
             "Unable to deserialize project file '{}'",
             path.display()
-        ))
+        ))?;
+        project.filepath = path.into();
+        project.project_dir = project
+            .filepath
+            .parent()
+            .context(format!(
+                "Unable to fine the parent directory of '{}'",
```
bcressey:
```suggestion
                "Unable to find the parent directory of '{}'",
```

twoliter/src/tools.rs
```diff
@@ -0,0 +1,41 @@
+use anyhow::{Context, Result};
+use flate2::read::ZlibDecoder;
+use log::debug;
+use std::path::Path;
+use tar::Archive;
+use tempfile::TempDir;
+
+const TAR_ZIP_DATA: &[u8] = include_bytes!(concat!(env!("OUT_DIR"), "/tools.tar.zip"));
```
bcressey: nit: to signal to developers that they don't need to install `unzip` to be able to interact with this artifact if they're trying to debug it for some reason.

```suggestion
const TAR_GZ_DATA: &[u8] = include_bytes!(concat!(env!("OUT_DIR"), "/tools.tar.gz"));
```

bcressey (approved):

webern:
> https://github.com/bottlerocket-os/twoliter/compare/2b687e518da1d9ead560f346df9551dffbc12add..767444b7f4fd2a93b264ee15afe0a9f003fd0356

Use gz instead of zip and fix a spelling error.

PR #53, opened 2023-09-14
cleanup chores

### LINKED ISSUE #38: integrate cargo-deny in ci cd
We need to use cargo-deny to make sure we don't introduce objectionably licensed software to twoliter or the tools.

> If we don't integrate `cargo deny` yet then I'd at least like a backlogged issue for it.

_Originally posted by @bcressey in https://github.com/bottlerocket-os/twoliter/pull/37#pullrequestreview-1594487942_

webern:
*Issue #, if available:*

Closes #38

*Description of changes:*

```
    update-metadata: fix warnings in tests
```

```
    updates and cargo deny

    Update dependencies. In particular, ensure we have only one copy of the
    aws-sdk by updating Tuftool.

    Use cargo deny to check licenses bans and sources.
```

```
    clippy fixes
```

```
    remove infrasys

    Unfortunately, infrasys is unmaintained and unused. This commit removes
    it from the workspace. We can always bring it back if we need it.
```


tools/testsys/src/base64.rs
```diff
@@ -0,0 +1,9 @@
+use base64::alphabet::STANDARD;
+use base64::engine::{GeneralPurpose, GeneralPurposeConfig};
+use base64::Engine;
+
+/// This function became deprecated in the base64 library but it's interface is much simpler than
```
webern:
```suggestion
/// This function became deprecated in the base64 library but its interface is much simpler than
```

Makefile
```diff
@@ -4,3 +4,26 @@ design: ## render design diagrams
...
+.PHONY: build
+build: ## build workspace in release mode
+	cargo build --release --locked
+
+.PHONY: test
+test: ## test workspace
+	cargo test --release --locked
+
+.PHONY: check
+check: fmt clippy build test
```
jpculp: It seems off to me for **build** to be included under **check**. Would it be better to remove **build** from there and add **check** as a pre-req of **build**? If someone wanted to build without the additional checks they can do a `cargo build` instead of a `make build`.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/2bf96152ad1feb01d5edcc4bd12c23ccaebb125b..fd5ea09c22642ff23f4cf627d6c73fae99796643

Re-organize make file per @jpculp's suggestion

jpculp (approved):

cbgbt (approved):

PR #89, opened 2023-09-29
add sdk to project schema

### LINKED ISSUE #83: alpha: take sdk version from twoliter.toml
## Details

For projects that have no dependencies on external kits (such as the Bottlerocket project itself), we should get the SDK version from Twoliter.toml. These fields should be optional in the schema since projects that depend on external kits are going to want to build with the same SDK as the kits they depend on.

## Description of Done

When Twoliter respects and uses the SDK version it finds in `Twoliter.toml`.

webern:
*Issue #, if available:*

Closes #83

*Description of changes:*

```
Add the ability to specify the desired Bottlerocket SDK in
Twoliter.toml.
```

```
Require the SDK to be specified in Twoliter.toml when using twoliter
make.
```

*Testing*

I used this version of Twoliter without changing anything in Bottlerocket and got the expected error:

```
Error: The environment variable 'BUILDSYS_SDK_NAME' can no longer be used. Specify the SDK in Twoliter.toml
[cargo-make] ERROR - Error while executing command, exit code: 1
```

I then removed these vars from Bottlerocket's Makefile.toml and got the expected error:

```
Error: When using twoliter make, it is required that the SDK be specified in Twoliter.toml
```

I added the SDK to Bottlerocket's Twoliter.toml and the build succeeded and the correct SDK was used:

```
v0.34.1: Pulling from bottlerocket/bottlerocket-sdk-x86_64
```	

### DIFF
--- a/twoliter/src/project.rs
+++ b/twoliter/src/project.rs
@@ -1,12 +1,17 @@
-use crate::docker::{ImageArchUri, DEFAULT_REGISTRY, DEFAULT_SDK_NAME, DEFAULT_SDK_VERSION};
+use crate::docker::ImageArchUri;
 use anyhow::{ensure, Context, Result};
 use async_recursion::async_recursion;
 use log::{debug, trace};
+use non_empty_string::NonEmptyString;
 use serde::de::Error;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use std::fmt;
 use std::path::{Path, PathBuf};
 use tokio::fs;
+
+const DEFAULT_REGISTRY: &str = "public.ecr.aws/bottlerocket";
+const DEFAULT_SDK_NAME: &str = "bottlerocket-sdk";
+const DEFAULT_SDK_VERSION: &str = "v0.34.1";
 
 /// Common functionality in commands, if the user gave a path to the `Twoliter.toml` file,
 /// we use it, otherwise we search for the file. Returns the `Project` and the path at which it was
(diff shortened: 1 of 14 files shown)


twoliter/src/project.rs
```diff
@@ -90,17 +99,39 @@ impl Project {
...
+    pub(crate) fn toolchain(&self, arch: &str) -> Option<ImageArchUri> {
+        self.sdk_name().map(|s| s.uri("toolchain", arch))
+    }
 }
 
+/// A base name for an image that can be suffixed using a naming convention. For example,
+/// `registry=public.ecr.aws/bottlerocket`, `name=bottlerocket`, `version=v0.50.0` can be suffixed
+/// via naming convention to produce:
+/// - `registry=public.ecr.aws/bottlerocket/bottlerocket-sdk-x86_64:v0.50.0`
+/// - `registry=public.ecr.aws/bottlerocket/bottlerocket-toolchain-aarch64:v0.50.0`
```
bcressey, line 128: I'd like to get rid of the "magic" implicit suffixing if we can since ultimately that constrains the end user in kind of an arbitrary way.

What do you think about making the name a template like `bottlerocket-sdk-{{ arch }}`? That's consistent with similar patterns elsewhere in the project.

I would also recommend making the toolchain image its own explicit field, e.g. `toolchain_name` . But toolchain (and kmod kits) are just inherently problematic so we'll need to find a different way to deal with them at some point.

webern: I thought about this and didn't like the suffixing when I encountered it. I'm fine with changing to the non-magical version. It's not great to have a field in the schema that needs to be removed at some point. Any idea of a way to avoid having a schema key that is effectively temporary?

webern:
> https://github.com/bottlerocket-os/twoliter/compare/9dff50a60b80cb1651b6326cc31df12f68591580..53abc2404fd5d4a7d92b70595f8f8e1073e834b0

Separate the specification of the `sdk` and `toolchain` containers.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/53abc2404fd5d4a7d92b70595f8f8e1073e834b0..af96ef7e0c441ddbf80326b61748d3aa88f277ed

Add `toolchain` to the other test project file.

twoliter/src/project.rs
bcressey: It feels like the correct fix is to have the SDK always build a "native" toolchain, possibly in addition to the "cross" toolchain, so that the SDK itself always has the archive and we don't need to keep a second container in sync. We do something like this for Rust already, for different reasons.

I've opened https://github.com/bottlerocket-os/bottlerocket-sdk/issues/137 for that.

> Any idea of a way to avoid having a schema key that is effectively temporary?

Not really. We could have a fallback path, like "if the toolchain key isn't set, assume the SDK has the toolchain and use it instead".

bcressey (approved):

ecpullen (approved):

Issue #82, opened 2023-09-27
alpha: twoliter build variant can create a custom variant from a twoliter project


webern:
## Description

This represents the completion of `twoliter build variant` in its "alpha" incarnation.

## Definition of Done

`twoliter build variant` can create an image with custom packages (without using kits).

PR #97, opened 2023-10-06
twoliter: Dedup cargo make logic

ecpullen:
*Issue #, if available:*

Part of the work on #82

*Description of changes:*

We will be calling cargo make from different CLI commands, so we want the cargo make invocation to be re-usable.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -23,4 +23,5 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - run: cargo install cargo-deny
+      - run: cargo install cargo-make
       - run: make build
--- a/twoliter/src/cargo_make.rs
+++ b/twoliter/src/cargo_make.rs
@@ -1,0 +1,260 @@
+use crate::common::exec_log;
+use crate::docker::ImageArchUri;
+use crate::project::Project;
+use anyhow::{bail, Result};
+use log::trace;
+use std::path::PathBuf;
+use tokio::process::Command;
+
+/// A struct used to invoke `cargo make` tasks with `twoliter`'s `Makefile.toml`.
+/// ```rust
+/// # use crate::project::Project;
+/// # use crate::test::data_dir;
+/// # use self::CargoMake;
+/// # let project_path = data_dir().join("Twoliter-1.toml");
+/// # let makefile_path = data_dir().join("Makefile.toml");
+/// # let project_dir = data_dir();
+///
+/// // First create a twoliter project.
+/// let project = Project::load(project_path).await.unwrap();
+/// // Add the architecture that cargo make will be invoked for.
+/// // This is required so that the correct sdk and toolchain are selected.
+/// let arch = "x86_64";
+///
+/// // Create the `cargo make` command.
+/// let cargo_make_command = CargoMake::new(&project, arch)
+///     .unwrap()
+///     // Specify path to the `Makefile.toml` (Default: `Makefile.toml`)
+///     .makefile(makefile_path)
+///     // Specify the project directory (Default: `.`)
+///     .project_dir(project_dir)
+///     // Add environment variable to the command
+///     .env("FOO", "bar")
+///     // Add cargo make arguments such as `-q` (quiet)
+///     ._arg("-q");
+///
+/// // Run the `cargo make` task
+/// cargo_make_command.clone()
+///     ._exec("verify-twoliter-env")
+///     .await
+///     .unwrap();
+///
+/// // Run the `cargo make` task with args
+/// cargo_make_command
+///     .exec_with_args("verify-env-set-with-arg", ["FOO"])
+///     .await
+///     .unwrap();
+/// ```
+#[derive(Debug, Clone, Default)]
+pub struct CargoMake {
+    makefile_path: Option<PathBuf>,
+    project_dir: Option<PathBuf>,
+    args: Vec<String>,
+}
+
+impl CargoMake {
+    /// Create a new `cargo make` command. The sdk and toolchain environment variables will be set
+    /// based on the definitions in `Twoliter.toml` and `arch`.
+    pub(crate) fn new<S>(project: &Project, arch: S) -> Result<Self>
+    where
+        S: Into<String>,
+    {
+        let (sdk, toolchain) = require_sdk(project, &arch.into())?;
+        Ok(Self::default()
+            .env("TLPRIVATE_SDK_IMAGE", sdk)
+            .env("TLPRIVATE_TOOLCHAIN", toolchain))
+    }
+
+    /// Specify the path to the `Makefile.toml` for the `cargo make` command
+    pub(crate) fn makefile<P>(mut self, makefile_path: P) -> Self
+    where
+        P: Into<PathBuf>,
+    {
+        self.makefile_path = Some(makefile_path.into());
+        self
+    }
+
+    /// Specify the project directory for the `cargo make` command
+    pub(crate) fn project_dir<P>(mut self, project_dir: P) -> Self
+    where
+        P: Into<PathBuf>,
+    {
+        self.project_dir = Some(project_dir.into());
+        self
+    }
+
+    /// Specify environment variables that should be applied for this comand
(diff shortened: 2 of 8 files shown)


ecpullen:
^ Add testing for all `CargoMake` functions.

twoliter/src/cargo_make.rs
```diff
@@ -0,0 +1,213 @@
...
+pub struct CargoMake {
+    makefile_path: Option<PathBuf>,
+    project_dir: Option<PathBuf>,
+    args: Vec<String>,
+}
+
+impl CargoMake {
+    pub fn new<S>(project: &Project, arch: S) -> Result<Self>
+    where
+        S: Into<String>,
```
webern, line 60: I keep trying Into<String> and regretting it because more things are` AsRef<str>` than `Into<String>` and for some reason there is no blanket impl that goes from `AsRef<str>` to `Into<String>`.

```suggestion
        S: AsRef<str>,
```

twoliter/src/cargo_make.rs
```diff
@@ -0,0 +1,213 @@
...
+    {
+        let (sdk, toolchain) = require_sdk(project, &arch.into())?;
+        Ok(Self::default()
+            .env("TLPRIVATE_SDK_IMAGE", sdk)
+            .env("TLPRIVATE_TOOLCHAIN", toolchain))
+    }
+
+    pub fn makefile<P>(mut self, makefile_path: P) -> Self
```
webern: Acknowledging that you may have copied this from me...  In case we do make a lib at some point, all things should start off as pub(crate). Then we would make certain things pub only if a lib were introduced.

```suggestion
    pub(crate) fn makefile<P>(mut self, makefile_path: P) -> Self
```

twoliter/src/cargo_make.rs
```diff
@@ -0,0 +1,213 @@
...
+        self.args.extend(args.into().into_iter().map(Into::into));
+        self
+    }
+
+    pub async fn _exec<S>(&self, task: S) -> Result<()>
+    where
+        S: Into<String>,
+    {
+        self.exec_with_args(task, Vec::<String>::new()).await
+    }
```
webern: Acknowledging that you copied this from me... we should remove unused functions (unless you are sure you are going to use them later). 

This one feels unnecessary either way since it is easy enough to pass `Vec::new()`, or `[]` when you have no args. Also see below, I think `IntoIter<Item = String>` might work without the type annotation.

```suggestion
```

twoliter/src/cargo_make.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    pub async fn exec_with_args<S1, S2, V>(&self, task: S1, args: V) -> Result<()>
```
webern: Also something like `IntoIter<Item = String>` seems more common.

```suggestion
    /// Blah blah blah, explain what this does.
    pub(crate) async fn exec<S1, S2, V>(&self, task: S1, args: V) -> Result<()>
```

twoliter/src/cargo_make.rs
```diff
@@ -0,0 +1,213 @@
...
+        (Some(s), Some(t)) => Ok((s, t)),
+        _ => bail!(
+            "When using twoliter make, it is required that the SDK and toolchain be specified in \
+            Twoliter.toml"
+        ),
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct CargoMake {
```
webern, line 49: Please document this. There is a lot to say about what this command is doing.

twoliter/src/cargo_make.rs
```diff
@@ -0,0 +1,213 @@
...
+
+use anyhow::{bail, Result};
+use log::trace;
+use tokio::process::Command;
+
+use crate::common::exec;
+use crate::docker::ImageArchUri;
+use crate::project::Project;
+
+fn require_sdk(project: &Project, arch: &str) -> Result<(ImageArchUri, ImageArchUri)> {
```
webern: Put this below with the other private functions like `is_build_system_env`

twoliter/src/cargo_make.rs
```diff
@@ -0,0 +1,213 @@
...
+
+impl CargoMake {
+    pub fn new<S>(project: &Project, arch: S) -> Result<Self>
+    where
+        S: Into<String>,
+    {
+        let (sdk, toolchain) = require_sdk(project, &arch.into())?;
+        Ok(Self::default()
+            .env("TLPRIVATE_SDK_IMAGE", sdk)
+            .env("TLPRIVATE_TOOLCHAIN", toolchain))
```
webern, line 65: I'm thinking about whether we should just hang on to `&'a Project` and do things like this during the exec call. I'm also slightly concerned about lack of clarity regarding what args could be passed during `exec` and which are managed by this object.

These are not huge concerns. Just some things to consider and change if it feels like we can do better.

twoliter/src/cargo_make.rs
ecpullen: Since we are taking ownership of each `String` it makes since to use `Into<String>`. We can change this if it's critical, but I believe `Into<String>` is the correct type here.

twoliter/src/cargo_make.rs
ecpullen: I'm keeping the `_*` functions until the build pr has been drafted.

twoliter/src/cargo_make.rs
```diff
@@ -0,0 +1,262 @@
+use std::path::PathBuf;
+
+use anyhow::{bail, Result};
+use log::trace;
+use tokio::process::Command;
+
+use crate::common::exec_log;
+use crate::docker::ImageArchUri;
+use crate::project::Project;
```
webern: Nit: we typically let rustfmt sort these.

```suggestion
use anyhow::{bail, Result};
use crate::common::exec_log;
use crate::docker::ImageArchUri;
use crate::project::Project;
use log::trace;
use std::path::PathBuf;
use tokio::process::Command;
```

twoliter/src/cargo_make.rs
```diff
@@ -0,0 +1,262 @@
...
+#[derive(Debug, Clone, Default)]
+pub struct CargoMake {
+    makefile_path: Option<PathBuf>,
+    project_dir: Option<PathBuf>,
+    args: Vec<String>,
+}
+
+impl CargoMake {
+    /// Create a new `cargo make` command. The sdk and toolchain environment variables will be set
+    /// based on the definitions in `Twoliter.toml` and the host arch.
```
webern:
```suggestion
    /// based on the definitions in `Twoliter.toml` and `arch`.
```

twoliter/src/cargo_make.rs
```diff
@@ -0,0 +1,262 @@
...
+    pub fn new<S>(project: &Project, arch: S) -> Result<Self>
```
webern:
```suggestion
    pub(crate) fn new<S>(project: &Project, arch: S) -> Result<Self>
```

twoliter/src/cargo_make.rs
webern: Follow up #104

twoliter/src/test/mod.rs
```diff
@@ -4,7 +4,7 @@ This directory and module are for tests, test data, and re-usable test code. Thi
 be compiled for `cfg(test)`, which is accomplished at its declaration in `main.rs`.
 
 !*/
-
+pub mod cargo_make;
```
webern: I think it can be private.

```suggestion
mod cargo_make;
```

webern (approved):

.github/workflows/rust.yml
```diff
@@ -23,4 +23,5 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - run: cargo install cargo-deny
+      - run: cargo install cargo-make
```
bcressey, line 26: nit: the changes to `cargo_make.rs` should probably be in the first commit, rather than the GH workflow one.

pedantic nit: I'd also prefix this commit with "github" vs "git"

bcressey (approved):

webern (changes requested):
Ethan said he found a bug. Marking this as do-not-merge.

webern (approved):

PR #96, opened 2023-10-05
Twoliter build variant alpha

ecpullen:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #74, opened 2023-09-20
tracking issue: alpha milestone


webern:
## Description

Create a `twoliter build variant` command that simulates how Twoliter will eventually operate. We will skip the implementation of kits for this. Instead we will
- Build Bottlerocket.
- Copy all of the RPMs into an SDK image.
- Build a Twoliter project (using the RPMs from the SDK image if/when they are needed as build dependencies)
- Add the Twoliter project's RPMs to the collection of Bottlerocket RPMs
- Run rpm2img and install both Bottlerocket's packages and the Twoliter packages to create a custom variant image

Items:

- [x] #72
- [x] #83
- [x] #77
- [x] #78
- [x] #79
- [x] #84
- [x] #80
- [x] #82

## Definition of Done

When we can create a custom variant with custom packages on top of Bottlerocket's RPMs using `twoliter build variant`.

webern:
this is don!

PR #99, opened 2023-10-06
delete dead code: docker build and twoliter container

webern:
*Issue #, if available:*

Related #82, #74

*Description of changes:*

We are proceeding along different lines for now where we do not need to create a "Twoliter bootstrapping container". We may return to this strategy later, in which case this code may be revived. But for now it is going to be dead code.

*Testing*

`twoliter build variant` was a dead command. This strips out all of the code it was using, to be replaced by #96.
The build works, and there is nothing to test.


bcressey (approved):

cbgbt (approved):

webern:
> https://github.com/bottlerocket-os/twoliter/compare/6c43c62c494c9206dd01afc223beefdda8e27df4..21489afe34bb2b56cf9f20bd82d6e51fd057298f

Rebase.

PR #102, opened 2023-10-09
switch to fixed directory for twoliter tools

bcressey:
**Issue #, if available:**
N/A

**Description of changes:**
I've been experimenting with a tool to forward file descriptors into a build, which requires the ability to run something larger than a shell script during the Docker execution. That hits the limitation of secrets which can't be larger than 500 KiB.

Switching to a fixed install location under `build/tools` avoids this limitation and makes the code a bit more intuitive. It's also relatively easy to see what `twoliter` has installed.

Because the tools need to be passed in the build context, I've also imported the current `.dockerignore` file from the core repo and renamed it to `Dockerfile.dockerignore`, so that [it will be used by the BuildKit backend](https://docs.docker.com/engine/reference/commandline/build/#use-a-dockerignore-file).

**Testing done:**
Built some variants successfully. I also confirmed through various hacks that the `.dockerignore` file was being respected. For example, excluding `/build/rpms/*.rpm` caused the expected build failure.

**Terms of contribution:**

### DIFF
--- a/twoliter/build.rs
+++ b/twoliter/build.rs
@@ -20,6 +20,7 @@ fn main() {
     let paths = Paths::new();
     println!("cargo:rerun-if-changed={}", paths.data_input_dir.display());
 
+    let _ = fs::remove_dir_all(&paths.prep_dir);
     fs::create_dir_all(&paths.prep_dir).expect(&format!(
         "Unable to create directory '{}'",
         paths.prep_dir.display()
--- a/twoliter/build.rs
+++ b/twoliter/build.rs
@@ -26,6 +26,7 @@ fn main() {
     ));
 
     paths.copy_file("Dockerfile");
+    paths.copy_file("Dockerfile.dockerignore");
     paths.copy_file("Makefile.toml");
     paths.copy_file("docker-go");
     paths.copy_file("partyplanner");
--- a/twoliter/src/cmd/make.rs
+++ b/twoliter/src/cmd/make.rs
@@ -1,9 +1,10 @@
 use crate::cargo_make::CargoMake;
-use crate::project;
-use crate::tools::{install_tools, tools_tempdir};
+use crate::project::{self};
+use crate::tools::install_tools;
 use anyhow::Result;
 use clap::Parser;
 use std::path::PathBuf;
+use tokio::fs;
 
 /// Run a cargo make command in Twoliter's build environment. Known Makefile.toml environment
 /// variables will be passed-through to the cargo make invocation.
@@ -33,12 +34,13 @@ pub(crate) struct Make {
 impl Make {
     pub(super) async fn run(&self) -> Result<()> {
         let project = project::load_or_find_project(self.project_path.clone()).await?;
-        let tempdir = tools_tempdir()?;
-        install_tools(&tempdir).await?;
-        let makefile_path = tempdir.path().join("Makefile.toml");
+        let toolsdir = project.project_dir().join("build/tools");
+        fs::remove_dir_all(&toolsdir).await?;
+        install_tools(&toolsdir).await?;
+        let makefile_path = toolsdir.join("Makefile.toml");
         CargoMake::new(&project, &self.arch)?
             .env("CARGO_HOME", self.cargo_home.display().to_string())
-            .env("TWOLITER_TOOLS_DIR", tempdir.path().display().to_string())
+            .env("TWOLITER_TOOLS_DIR", toolsdir.display().to_string())
             .env("BUILDSYS_VERSION_IMAGE", project.release_version())
             .makefile(makefile_path)
             .project_dir(project.project_dir())
--- a/twoliter/src/tools.rs
+++ b/twoliter/src/tools.rs
@@ -1,9 +1,9 @@
 use anyhow::{Context, Result};
+use filetime::{set_file_handle_times, set_file_mtime, FileTime};
 use flate2::read::ZlibDecoder;
 use log::debug;
 use std::path::Path;
 use tar::Archive;
-use tempfile::TempDir;
 use tokio::fs;
 use tokio::io::AsyncWriteExt;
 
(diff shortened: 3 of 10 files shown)


twoliter/src/cmd/make.rs
```diff
@@ -34,9 +33,11 @@ pub(crate) struct Make {
 impl Make {
     pub(super) async fn run(&self) -> Result<()> {
         let project = project::load_or_find_project(self.project_path.clone()).await?;
-        let tempdir = tools_tempdir()?;
-        install_tools(&tempdir).await?;
-        let makefile_path = tempdir.path().join("Makefile.toml");
+        let toolsdir = project.project_dir().join("build/tools");
```
webern, line 37: Maybe the project object should provide such things (build_dir, tools_dir, etc)

```suggestion
        let toolsdir = project.tools_dir();
```

twoliter/src/cmd/make.rs
```diff
@@ -34,9 +33,11 @@ pub(crate) struct Make {
...
+        let _ = std::fs::remove_dir_all(&toolsdir);
+        std::fs::create_dir_all(&toolsdir)?;
```
webern: Use `tokio`. Also I don't think even `tokio` gives a helpful error message. I have planned to add this to `common.rs` for convenience. The idea would be to move error message creation somewhere else.

```rust
mod fs {
    pub(crate) async fn create_dir_all(path: impl AsRef<Path>) -> Result<()> {
        tokio::fs::create_dir_all(path.as_ref()).context(format!("Unable to create directory '{}'", path.as_ref().display());
    }
}

```

twoliter/src/cmd/make.rs
```diff
@@ -34,9 +33,11 @@ pub(crate) struct Make {
 impl Make {
     pub(super) async fn run(&self) -> Result<()> {
```
webern, line 35: @ecpullen these are changes to the code that you are moving.

twoliter/src/tools.rs
```diff
@@ -16,36 +16,38 @@ const PUBSYS_SETUP: &[u8] = include_bytes!(env!("CARGO_BIN_FILE_PUBSYS_SETUP"));
...
-    write_bin("pubsys-setup", PUBSYS_SETUP, &dir).await?;
-    write_bin("testsys", TESTSYS, &dir).await?;
-    write_bin("tuftool", TUFTOOL, &dir).await?;
-
     // Write out the embedded tools and scripts.
     unpack_tarball(&dir)
         .await
         .context("Unable to install tools")?;
 
+    // Pick one of the embedded files for use as the canonical mtime.
```
webern, line 31: Why do we need a canonical mtime? Is this preventing docker rebuilds?

bcressey: > Why do we need a canonical mtime? Is this preventing docker rebuilds?

I'll add a comment to the code. It's a minor optimization to make the build context more efficient by not always copying the tools through.

For example on a local build with a 180 MiB file:
```
❯ docker build --progress=plain .
...
#10 [internal] load build context
#10 sha256:8fe1ebd2168d7202972f3fed513e650e6bc14c1252ef426d770005b1b769b70b
#10 transferring context: 183.54MB 0.5s done
#10 DONE 0.6s

❯ docker build --progress=plain .
...
#10 [internal] load build context
#10 sha256:942722aa20dadaa1f21a60c65ffa04d801e48ac344603e42b368fda45457366a
#10 transferring context: 291B done
#10 DONE 0.0s

❯ touch -r Dockerfile NVIDIA-Linux-aarch64-470.82.01.run

❯ docker build --progress=plain .
...
#10 [internal] load build context
#10 sha256:8fe1ebd2168d7202972f3fed513e650e6bc14c1252ef426d770005b1b769b70b
#10 transferring context: 183.54MB 0.5s done
#10 DONE 0.6s

❯ docker build --progress=plain .
...
#10 [internal] load build context
#10 sha256:ec46d1475e765e302cc924f8682c05c897c1a56df7a3b356bae7b1b866eb4c74
#10 transferring context: 291B done
#10 DONE 0.0s
```

The build context is more intelligent than a Docker layer in the sense that modifying a single file results in only that file being transferred, not the entire build context. But it still seems worth avoiding unnecessary copies.

The direction I'm exploring with passing file descriptors could let us get rid of all the changes to the build context. `/packages/*/*.tar.*`, `/build/rpms`, `/.gomodcache`, and `/.cargo` are the current offenders. I'd like to avoid others.

twoliter/build.rs
```diff
@@ -20,12 +20,14 @@ fn main() {
     let paths = Paths::new();
     println!("cargo:rerun-if-changed={}", paths.data_input_dir.display());
 
+    let _ = fs::remove_dir_all(&paths.prep_dir);
```
ecpullen, line 23: If we can avoid updating build.rs, that would be great. The entire file will be changed in an upcoming pr.

ecpullen (approved):

twoliter/build.rs
webern: > If we can avoid updating build.rs, that would be great. The entire file will be changed in an upcoming pr.

The addition of a file appears to be necessary for this PR, that is `Dockerfile.dockerignore`, so at least one change in `build.rs` is not optional. Anyway, the changes here should be easy to replay. Delete the prep dir and add a file.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/c05d9e1e24afb6b223b93234c68bda340400619d..c08da4cf3730fe54f616cee4b5bec52eef130fd2

I think this rebase is correct.

webern (approved):
I can follow up with some of the refactoring things I mentioned.

webern:
Rebased (twice because of a mistake):

- https://github.com/bottlerocket-os/twoliter/compare/c08da4cf3730fe54f616cee4b5bec52eef130fd2..c0ef5c4dc9509f2fe1859199257305f3a48023d0
- https://github.com/bottlerocket-os/twoliter/compare/c0ef5c4dc9509f2fe1859199257305f3a48023d0..d7aba2510e637df589ffdce5054abc0957d789ac

twoliter/src/cmd/make.rs
```diff
@@ -1,9 +1,10 @@
 use crate::cargo_make::CargoMake;
-use crate::project;
-use crate::tools::{install_tools, tools_tempdir};
+use crate::project::{self};
```
webern, line 2:
```suggestion
use crate::project;
```

webern:
I am merging this. I've been bitten for days with `rpm2img` changes not being reflected in the docker secret mount. It seems like docker is using a cached version of the file for the secret instead of the newly written-out changed version. I'm assuming this would fix the problem and we have two green check marks, so YOLO merging.

PR #106, opened 2023-10-11
twoliter: Create a test project

### LINKED ISSUE #72: kits: create a test project
## Description

We need a small project, quick to build, checked-in to the twoliter git repository, that can be used to test `build kit` and `build variant`. The "bottlerocket" image it creates will not be usable for anything, but it should install a few RPM packages from kits.

Perhaps at `twoliter/tests/projects/project-1`.

## Definition of Done

When a test project exists in the repository.

ecpullen:
**Issue number:**

Closes #72 

**Description of changes:**

Adds a new test project for OOTB. There are 2 packages included, one written in rust and the other in go.

**Testing done:**

This has been tested with #96 

**Terms of contribution:**


tests/projects/project1/packages/build.rs
```diff
@@ -0,0 +1,9 @@
+use std::process::{exit, Command};
+
+fn main() -> Result<(), std::io::Error> {
+    let ret = Command::new("buildsys").arg("build-package").status()?;
+    if !ret.success() {
+        exit(1);
+    }
+    Ok(())
+}
```
bcressey, line 9: Should we inject these files via `twoliter`? It'd be somewhat magical but would cut down on boilerplate.

tests/projects/project1/.dockerignore
```diff
@@ -0,0 +1,10 @@
+/.git
+/.gomodcache
+/build/*
+!/build/rpms/
+/build/rpms/*
```
bcressey, line 5: I don't love copying this file around but I'll clean it up in #102 assuming this merges first.

bcressey (approved):

tests/projects/project1/packages/build.rs
webern: I don't love the idea because someone might want to augment this code for their own purposes (pulling from a cache or generating some code, for example). On the other hand, without controlling this code, it is hard to update the calling pattern without breaking everyone.

One idea might be to vend a crates.io project that simplifies this down to:

```rust
fn main() -> Result<(), Error> {
    buildsys::run()
}
```

This would also eliminate the need to vend buildsys as a binary. But, there would be a strict calling convention based on environment variables and some sort of version check to make sure the customer isn't using a too-old version of buildsys for Twoliter's expectations.

So Twoliter might pass `BUILDSYS_REQUIRED_VERSION=v0.22` and buidsys might error `Buildsys version v0.10.0 is not compatible with the current version of Twoliter, please update to version v0.22.0`.

webern (approved):

cbgbt (approved):

cbgbt (approved):

Issue #88, opened 2023-09-28
dynamic gomodules


webern:
TODO - write this issue.

GOMODULES are kind of hardcoded in Makefile.toml
Might not be an issue? But maybe we should discover them?

Ben's suggestion from https://github.com/bottlerocket-os/twoliter/pull/108#discussion_r1377922765

> Potentially we want a contract like:
>
> Twoliter will search the sources path for Cargo.lock and go.mod files and download the referenced Rust crates and Go modules prior to package builds.
>
> And eventually support for .twoliterignore files to suppress this behavior by excluding certain paths.

ecpullen:
This was completed for alpha build in #108

PR #108, opened 2023-10-26
twoliter: `twoliter build` alpha

### LINKED ISSUE #79: alpha: copy SDK rpms to the project build dir
## Description

Before starting the build, `twoliter build variant` (in it's alpha incarnation) needs to copy `twoliter-alpha/build/packages` to `build/packages` in the project that is about to be built. This way when `rpm2img` happens, all of the core Bottlerocket packages will be there for installation.

## Definition of Done

When `twoliter build variant` copies rpms from the SDK to `build/rpms`.

### LINKED ISSUE #80: bottlerocket license file copy in rpm2img is problematic
This is an exact duplicate of https://github.com/bottlerocket-os/bottlerocket/issues/3177 since the solution touches both GitHub repositories.

### Description

The script, `rpm2img` assumes that we are running in the Bottlerocket monorepo.
During the script execution it copies `LICENSE-MIT` directly from the git repo.
This is problematic because an out-of-tree build may not have these files.

Instead, we should package licenses into an RPM file during the package build step and install them from the RPM package during rpm2img.

In other words, we can no longer do this:
https://github.com/bottlerocket-os/bottlerocket/blob/08497fd1503397a0e02858be10917c34965b7ca8/tools/rpm2img#L233

### Definition of Done

When the build no longer relies on Bottlerocket's license files being being present during rpm2img.
```
id: br_licenses
epic: twoliter_exec
```

### LINKED ISSUE #82: alpha: twoliter build variant can create a custom variant from a twoliter project

### LINKED ISSUE #104: Cleanup Unused CargoMake functions
Acknowledging that you copied this from me... we should remove unused functions (unless you are sure you are going to use them later). 

This one feels unnecessary either way since it is easy enough to pass `Vec::new()`, or `[]` when you have no args. Also see below, I think `IntoIter<Item = String>` might work without the type annotation.

```suggestion
```

_Originally posted by @webern in https://github.com/bottlerocket-os/twoliter/pull/97#discussion_r1350547306_
            

Definition of done:

When the `CargoMake` object does not have unused functions and does not need underscore prefixes to silence warnings.

ecpullen:
**Issue number:**

Closes #104 
Closes #82
Closes #80
Closes #79

**Description of changes:**

**Testing done:**

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/Makefile.toml
+++ b/twoliter/embedded/Makefile.toml
@@ -97,7 +97,8 @@ BUILDSYS_JOBS = "8"
 CARGO_HOME = "${BUILDSYS_ROOT_DIR}/.cargo"
 # This needs to end with pkg/mod so that we can mount the parent of pkg/mod as GOPATH.
 GO_MOD_CACHE = "${BUILDSYS_ROOT_DIR}/.gomodcache/pkg/mod"
-GO_MODULES = "ecs-gpu-init host-ctr"
+# Dynamically load a list of go modules from ${BUILDSYS_SOURCE_DIR}
+GO_MODULES = { script = ['find ${BUILDSYS_SOURCES_DIR} -name go.mod -type f -printf "%h\n" | xargs -n1 basename'] }
 DOCKER_BUILDKIT = "1"
 
 # This is the filename suffix for operations that write out AMI information to
--- a/twoliter/src/cmd/build.rs
+++ b/twoliter/src/cmd/build.rs
@@ -41,6 +41,7 @@ pub(crate) struct BuildVariant {
 impl BuildVariant {
     pub(super) async fn run(&self) -> Result<()> {
         let project = project::load_or_find_project(self.project_path.clone()).await?;
+        let token = project.token();
         let tempdir = tools_tempdir()?;
         install_tools(&tempdir).await?;
         let makefile_path = tempdir.path().join("Makefile.toml");
@@ -51,7 +52,7 @@ impl BuildVariant {
         fs::create_dir_all(&packages_dir)?;
 
         let sdk_container = DockerContainer::new(
-            "sdk",
+            format!("sdk-{}", token),
             project
                 .sdk(&self.arch)
                 .context(format!(
--- a/twoliter/src/cmd/build.rs
+++ b/twoliter/src/cmd/build.rs
@@ -1,6 +1,14 @@
-use anyhow::Result;
+use crate::cargo_make::CargoMake;
+use crate::docker::DockerContainer;
+use crate::project;
+use crate::tools::{install_tools, tools_tempdir};
+use anyhow::{Context, Result};
 use clap::Parser;
-use std::path::PathBuf;
+use log::debug;
+use std::fs;
+use std::path::{Path, PathBuf};
+use tempfile::TempDir;
+use tokio::fs::{remove_dir_all, remove_file};
 
 #[derive(Debug, Parser)]
 pub(crate) enum BuildCommand {
@@ -25,10 +33,94 @@ pub(crate) struct BuildVariant {
     /// The architecture to build for.
     #[clap(long = "arch", default_value = "x86_64")]
     arch: String,
+
+    /// The variant to build.
+    variant: String,
 }
 
 impl BuildVariant {
     pub(super) async fn run(&self) -> Result<()> {
-        Ok(())
+        let project = project::load_or_find_project(self.project_path.clone()).await?;
+        let tempdir = tools_tempdir()?;
+        install_tools(&tempdir).await?;
+        let makefile_path = tempdir.path().join("Makefile.toml");
+        // A temporary directory in the `build` directory
+        let build_temp_dir = TempDir::new_in(project.project_dir())
+            .context("Unable to create a tempdir for Twoliter's build")?;
+        let packages_dir = build_temp_dir.path().join("sdk_rpms");
+        fs::create_dir_all(&packages_dir)?;
+
+        let sdk_container = DockerContainer::new(
+            "sdk",
+            project
+                .sdk(&self.arch)
+                .context(format!(
+                    "No SDK defined in {} for {}",
+                    project.filepath().display(),
+                    &self.arch
+                ))?
+                .uri(),
+        )
+        .await?;
+        sdk_container
+            .cp_out(Path::new("twoliter/alpha/build/rpms"), &packages_dir)
+            .await?;
+
+        let rpms_dir = project.project_dir().join("build").join("rpms");
+        fs::create_dir_all(&rpms_dir)?;
+        debug!("Moving rpms to build dir");
+        for maybe_file in fs::read_dir(packages_dir.join("rpms"))? {
+            let file = maybe_file?;
+            debug!("Moving '{}'", file.path().display());
+            fs::rename(file.path(), rpms_dir.join(file.file_name()))?;
+        }
+
+        let mut created_files = Vec::new();
+
+        let sbkeys_dir = project.project_dir().join("sbkeys");
+        if !sbkeys_dir.is_dir() {
+            // Create a sbkeys directory in the main project
+            debug!("sbkeys dir not found. Creating a temporary directory");
+            fs::create_dir_all(&sbkeys_dir)?;
+            sdk_container
+                .cp_out(
+                    Path::new("twoliter/alpha/sbkeys/generate-local-sbkeys"),
+                    &sbkeys_dir,
+                )
+                .await?;
+        };
+
+        // TODO: Remove once models is no longer conditionally compiled.
+        // Create the models directory for the sdk to mount
+        let models_dir = project.project_dir().join("sources/models");
+        if !models_dir.is_dir() {
+            debug!("models source dir not found. Creating a temporary directory");
+            fs::create_dir_all(&models_dir.join("src/variant"))
+                .context("Unable to create models source directory")?;
+            created_files.push(models_dir)
+        }
+
+        // Hold the result of the cargo make call so we can clean up the project directory first.
+        let res = CargoMake::new(&project, &self.arch)?
+            .env("TWOLITER_TOOLS_DIR", tempdir.path().display().to_string())
+            .env("BUILDSYS_ARCH", &self.arch)
+            .env("BUILDSYS_VARIANT", &self.variant)
+            .env("BUILDSYS_SBKEYS_DIR", sbkeys_dir.display().to_string())
+            .makefile(makefile_path)
+            .project_dir(project.project_dir())
+            .exec("build")
+            .await;
+
+        // Clean up all of the files we created
+        for file_name in created_files {
+            let added = Path::new(&file_name);
+            if added.is_file() {
+                remove_file(added).await?;
+            } else if added.is_dir() {
+                remove_dir_all(added).await?;
+            }
+        }
+
+        res
     }
 }
--- a/twoliter/src/docker/container.rs
+++ b/twoliter/src/docker/container.rs
@@ -1,0 +1,74 @@
+use crate::common::exec;
+use anyhow::Result;
+use log::{debug, log, Level};
+use std::path::Path;
+use tokio::process::Command;
+
+pub(crate) struct DockerContainer {
+    name: String,
+}
+
+impl DockerContainer {
+    /// Create a docker image with the given name from the image by using `docker create`.
+    pub(crate) async fn new<S1, S2>(container_name: S1, image: S2) -> Result<Self>
+    where
+        S1: Into<String>,
+        S2: Into<String>,
+    {
+        let name = container_name.into();
+        let image = image.into();
+
+        // Make sure previous versions of this container are stopped deleted.
+        cleanup_container(&name, Level::Trace).await;
+
+        debug!("Creating docker container '{name}' from image '{image}'");
+
+        // Create the new container.
+        let args = vec![
+            "create".to_string(),
+            "--rm".to_string(),
+            "--name".to_string(),
+            name.to_string(),
+            image.to_string(),
+        ];
+
+        exec(Command::new("docker").args(args), true).await?;
+        Ok(Self { name })
+    }
+
+    /// Copy the data from this container to a local destination.
+    pub(crate) async fn cp_out<P1, P2>(&self, src: P1, dest: P2) -> Result<()>
+    where
+        P1: AsRef<Path>,
+        P2: AsRef<Path>,
+    {
+        debug!(
+            "Copying '{}' from '{}' to '{}'",
+            src.as_ref().display(),
+            self.name,
+            dest.as_ref().display()
+        );
+        let mut args = vec!["cp".to_string()];
+        args.push(format!("{}:{}", self.name, src.as_ref().display()));
+        args.push(dest.as_ref().display().to_string());
+        exec(Command::new("docker").args(args), true).await
+    }
+}
+
+impl Drop for DockerContainer {
+    fn drop(&mut self) {
+        let name = self.name.clone();
+        tokio::task::spawn(async move { cleanup_container(&name, Level::Error).await });
+    }
+}
+
+async fn cleanup_container(name: &str, log_level: Level) {
+    let args = vec!["stop".to_string(), name.to_string()];
+    if let Err(e) = exec(Command::new("docker").args(args), true).await {
+        log!(log_level, "Unable to stop container '{}': {e}", name)
+    }
+    let args = vec!["rm".to_string(), name.to_string()];
+    if let Err(e) = exec(Command::new("docker").args(args), true).await {
+        log!(log_level, "Unable to remove container '{}': {e}", name)
+    }
+}
--- a/twoliter/embedded/Dockerfile
+++ b/twoliter/embedded/Dockerfile
@@ -144,7 +144,6 @@ USER builder
 RUN --mount=source=.cargo,target=/home/builder/.cargo \
     --mount=type=cache,target=/home/builder/.cache,from=cache,source=/cache \
     --mount=type=cache,target=/home/builder/rpmbuild/BUILD/sources/models/src/variant,from=variantcache,source=/variantcache \
-    --mount=type=cache,target=/home/builder/rpmbuild/BUILD/sources/logdog/conf/current,from=variantcache,source=/variantcache \
     --mount=source=sources,target=/home/builder/rpmbuild/BUILD/sources \
     rpmbuild -ba --clean \
       --undefine _auto_set_build_flags \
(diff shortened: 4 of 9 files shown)


twoliter/src/cmd/build.rs
```diff
@@ -1,6 +1,20 @@
...
-use std::path::PathBuf;
+use log::{debug, trace};
+use std::fs;
+use std::path::{Path, PathBuf};
+use tempfile::TempDir;
+use tokio::fs::{create_dir, remove_dir_all, remove_file};
+
+const ALPHA_INFRA: &[&str] = &[
+    "sbkeys/generate-local-sbkeys",
+    "sbkeys/generate-aws-sbkeys",
```
webern: We are putting these in the Alpha SDK image when we create it.

twoliter/src/cmd/build.rs
```diff
@@ -23,12 +37,102 @@ pub(crate) struct BuildVariant {
     project_path: Option<PathBuf>,
 
     /// The architecture to build for.
-    #[clap(long = "arch", default_value = "x86_64")]
+    #[clap(long = "arch", env = "BUILDSYS_ARCH", default_value = "x86_64")]
     arch: String,
+
+    /// The variant to build.
+    #[clap(env = "BUILDSYS_VARIANT")]
```
webern: In the long term these environment variables should become internal implementation details of the `Makefile.toml`. We shouldn't propagate them outward into the "new" interface for building Bottlerocket. Users can define whatever environment variables they want, but I think we should favor passing values like these as arguments.

```suggestion
    #[clap(long)]
```

twoliter/src/cmd/build.rs
```diff
@@ -23,12 +37,102 @@ pub(crate) struct BuildVariant {
...
+            continue;
+        }
+        if Path::new(file_name).is_dir() {
+            trace!("Alpha directory '{file_name}' already exists. Skipping");
+            continue;
+        }
+        sdk_container
+            .cp(
+                &Path::new("twoliter/alpha/").join(file_name),
+                Path::new(file_name),
```
webern: Are we putting these into the user's project directory? If so that's wonky. Can we place them in a tempdir each time we are invoked?

twoliter/src/cmd/build.rs
```diff
@@ -23,12 +37,102 @@ pub(crate) struct BuildVariant {
...
+        sdk_container
+            .cp(Path::new("twoliter/alpha/build/rpms"), packages_dir.path())
+            .await?;
+
+        let rpms_dir = project.project_dir().join("build").join("rpms");
+        fs::create_dir_all(&rpms_dir)?;
+        debug!("Copying rpms to build dir");
+        for maybe_file in fs::read_dir(packages_dir.path().join("rpms"))? {
+            let file = maybe_file?;
+            if !file.file_type()?.is_file() {
```
webern: I'm concerned that this check is not deep enough. If I update my SDK, I might be confused as to why I am not getting the newer versions of the RPMs. How expensive is it to overwrite all RPMs every time?

twoliter/src/cmd/build.rs
```diff
@@ -23,12 +37,102 @@ pub(crate) struct BuildVariant {
...
+            .env("GO_MODULES", &self.go_modules)
+            .env("BUILDSYS_ARCH", &self.arch)
+            .env("BUILDSYS_VARIANT", &self.variant)
+            .makefile(makefile_path)
+            .project_dir(project.project_dir())
+            .exec("build")
+            .await;
+
+        // Clean up all of the files we created
+        for file_name in created_files {
```
webern, line 116: This won't run if there's an error somewhere above this. We should try adding these files to a tempdir and then pointing to it. Although... maybe the scripts expect to be in the same dir as the keys that they might use :thinking:

twoliter/src/cmd/publish.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub(crate) async fn run(self) -> Result<()> {
+        match self {
+            PublishCommand::Ami(publish_ami) => publish_ami.run().await,
+        }
+    }
+}
+
+/// Publish a Bottlerocket variant image.
+#[derive(Debug, Parser)]
+pub(crate) struct PublishAmi {
```
webern: Do these args match the pubsys command that is being surfaced? The same concern applies in that I don't think we should reproduce the internal build system's environment variable interface.

twoliter/src/docker/container.rs
```diff
@@ -0,0 +1,59 @@
...
+        if let Err(e) = exec(Command::new("docker").args(args), true).await {
+            warn!("Unable to remove container '{name}': {e}")
+        }
+
+        debug!("Creating docker container '{name}' from image '{image}'");
+
+        // Create the new container.
+        let args = vec![
+            "create".to_string(),
+            "--rm".to_string(),
```
webern, line 29: When is the container removed? I'm used to seeing this with `docker run` where the container is removed when the process stops.

twoliter/src/cmd/build.rs
```diff
@@ -23,12 +37,102 @@ pub(crate) struct BuildVariant {
...
+    #[clap(long = "arch", env = "BUILDSYS_ARCH", default_value = "x86_64")]
     arch: String,
+
+    /// The variant to build.
+    #[clap(env = "BUILDSYS_VARIANT")]
+    variant: String,
+
+    /// The go modules that should be build
+    #[clap(long = "go-modules", env = "GO_MODULES", default_value = "")]
+    go_modules: String,
```
bcressey: I'd rather not expose this here. Potentially we want a contract like:
> Twoliter will search the `sources` path for `Cargo.lock` and `go.mod` files and download the referenced Rust crates and Go modules prior to package builds.

And eventually support for `.twoliterignore` files to suppress this behavior by excluding certain paths.

twoliter/src/cmd/build.rs
```diff
@@ -23,12 +37,102 @@ pub(crate) struct BuildVariant {
...
+            "sdk",
+            project
+                .sdk(&self.arch)
+                .context("The project was missing an sdk")?
+                .uri(),
+        )
+        .await?;
+        sdk_container
+            .cp(Path::new("twoliter/alpha/build/rpms"), packages_dir.path())
+            .await?;
```
bcressey: I would rather we modify the way we're building the alpha SDK to add repodata to the local path where we're copying the RPMs. Then adjust the `dnf` install lines in the `Dockerfile` to enable the "alpha" repo.

Ultimately kits will work something like this - RPMs copied out of containers along with repo definitions, to be used automatically when `dnf` runs.

twoliter/src/cmd/publish.rs
```diff
@@ -0,0 +1,57 @@
...
+    Ami(PublishAmi),
+}
+
+impl PublishCommand {
+    pub(crate) async fn run(self) -> Result<()> {
+        match self {
+            PublishCommand::Ami(publish_ami) => publish_ami.run().await,
+        }
+    }
+}
```
bcressey: IMO - we should have the "type" of the publication target be recorded in the variant manifest somehow, so that "AMI" doesn't leak into this interface.

```
# good
twoliter publish
# publishes an AMI for aws-*
# publishes an OVA for vmware-*
# prints a "skipping" or no-op message for unrecognized types

# bad
twoliter publish ami
twoliter publish ova
twoliter publish metal # what would this do?
```

This is starting to get into the question of how we merge `Infra.toml` with whatever shape we would use to specify the VMware equivalent, and pull all that back into `Twoliter.toml`.

I don't think we want to resolve that right now but `twoliter publish ami` feels like a mistake to avoid.

twoliter/src/cmd/build.rs
ecpullen: The script requires it to be in the sb-keys dir.

twoliter/src/cmd/build.rs
webern: > I'd rather not expose [Go Modules] here. Potentially we want a contract like:

This is tricky because it either has to be implemented now, or there has to be a way to provide the list of Go modules (because the build system currently operates using a list of Go modules, effectively hard-coded in the Makefile, but which can be overridden).

A possible middle ground would be to include it here with a high priority follow-up issue to remove it.

twoliter/src/cmd/publish.rs
webern: > twoliter publish

We probably want it to be `twoliter publish image` to distinguish it from `twoliter publish kit`.

twoliter/src/cmd/build.rs
webern: Oh look, I have a very well-written (:sweat_smile: ) issue for this already. #88

twoliter/src/cmd/build.rs
webern: Per our discussion we are going to defer this and work on it as part of the kits milestone.

twoliter/src/cmd/publish.rs
ecpullen: I've removed publishing from this pr.

twoliter/src/docker/container.rs
ecpullen: Added implementation for drop.

twoliter/src/cmd/build.rs
ecpullen: `twoliter build variant` now uses the existing sbkeys dir, or uses a tempdir if one doesn't exist.

twoliter/src/cmd/build.rs
ecpullen: We now copy all files over.

twoliter/src/cmd/build.rs
ecpullen: We are only adding a models directory to the users tree, and it is automatically cleaned up.

twoliter/src/cmd/build.rs
ecpullen: I've removed all env variables.

twoliter/embedded/Makefile.toml
```diff
@@ -97,7 +97,8 @@ BUILDSYS_JOBS = "8"
 CARGO_HOME = "${BUILDSYS_ROOT_DIR}/.cargo"
 # This needs to end with pkg/mod so that we can mount the parent of pkg/mod as GOPATH.
 GO_MOD_CACHE = "${BUILDSYS_ROOT_DIR}/.gomodcache/pkg/mod"
-GO_MODULES = "ecs-gpu-init host-ctr"
+# Dynamically load a list of go modules from ${BUILDSYS_SOURCE_DIR}
+GO_MODULES = { script = ['find ${BUILDSYS_SOURCES_DIR} -name go.mod -type f -printf "%h\n" | xargs -n1 basename'] }
```
webern, line 101: Do we want to do this with Rust prior to the calling of `cargo make` and pass the found go modules with `-e GO_MODULES`? I'm thinking now is as good a time as any to start moving logic away from Makefile.toml (or at least not adding more logic to it).

twoliter/src/cmd/build.rs
```diff
@@ -25,10 +33,94 @@ pub(crate) struct BuildVariant {
...
+
+        let rpms_dir = project.project_dir().join("build").join("rpms");
+        fs::create_dir_all(&rpms_dir)?;
+        debug!("Copying rpms to build dir");
+        for maybe_file in fs::read_dir(packages_dir.path().join("rpms"))? {
+            let file = maybe_file?;
+            debug!("Copying '{}'", file.path().display());
+            fs::copy(file.path(), rpms_dir.join(file.file_name()))?;
```
webern: Can this be a move instead (faster)? i.e. the equivalent of the `mv` command, but from tokio::fs?

twoliter/src/cmd/build.rs
```diff
@@ -25,10 +33,94 @@ pub(crate) struct BuildVariant {
...
+
+        // Clean up all of the files we created
+        for file_name in created_files {
+            let added = Path::new(&file_name);
+            if added.is_file() {
+                remove_file(added).await?;
+            } else if added.is_dir() {
+                remove_dir_all(added).await?;
+            }
+        }
```
webern, line 123: I'm a bit skeptical of this. The frequency with which it will never run (a previous error occurred, or the user cancelled the program), and that all it contains is the directory `sources/models` makes me think it's fine to just leave these cryptic directories in place for alpha.

webern (commented):
Looks good, mostly a thought here about not using bash in the Makefile for go module resolution. I also don't like the fact that we are forcing users to have a flat directory structure for their go modules, however that is currently a limitation of downstream logic in the Makefile.toml.

One thing we could do is make the GO_MODULES variable a list of directory paths and change the downstream Makefile.toml logic to be less strict about them being top level sub-directories of `sources`.

twoliter/src/cmd/build.rs
ecpullen: Does it hurt to try to clean it up?

webern: I guess not, other than delaying the surprise of finding them until an error or cancelled execution has occurred.

webern (approved):

twoliter/src/docker/container.rs
```diff
@@ -0,0 +1,75 @@
...
+            "--name".to_string(),
+            name.to_string(),
+            image.to_string(),
+        ];
+
+        exec(Command::new("docker").args(args), true).await?;
```
bcressey, line 35: This gets repeated often enough that a `docker_exec` function that just took `args` might be useful.

twoliter/src/docker/container.rs
```diff
@@ -0,0 +1,75 @@
...
+
+        // Make sure previous versions of this container are stopped deleted.
+        let args = vec!["stop".to_string(), name.to_string()];
+        if let Err(e) = exec(Command::new("docker").args(args), true).await {
+            trace!("Unable to stop container '{name}': {e}")
+        }
+        let args = vec!["rm".to_string(), name.to_string()];
+        if let Err(e) = exec(Command::new("docker").args(args), true).await {
+            trace!("Unable to remove container '{name}': {e}")
+        }
```
bcressey: Might be good to factor this into a helper function since we repeat this twice (here and in Drop).

twoliter/src/cmd/build.rs
```diff
@@ -25,10 +33,94 @@ pub(crate) struct BuildVariant {
...
+        install_tools(&tempdir).await?;
+        let makefile_path = tempdir.path().join("Makefile.toml");
+        let packages_dir =
+            TempDir::new().context("Unable to create a tempdir for Twoliter's packages")?;
+
+        let sdk_container = DockerContainer::new(
+            format!("sdk-{}", token),
+            project
+                .sdk(&self.arch)
+                .context("The project was missing an sdk")?
```
bcressey: nit: maybe add more context:
```suggestion
                .context(format!("No SDK defined in {} for {}", &self.project_path.display(), &self.arch))?
```

twoliter/src/cmd/build.rs
```diff
@@ -25,10 +33,94 @@ pub(crate) struct BuildVariant {
...
 impl BuildVariant {
     pub(super) async fn run(&self) -> Result<()> {
-        Ok(())
+        let project = project::load_or_find_project(self.project_path.clone()).await?;
+        let token = project.token();
+        let tempdir = tools_tempdir()?;
+        install_tools(&tempdir).await?;
+        let makefile_path = tempdir.path().join("Makefile.toml");
+        let packages_dir =
+            TempDir::new().context("Unable to create a tempdir for Twoliter's packages")?;
```
bcressey: It would be better to use `NamedTempFile` to create the temp directory here. `TempDir::new` is likely to give a path on `/tmp`, which may be on a different filesystem than the project - for example, if `tmp` is a tmpfs mount, or if the developer's home directory is on a separate volume. `fs::rename` below relies on the files being on the same filesystem.

twoliter/src/cmd/build.rs
```diff
@@ -25,10 +33,94 @@ pub(crate) struct BuildVariant {
...
+        let sdk_container = DockerContainer::new(
+            format!("sdk-{}", token),
+            project
+                .sdk(&self.arch)
+                .context("The project was missing an sdk")?
+                .uri(),
+        )
+        .await?;
+        sdk_container
+            .cp(Path::new("twoliter/alpha/build/rpms"), packages_dir.path())
```
bcressey: nit: make the function name more descriptive so the direction of the copy is clear:
```suggestion
            .copy_out(Path::new("twoliter/alpha/build/rpms"), packages_dir.path())
```

Also, the function can take an `AsRef<Path>` so we don't need to instantiate a `Path` here and can just pass a `&str`.

twoliter/src/cmd/build.rs
```diff
@@ -25,10 +33,94 @@ pub(crate) struct BuildVariant {
...
+            Box::new(in_tree_sbkeys_dir)
+        } else {
+            // Create a temp sbkeys directory
+            debug!("sbkeys dir not found. Creating a temporary directory");
+            let sbkeys_dir = TempDir::new().context("Unable to create a tempdir for sbkeys")?;
+            trace!("Using sbkeys tempdir: '{}'", sbkeys_dir.path().display());
+            sdk_container
+                .cp(Path::new("twoliter/alpha/sbkeys"), sbkeys_dir.path())
+                .await?;
+            Box::new(sbkeys_dir)
```
bcressey: This needs to be a persistent path, otherwise locally generated keys may end up lost. This directory should have the same lifetime as the `keys` and `roles` directories for TUF signing (which should also be persistent).

twoliter/src/cmd/build.rs
```diff
@@ -25,10 +33,94 @@ pub(crate) struct BuildVariant {
...
+
+        // TODO: Remove once models is no longer conditionally compiled.
+        // Create the models directory for the sdk to mount
+        let models_dir = project.project_dir().join("sources/models");
+        if !models_dir.is_dir() {
+            debug!("models source dir not found. Creating a temporary directory");
+            fs::create_dir_all(&models_dir.join("src/variant"))
+                .context("Unable to create models source directory")?;
+            created_files.push(models_dir)
+        }
```
bcressey, line 102: Can we `mkdir -p` this inside the embedded `Makefile.toml` so that the directory only gets created in the build context, not out on the host? That would eliminate the need for the created files cleanup later.

twoliter/src/cmd/build.rs
```diff
@@ -25,10 +33,94 @@ pub(crate) struct BuildVariant {
...
+
+        // Hold the result of the cargo make call so we can clean up the project directory first.
+        let res = CargoMake::new(&project, &self.arch)?
+            .env("TWOLITER_TOOLS_DIR", tempdir.path().display().to_string())
+            .env("BUILDSYS_ARCH", &self.arch)
+            .env("BUILDSYS_VARIANT", &self.variant)
+            // TODO Support dynamically loading go modules based on packages
+            .env("GO_MODULES", "")
```
bcressey: This is supported now, right?

bcressey (approved):

PR #112, opened 2023-11-14
twoliter: Remove Release.toml from variant build

### LINKED ISSUE #109: eliminate release.toml from build variant command
## Description

Release.toml should become a relic of the "old build system". Instead we intend to specify the version of the images being built either in Twoliter.toml as a global to be applied to all images, (and perhaps as an input to the build variant command to override what is found in Twoliter.toml?)

Since we are not handling migrations, and migrations are going away with the new settings system, we should not propagate the Release.toml concept into Twoliter's project layout.

For the time being, `twoliter build variant` can create a temporary, one-line Release.toml file to satisfy the build system.

This will not affect Bottlerocket until later because it is using `twoliter make` for now.

## Definition of Done

When the version of the images being built is in Twoliter.toml, Release.toml file is not needed, and Release.toml is removed from `tests/projects/project1`.

ecpullen:
**Issue number:**

Closes #109 

**Description of changes:**

Instead of using `Release.toml` to determine the version, use the version listed in the variants `Cargo.toml` instead.

**Testing done:**

Unit test added to `tests/variant_version`
Integ test with `project1`

```
$ cargo run --manifest-path /home/ec2-user/twoliter/Cargo.toml -p twoliter build variant  hello-ootb

[cargo-make] INFO - Running Task: publish-setup
19:48:02 [INFO] No infra config at '/home/ec2-user/twoliter/tests/projects/project1/Infra.toml' - using local roles/keys
[cargo-make] INFO - Running Task: build-variant
   Compiling hello-ootb v0.1.0 (/home/ec2-user/twoliter/tests/projects/project1/variants/hello-ootb)
    Finished dev [optimized] target(s) in 1m 07s
[cargo-make] INFO - Build Done in 77.66 seconds.
```

**Terms of contribution:**


webern:
Rebase me!

twoliter/src/variant.rs
```diff
@@ -0,0 +1,46 @@
+use crate::common::exec;
+use anyhow::{Context, Result};
+use serde_json::Value;
+use std::path::Path;
+use tokio::process::Command;
+
+pub(crate) async fn variant_version<S>(variants_dir: &Path, variant: S) -> Result<String>
```
webern: I think we should execute this cargo command inside the SDK container because eventually we want #114. This command is an easy one to run inside docker because it doesn't cause `build.rs` to be executed (which would create another container).

Ultimately this doesn't matter so much right now, but it would be one less thing to change later.

twoliter/src/variant.rs
```diff
@@ -0,0 +1,46 @@
...
+where
+    S: AsRef<str>,
+{
+    serde_json::from_str::<Value>(
+        &exec(
+            Command::new("cargo").args([
+                "metadata",
```
webern: @bcressey we discussed a few ideas about where we should put image (i.e. variant) versions. There could be a single one for all variants in Twoliter.toml, variants could be enumerated and versioned in Twoliter.toml, we could put it in the variant's custom Cargo.toml metadata...

Here we are using the actual cargo package version as the variant's version. What do you think?

Note, we are doing this now to remove the Release.toml requirement.

twoliter/embedded/Makefile.toml
```diff
@@ -137,6 +137,8 @@ TESTSYS_LOG_LEVEL = "info"
 # Certain variables are defined here to allow us to override a component value
 # on the command line.
 
+BUILDSYS_VERSION_IMAGE = { script = ["awk -F '[ =\"]+' '$1 == \"version\" {print $2}' ${BUILDSYS_RELEASE_CONFIG_PATH}"], condition = { env_not_set = ["BUILDSYS_VERSION_IMAGE"]}}
```
webern: I think we could require this to be set instead of discovering it from Release.toml here. This code is duplicated in Bottlerocket's Makefile.toml which means that Bottlerocket's build is always going to be setting this.

webern (commented):
Can't comment on `twoliter/src/cmd/new.rs` which appears to be extraneous in this PR.

local/README.md
```diff
@@ -1,3 +1,55 @@
+# Building The Alpha SDK
+
+## Required tools
+* git
+* cargo
+* docker
+
+## Clone Bottlerocket
```
bcressey: Should this be in a different commit? It doesn't seem related to removing Release.toml.

twoliter/src/common.rs
```diff
@@ -9,15 +9,16 @@ pub(crate) async fn exec_log(cmd: &mut Command) -> Result<()> {
...
+    exec(cmd, quiet).await?;
+    Ok(())
 }
 
 /// Run a `tokio::process::Command` and return a `Result` letting us know whether or not it worked.
 /// `quiet` determines whether or not the command output will be piped to `stdout/stderr`. When
-/// `quiet=true`, no output will be shown.
-pub(crate) async fn exec(cmd: &mut Command, quiet: bool) -> Result<()> {
+/// `quiet=true`, no output will be shown and will be returned instead.
+pub(crate) async fn exec(cmd: &mut Command, quiet: bool) -> Result<String> {
```
bcressey: Rather than sometimes returning an empty string, I'd prefer to return a `Result<Option<String>>`.

twoliter/src/variant.rs
bcressey: > Here we are using the actual cargo package version as the variant's version. What do you think?

My biggest gripe is that it's inconsistent with the cargo manifests for packages, where the crate version is ignored and only the version in the package's spec file has any external effect. As a minor nit, I also don't love parsing the crate manifest in different ways with different tools; we already have a parser in buildsys and could factor that out into a standalone crate if we needed to.

A single version in `Twoliter.toml` is the most consistent with existing logic and would make what I consider to be the common case easier - a project where multiple variants are built and released together, all tied to a specific Git commit tagged with a version. Updating a dozen versions across all variant manifests would be a lot of boilerplate for a routine chore, vs. just editing the top-level file.

I am still somewhat interested in the direction where variants (and kits) all have associated RPM spec files, to include package dependencies via Requires and stray files that don't make sense elsewhere. If we did that we could get the variant's version from the spec file, which would be consistent with package versioning.

But I'm not convinced that would be better. Before we added `application-inventory.json`, none of these internal versions had any external significance. The spec file versions were mostly there as an implementation detail and a nod to spec file conventions. They're also a bit annoying to maintain since every package update requires at least two files to edit.

My preference would be to start with a top-level `release-version` field in `Twoliter.toml`, and if we need variant-specific versions for some reason, we could add a `release-version-override` field to `package.metadata.build-variant`.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/7b0123f22840fd2275ac7a58965676cee73f6124..65b153ae78c3b76d7bd08bc2da7d2ff8bc63287e

Rebase.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/249571779ab10b569546d86194c81bd58dbffd7d..ff6125c786f12f25f3ecf0362a5332d01972eec3

- Broke the README change out to its own commit.
- Added `release-version` to `Twoliter.toml`
- Created a validating system when loading `Twoliter.toml` than can issue a warning about `Release.toml` and make sure it has the same version as `Twoliter.toml`.
- Updated Ethan's commit to take the version from `Twoliter.toml`

TODO: Probably should add tests of the `Release.toml` validation.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/ff6125c786f12f25f3ecf0362a5332d01972eec3..fd9907d6abeb3e8c1c962f92c5290dcb54e63e06

Improve a comment.

local/README.md
```diff
@@ -1,3 +1,55 @@
+# Building The Alpha SDK
```
ecpullen: This file can be removed. It wasn't supposed to be a part of this branch.

twoliter/embedded/Makefile.toml
webern: > I think we could require this to be set instead of discovering it from Release.toml here. This code is duplicated in Bottlerocket's Makefile.toml which means that Bottlerocket's build is always going to be setting this.

Note, we discussed, in `twoliter make` we can look for `BUILDSYS_VERSION_IMAGE`, if it's not there or empty, we can get it from Twoliter.toml and set it before calling `cargo make`.

twoliter/src/project.rs
```diff
@@ -157,64 +155,85 @@ impl ImageName {
...
+        let path = project_dir.join("Release.toml");
+        if !path.exists() || !path.is_file() {
+            // There is no Release.toml file. This is a good thing!
+            trace!("This project does not have a Release.toml file (this is not a problem)");
+            return Ok(());
+        }
+        warn!(
+            "A Release.toml file was found. Release.toml is deprecated. Please remove it from \
+             your project."
+        );
```
bcressey, line 207: nit: this could encourage people building from the main repo to send a PR to remove `Release.toml` - might want to drop it.

bcressey (approved):

twoliter/embedded/Makefile.toml
```diff
@@ -15,9 +15,13 @@ BUILDSYS_IMAGES_DIR = "${BUILDSYS_BUILD_DIR}/images"
 BUILDSYS_TOOLS_DIR = "${BUILDSYS_ROOT_DIR}/tools"
 BUILDSYS_SOURCES_DIR = "${BUILDSYS_ROOT_DIR}/sources"
 BUILDSYS_SBKEYS_DIR = "${BUILDSYS_ROOT_DIR}/sbkeys"
-BUILDSYS_SBKEYS_PROFILE = { script = ['echo "${BUILDSYS_SBKEYS_PROFILE:-local}"'] }
+BUILDSYS_SBKEYS_PROFILE = { script = [
```
webern: Please undo your IDE's auto-reformatting of this file.

webern (changes requested):

webern (approved):

twoliter/src/variant.rs
webern: We will work on this later.

Issue #56, opened 2023-09-18
tracking issue: kits milestone


webern:
Use this as a tracking issue for all issues that need to be completed for kit support.

- [x] #16 
- [x] #72 
- [x] #70 
- [x] #17 
- [x] #18 
- [x] #229

webern:
done!

PR #115, opened 2023-11-22
design: refine kits

webern:
**Issue number:**

Related to #56 

**Description of changes:**

Refine some concepts in the design of kits.

**Testing done:**

N/A

**Terms of contribution:**


ecpullen (approved):

docs/design/README.md
```diff
@@ -171,12 +170,11 @@ A variant such as `metal-k8s-1.22` would consume the necessary kits.
 For example, it would reference `core-kit`, `k8s-kit` and `metal-kit`.
 To add `aws-iam-authenticator`, `aws-kit` would be added.
 
-Certain build steps will benefit from the aggregation of all packages from a kit into a single RPM, such as `bottlerocket-aarch64-core-kit-0.1-1.aarch64.rpm`.
-These spec files could be added to the above directory tree, or else they could be auto-generated by the Kit build process.
-
 ### Kit Structure
 
 A kit container will be built using `FROM scratch` because it does not need a runtime environment.
+Kits need not be published with a multiarch manifest (because they are `FROM scratch`), but they do need to be published for both target architectures.
```
webern, line 176: Better explain this. `FROM scratch` is probably not arch specific. What I'm saying is that we probably don't need different images for different host architectures because the underlying image would be the same (just RPM files).

docs/design/README.md
```diff
@@ -256,6 +255,36 @@ etc.
 These metadata files will give twoliter the information it needs to resolve, pull, and aggregate all dependency kits and the correct SDK needed for a build.
 This dependency resolution will be modeled in Cargo and kept in a `Cargo.lock` file (see [Twoliter Update](#twoliter-update)).
 
+### Kit Aggregation when Building Variants
```
webern, line 258: Change this. It will require multiple Dockerfiles because a kit can depend on an external kit during it's build, and a local kit can depend on another local kit.

Instead, copy RPMs out to the build directory in a kit-namespaced yum repo structure.

docs/design/README.md
```diff
@@ -256,6 +255,36 @@ etc.
...
+FROM scratch
+
+# External Kits pulled from registries
+# TODO - is the registry name part of the path or kit name?
+COPY --from=kit1-x86_64:v0.1.0 /kit-data /kits/x86_64/kit1-v0.1.0
+COPY --from=kit2-x86_64:v0.2.0 /kit-data /kits/x86_64/kit2-v0.2.0
+
+# Local kits built on the host
+COPY ./build/kits/x86_64/mylocalkit1-v0.1.0 /kits/x86_64/mylocalkit1-v0.1.0
+COPY ./build/kits/x86_64/mylocalkit2-v0.2.0 /kits/x86_64/mylocalkit1-v0.2.0
```
bcressey, line 277: Summarizing discussion - the monolithic kit-of-kits approach will be challenging with local kits that depend on other local kits, where we need to instantiate the image multiple times.

docs/design/README.md
```diff
@@ -387,6 +416,69 @@ Directory descriptions:
...
+1) A Twoliter project may be used to create variant images.
+2) A Twoliter project may be used to create kits.
+
+Each of the above purposes may be achieved with or without depending on external kits[^nokits].
+
+[^nokits]: The use case of creating a variant without depending on external kits may seem remote.
+    But at least initially, this is what the Bottlerocket monorepo will be doing.
+
+When a user has no intention of creating a kit, it should not be necessary to define a kit in the kits directory.
+When a user has no intention of creating a variant image, it should not be necessary to define a variant in the variants directory.
```
bcressey, line 431: Do we want to support this "implicit kit" mode at the outset?

docs/design/README.md
```diff
@@ -387,6 +416,69 @@ Directory descriptions:
...
+
+To further complicate matters, some users may have projects with which have started with the [alpha](#twoliter-alpha) version of Twoliter.
+For the time being, if we assume backward compatibility with these projects, then they introduce another use case.
```
bcressey, line 434: I'm open to breaking compatibility here, possibly with a fix-up script or tool that can help with the migration.

docs/design/README.md
bcressey: To extend the thought - it feels like syntactic sugar or boilerplate reduction, where we could just force some local kit to be defined and use it.

The need for it may be more acute from a lack of tooling support for operations that a developer wants to do, like:
```
# Modifies metadata such that the named package is built when building the target kit.
twoliter add <package> <kit>

# Adds a dependency from the variant to the kit that contains the package.
# This becomes more complex if we also support implicit kits.
twoliter add <package> <variant>

# Or a lower-level version that requires the developer to know where the package comes from:
twoliter add <kit> <variant>
```

PR #127, opened 2023-12-08
add go and proxy vars to twoliter make

webern:
Certain environments will need variables like GOPROXY and HTTP_PROXY passed to the SDK container in docker run commands. The list of variables added in this commit were found in the docker-go script.

**Terms of contribution:**


etungsten:
For testing, can you try building Bottlerocket on a host without golang installed? I suspect the `docker-go` script will still try running `go env` and set `GOPROXY` to an empty string which causes problems. To completely fix it we need to set the default in `Makefile.toml` somewhere.

bcressey (approved):

jpculp (approved):

PR #129, opened 2023-12-26
provide crate fs functions and fix a remove_dir_all bug

webern:
**Description of changes:**

```
commit

    fix bug in make.rs

    If a directory does not exist, remove_dir_all will return an error but
    we do not care if the directory does not exist. Ignoring all errors in
    this case is a simple way to ignore the very common case that directory
    does not exist.

commit

    use crate::common::fs instead of tokio::fs

    Convert tokio::fs call sites to use crate::common::fs for better error
    messages. Also fix a case of blocking io.

commit

    add fs functions for better error messages

    Create simple wrapper functions for tokio::fs that provide better error
    messages.
```

**Terms of contribution:**

### DIFF
--- a/twoliter/src/cmd/debug.rs
+++ b/twoliter/src/cmd/debug.rs
@@ -1,9 +1,9 @@
+use crate::common::fs;
 use crate::tools::install_tools;
-use anyhow::{Context, Result};
+use anyhow::Result;
 use clap::Parser;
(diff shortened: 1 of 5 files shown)


yeazelm (approved):
LGTM!

vyaghras (approved):

Issue #70, opened 2023-09-19
kits: buildsys build-kit command


webern:
## Description

In order to build a kit image, buildsys will need to be invoked from a kit's `build.rs` file with the command `buildsys build-kit`. This is a new subcommand for buildsys

## Definition of Done

For the sake of making smaller, incremental PRs, it is Ok if this is not fully functional at first. It will be made fully functional in https://github.com/bottlerocket-os/twoliter/issues/17

This issue is done when some semblance of the command exists in Buildsys.

webern:
Closing as duplicate of #17

PR #134, opened 2024-01-17
buildsys: use clap for env vars

webern:
**Issue number:**

Related to, refactor before #70 

**Description of changes:**

Required input in the form of environment variables was sprinkled throughout the code. Here we aggregate the inputs into a CLI interface using Clap, while still allowing all to specified via environment variables.

**Testing done:**

- [x] Built an aws-k8s-1.27 image
- [x] Built it again and observed that packages were not rebuilt
- [x] touched a file (binutils.spec) and built it again - observed that binutils was rebuilt but other packages were not
- [x] Built an aws-ecs-2 image, many (but not all) packages were rebuilt due to changed variant
- [x] Built it again and observed that packages were not rebuilt
- [x] touched a file (binutils.spec) and built it again - observed that binutils was rebuilt but other packages were not

Also, during the refactoring process, I captured the output of the docker build command before and after this PR's changes. I found that I was getting the same build args and secrets args before and after (with the exception that SDK, TOOLCHAIN, TOKEN and NOCACHE have been added during build-variant, which seems fine).

**Terms of contribution:**

### DIFF
--- a/tools/buildsys/src/args.rs
+++ b/tools/buildsys/src/args.rs
@@ -5,6 +5,7 @@ of its input arguments from environment variables.
 
 !*/
 
+use buildsys::manifest::SupportedArch;
 use clap::{Parser, Subcommand};
 use std::path::PathBuf;
 use url::Url;
@@ -55,10 +56,10 @@ impl Command {
 #[derive(Debug, Parser)]
 pub(crate) struct Common {
     #[arg(long, env = "BUILDSYS_ARCH")]
-    pub(crate) arch: String,
+    pub(crate) arch: SupportedArch,
 
     #[arg(long, env = "BUILDSYS_OUTPUT_DIR")]
-    pub(crate) output_dir: PathBuf,
+    pub(crate) image_arch_variant_dir: PathBuf,
 
     #[arg(long, env = "BUILDSYS_ROOT_DIR")]
     pub(crate) root_dir: PathBuf,
@@ -80,6 +81,9 @@ pub(crate) struct Common {
 
     #[arg(long, env = "TLPRIVATE_TOOLCHAIN")]
     pub(crate) toolchain: String,
+
+    #[arg(long, env = "TWOLITER_TOOLS_DIR")]
+    pub(crate) tools_dir: PathBuf,
 }
 
 /// Build RPMs from a spec file and sources.
@@ -90,6 +94,21 @@ pub(crate) struct BuildPackageArgs {
 
     #[arg(long, env = "BUILDSYS_VARIANT")]
     pub(crate) variant: String,
+
+    #[arg(long, env = "BUILDSYS_VARIANT_PLATFORM")]
+    pub(crate) variant_platform: String,
+
+    #[arg(long, env = "BUILDSYS_VARIANT_RUNTIME")]
+    pub(crate) variant_runtime: String,
+
+    #[arg(long, env = "BUILDSYS_VARIANT_FAMILY")]
+    pub(crate) variant_family: String,
+
+    #[arg(long, env = "BUILDSYS_VARIANT_FLAVOR")]
+    pub(crate) variant_flavor: String,
+
+    #[arg(long, env = "PUBLISH_REPO")]
+    pub(crate) publish_repo: String,
 
     #[arg(long, env = "BUILDSYS_SOURCES_DIR")]
     pub(crate) sources_dir: PathBuf,
--- a/tools/buildsys/src/args.rs
+++ b/tools/buildsys/src/args.rs
@@ -1,0 +1,195 @@
+/*!
+
+These structs provide the CLI interface for buildsys which is called from Cargo.toml and accepts all
+of its input arguments from environment variables.
+
+!*/
+
+use clap::{Parser, Subcommand};
+use std::path::PathBuf;
+use url::Url;
+
+/// A list of environment variables and the type of build that should be rerun if that environment
+/// variable changes. The build type is represented with bit flags so that we can easily list
+/// multiple build types for a single variable. See `[BuildType]` and `[rerun_for_envs]` below to
+/// see how this list is used.
+const REBUILD_VARS: [(&str, u8); 13] = [
+    ("BUILDSYS_ARCH", PACKAGE | VARIANT),
+    ("BUILDSYS_NAME", VARIANT),
+    ("BUILDSYS_OUTPUT_DIR", VARIANT),
+    ("BUILDSYS_PACKAGES_DIR", PACKAGE),
+    ("BUILDSYS_PRETTY_NAME", VARIANT),
+    ("BUILDSYS_ROOT_DIR", PACKAGE | VARIANT),
+    ("BUILDSYS_STATE_DIR", PACKAGE | VARIANT),
+    ("BUILDSYS_TIMESTAMP", VARIANT),
+    ("BUILDSYS_VARIANT", VARIANT),
+    ("BUILDSYS_VERSION_BUILD", VARIANT),
+    ("BUILDSYS_VERSION_IMAGE", VARIANT),
+    ("TLPRIVATE_SDK_IMAGE", PACKAGE | VARIANT),
+    ("TLPRIVATE_TOOLCHAIN", PACKAGE | VARIANT),
+];
+
+/// A tool for building Bottlerocket images and artifacts.
+#[derive(Debug, Parser)]
+pub(crate) struct Buildsys {
+    #[command(subcommand)]
+    pub(crate) command: Command,
+}
+
+#[derive(Subcommand, Debug)]
+pub(crate) enum Command {
+    BuildPackage(Box<BuildPackageArgs>),
+    BuildVariant(Box<BuildVariantArgs>),
+}
+
+impl Command {
+    pub(crate) fn build_type(&self) -> BuildType {
+        match self {
+            Command::BuildPackage(_) => BuildType::Package,
+            Command::BuildVariant(_) => BuildType::Variant,
+        }
+    }
+}
+
+/// Arguments common to all subcommands.
+#[derive(Debug, Parser)]
+pub(crate) struct Common {
+    #[arg(long, env = "BUILDSYS_ARCH")]
+    pub(crate) arch: String,
+
+    #[arg(long, env = "BUILDSYS_OUTPUT_DIR")]
+    pub(crate) output_dir: PathBuf,
+
+    #[arg(long, env = "BUILDSYS_ROOT_DIR")]
+    pub(crate) root_dir: PathBuf,
+
+    #[arg(long, env = "BUILDSYS_STATE_DIR")]
+    pub(crate) state_dir: PathBuf,
+
+    #[arg(long, env = "BUILDSYS_TIMESTAMP")]
+    pub(crate) timestamp: String,
+
+    #[arg(long, env = "BUILDSYS_VERSION_FULL")]
+    pub(crate) version_full: String,
+
+    #[arg(long, env = "CARGO_MANIFEST_DIR")]
+    pub(crate) cargo_manifest_dir: PathBuf,
+
+    #[arg(long, env = "TLPRIVATE_SDK_IMAGE")]
+    pub(crate) sdk_image: String,
+
+    #[arg(long, env = "TLPRIVATE_TOOLCHAIN")]
+    pub(crate) toolchain: String,
+}
+
+/// Build RPMs from a spec file and sources.
+#[derive(Debug, Parser)]
+pub(crate) struct BuildPackageArgs {
+    #[arg(long, env = "BUILDSYS_PACKAGES_DIR")]
+    pub(crate) packages_dir: PathBuf,
+
+    #[arg(long, env = "BUILDSYS_VARIANT")]
+    pub(crate) variant: String,
+
+    #[arg(long, env = "BUILDSYS_SOURCES_DIR")]
+    pub(crate) sources_dir: PathBuf,
+
+    #[arg(long, env = "BUILDSYS_LOOKASIDE_CACHE")]
+    pub(crate) lookaside_cache: Url,
+
+    #[arg(long, env = "BUILDSYS_UPSTREAM_SOURCE_FALLBACK")]
+    pub(crate) upstream_source_fallback: String,
+
+    #[arg(long, env = "CARGO_PKG_NAME")]
+    pub(crate) cargo_package_name: String,
+
+    #[command(flatten)]
+    pub(crate) common: Common,
+}
+
+/// Build filesystem and disk images from RPMs.
+#[derive(Debug, Parser)]
+pub(crate) struct BuildVariantArgs {
+    #[arg(long, env = "BUILDSYS_NAME")]
+    pub(crate) name: String,
+
+    #[arg(long, env = "BUILDSYS_PRETTY_NAME")]
+    pub(crate) pretty_name: String,
+
+    #[arg(long, env = "BUILDSYS_VARIANT")]
+    pub(crate) variant: String,
+
+    #[arg(long, env = "BUILDSYS_VERSION_BUILD")]
+    pub(crate) version_build: String,
+
+    #[arg(long, env = "BUILDSYS_VERSION_IMAGE")]
+    pub(crate) version_image: String,
+
+    #[command(flatten)]
+    pub(crate) common: Common,
+}
+
+/// Returns the environment variables that need to be watched for a given `[BuildType]`.
+fn sensitive_env_vars(build_type: BuildType) -> impl Iterator<Item = &'static str> {
+    REBUILD_VARS
+        .into_iter()
+        .filter(move |(_, flags)| build_type.includes(*flags))
+        .map(|(var, _)| var)
+}
+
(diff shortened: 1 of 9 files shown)


tools/buildsys/src/builder.rs
```diff
@@ -222,17 +231,24 @@ impl VariantBuilder {
...
+            BuildType::Variant,
+            &build_variant_args.variant,
+            arch,
+            args,
+            &tag,
+            &build_variant_args.common.output_dir,
+            &build_variant_args.common.root_dir,
+            &build_variant_args.common.state_dir,
+            &build_variant_args.common.sdk_image,
+            &build_variant_args.common.toolchain,
```
bcressey: I see there's a TODO for this, but it would be nice to have a `BuildContext` type to hold these values so that the `build` function signature isn't so complicated.

tools/buildsys/src/builder.rs
```diff
@@ -147,18 +158,16 @@ pub(crate) struct VariantBuilder;
...
     ) -> Result<Self> {
-        let output_dir: PathBuf = getenv("BUILDSYS_OUTPUT_DIR")?.into();
-
-        let variant = getenv("BUILDSYS_VARIANT")?;
-        let arch = getenv("BUILDSYS_ARCH")?;
-        let goarch = serde_plain::from_str::<SupportedArch>(&arch)
-            .context(error::UnsupportedArchSnafu { arch: &arch })?
+        let arch = &build_variant_args.common.arch;
+        let goarch = serde_plain::from_str::<SupportedArch>(arch)
+            .context(error::UnsupportedArchSnafu { arch })?
```
bcressey: Can we do this conversion from `&str` to `SupportedArch` just once, perhaps when parsing arguments? Otherwise we end up doing it in a few different places - here and every time `supported_arch` is called.

tools/buildsys/src/cache.rs
```diff
@@ -79,11 +78,9 @@ impl LookasideCache {
 
     /// Retrieves a file from the specified URL and write it to the given path,
     /// then verifies the contents against the SHA-512 hash provided.
-    fn fetch_file<P: AsRef<Path>>(url: &str, path: P, hash: &str) -> Result<()> {
+    fn fetch_file<P: AsRef<Path>>(url: &str, path: P, hash: &str, version: &str) -> Result<()> {
```
bcressey: I don't love this function signature change - it feels like one thing to add additional context to a request header based on an environment variable, and another to take an otherwise useless parameter just to do the same.

I'd prefer to either not emit the version at all - do we really care? - or else to add it as a field to `LookasideCache`, pass it into `LookasideCache::new`, and retrieve it from `self`.

tools/buildsys/src/args.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub(crate) toolchain: String,
+}
+
+impl Common {
+    pub(crate) fn rerun_for_envs() {
+        [
+            "BUILDSYS_ARCH",
+            "BUILDSYS_OUTPUT_DIR",
+            "BUILDSYS_ROOT_DIR",
+            "BUILDSYS_VERSION_FULL",
```
bcressey: This would trigger rebuilds of everything whenever the project version changes - new commit or dirty tree.

(The different `getenv` functions sometimes had the "rerun-if-changed" side effect and sometimes did not - it should just be the lookups in `builder.rs` that had the side effect.)

tools/buildsys/src/args.rs
```diff
@@ -0,0 +1,143 @@
...
+    #[command(flatten)]
+    pub(crate) common: Common,
+}
+
+impl BuildPackageArgs {
+    pub(crate) fn rerun_for_envs() {
+        Common::rerun_for_envs();
+        [
+            "BUILDSYS_PACKAGES_DIR",
+            "BUILDSYS_SOURCES_DIR",
```
bcressey: I'm OK with unconditionally emitting these for variables we expect to be constant for the life of a given workspace, if we want to simplify slightly and confine the package- and variant-specific `rerun_for_envs()` to truly special cases.

tools/buildsys/src/builder.rs
```diff
@@ -590,14 +607,6 @@ where
         .filter(|e| e.is_file() || e.is_symlink())
 }
 
-/// Retrieve a BUILDSYS_* variable that we expect to be set in the environment,
-/// and ensure that we track it for changes, since it will directly affect the
-/// output.
-fn getenv(var: &str) -> Result<String> {
-    println!("cargo:rerun-if-env-changed={}", var);
```
bcressey, line 597: It's really just this one `getenv` function that emits `rerun-if-changed`, which is why you picked up some new variables for tracking.

tools/buildsys/src/args.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+impl Common {
+    pub(crate) fn rerun_for_envs() {
+        [
+            "BUILDSYS_ARCH",
+            "BUILDSYS_OUTPUT_DIR",
+            "BUILDSYS_ROOT_DIR",
+            "BUILDSYS_VERSION_FULL",
+            "CARGO_MANIFEST_DIR",
```
bcressey: I don't think we were emitting rerun-if-changed for this previously under any condition.
```suggestion
```

tools/buildsys/src/args.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub(crate) common: Common,
+}
+
+impl BuildPackageArgs {
+    pub(crate) fn rerun_for_envs() {
+        Common::rerun_for_envs();
+        [
+            "BUILDSYS_PACKAGES_DIR",
+            "BUILDSYS_SOURCES_DIR",
+            "BUILDSYS_VARIANT",
```
bcressey: This will trigger a rebuild of all packages whenever the variant changes. We need to continue using the selective logic in main.rs instead.
```suggestion
```

tools/buildsys/src/args.rs
```diff
@@ -0,0 +1,143 @@
...
+            "CARGO_PKG_NAME",
```
bcressey: We don't track this for changes today.
```suggestion
```

tools/buildsys/src/args.rs
```diff
@@ -0,0 +1,143 @@
...
+
+impl BuildVariantArgs {
+    pub(crate) fn rerun_for_envs() {
+        Common::rerun_for_envs();
+        [
+            "BUILDSYS_NAME",
+            "BUILDSYS_PRETTY_NAME",
+            "BUILDSYS_VARIANT",
+            "BUILDSYS_VERSION_BUILD",
+            "BUILDSYS_VERSION_IMAGE",
```
bcressey: This list should have the other variables that only drive variant rebuilds.
```suggestion
            "BUILDSYS_VERSION_BUILD",
            "BUILDSYS_VERSION_IMAGE",
            "BUILDSYS_VERSION_FULL",
            "BUILDSYS_TIMESTAMP",
```

tools/buildsys/src/args.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub(crate) state_dir: PathBuf,
+
+    #[arg(long, env = "BUILDSYS_TIMESTAMP")]
+    pub(crate) timestamp: String,
+
+    #[arg(long, env = "BUILDSYS_VERSION_FULL")]
+    pub(crate) version_full: String,
+
+    #[arg(long, env = "CARGO_MANIFEST_DIR")]
+    pub(crate) cargo_manifest_dir: PathBuf,
```
bcressey, line 77: I'm not as excited about converting cargo's own environment variables into arguments since the environment is the canonical way to pass this information.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/b535980ed80a54ce9aaa803de228e4120a3ad096..a1eae7b2165900eca3e07e754c7c0b5a09469b89

Fixup the cargo:rerun-if-env-changed list to make it the same as it used to be.

tools/buildsys/src/args.rs
webern: My preference would be to include it here anyway for two reasons:
1) Readability: self-documenting code
2) If I want to write buildsys tests, I may call it directly instead of invoking it from a build.rs script and using args might be helpful for that

tools/buildsys/src/cache.rs
webern: I don't know if we care about emitting the version, so I've kept it but with your suggestion to make it self.version.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/a1eae7b2165900eca3e07e754c7c0b5a09469b89..c71f56ece7a6e1df398e5552a96c756b34a1e51f

Make the version a parameter of LookasideCache

webern:
I've gotten myself into a much bigger refactor of builder.rs so converted this to a draft.

webern:
@bcressey this is ready for you to re-review. The refactor is rather substantial, but I think it is an improvement. My goal was to separate the concerns of a) gathering data from the environment, b) manufacturing and constructing args for docker, and c) the program flow of calling docker. Previously all of those things were happening in the same code-paths. I think having them in separate code paths will make it easier to hack on this for kits.

tools/buildsys/src/builder.rs
```diff
@@ -86,269 +85,355 @@ lazy_static! {
...
-) -> Result<()> {
-    let root = getenv("BUILDSYS_ROOT_DIR")?;
-    env::set_current_dir(&root).context(error::DirectoryChangeSnafu { path: &root })?;
+pub(crate) struct DockerBuild {
+    dockerfile: PathBuf,
+    context: PathBuf,
+    target: String,
+    tag: String,
+    root_dir: PathBuf,
+    // build_dir: PathBuf,
```
webern: I fixed this.
```suggestion
```

webern:
> https://github.com/bottlerocket-os/twoliter/compare/9a89bf58869592d3f451af59a5cb35b5c81a9329..be37869a03eda322ed07e527aacd978a5b0a8bb9

Slightly painful rebase.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/be37869a03eda322ed07e527aacd978a5b0a8bb9..8b857d98fb2aed710b662049273591b2e2888596

Clippy fix.

tools/buildsys/src/builder.rs
```diff
@@ -86,269 +85,355 @@ lazy_static! {
...
+                args.common.sdk_image,
+                args.common.toolchain,
+                args.common.arch,
+            ),
+            target_build_args: TargetBuildArgs::Package(PackageBuildArgs {
+                image_features: manifest
+                    .image_features()
+                    .map(|set| set.iter().map(|&x| x.to_owned()).collect()),
+                package,
+                publish_repo: args.publish_repo,
```
bcressey, line 255: I opened #151 to correct this since it felt out of place here.

tools/buildsys/src/builder.rs
```diff
@@ -86,269 +85,355 @@ lazy_static! {
...
-
-    // Clean up our image now that we're done.
-    docker(&rmi, Retry::No)?;
-
-    // Copy artifacts to the expected directory and write markers to track them.
-    copy_build_files(&build_dir, output_dir)?;
+            context = self.context.display(),
+            dockerfile = self.dockerfile.display(),
+            target = self.target,
+            tag = self.tag,
```
bcressey: nit: the spacing on these lines seems off?

tools/buildsys/src/args.rs
```diff
@@ -0,0 +1,214 @@
...
+use clap::{Parser, Subcommand};
+use std::path::PathBuf;
+use url::Url;
+
+/// A list of environment variables and the type of build that should be rerun if that environment
+/// variable changes. The build type is represented with bit flags so that we can easily list
+/// multiple build types for a single variable. See `[BuildType]` and `[rerun_for_envs]` below to
+/// see how this list is used.
+const REBUILD_VARS: [(&str, u8); 13] = [
+    ("BUILDSYS_ARCH", PACKAGE | VARIANT),
```
bcressey, line 18: ❤️ 

This is a nice way to organize something that was very ad-hoc and organic before.

tools/buildsys/src/args.rs
```diff
@@ -0,0 +1,214 @@
...
+    ("BUILDSYS_NAME", VARIANT),
+    ("BUILDSYS_OUTPUT_DIR", VARIANT),
+    ("BUILDSYS_PACKAGES_DIR", PACKAGE),
```
bcressey, line 21: nit: it feels like a (pre-existing) mistake to not track `BUILDSYS_SOURCES_DIR` if we are tracking `BUILDSYS_PACKAGES_DIR` - if we override the build to point to a completely different sources tree then that should invalidate any builds that used those sources.

tools/buildsys/src/args.rs
```diff
@@ -0,0 +1,214 @@
...
+}
+
+/// Build RPMs from a spec file and sources.
+#[derive(Debug, Parser)]
+pub(crate) struct BuildPackageArgs {
+    #[arg(long, env = "BUILDSYS_PACKAGES_DIR")]
+    pub(crate) packages_dir: PathBuf,
+
+    #[arg(long, env = "BUILDSYS_VARIANT")]
+    pub(crate) variant: String,
```
bcressey, line 96: To check my understanding - we don't want this in `Common` because kits won't need this variable?

bcressey (approved):

tools/buildsys/src/args.rs
```diff
@@ -0,0 +1,214 @@
...
+    pub(crate) tools_dir: PathBuf,
+}
+
+/// Build RPMs from a spec file and sources.
+#[derive(Debug, Parser)]
+pub(crate) struct BuildPackageArgs {
+    #[arg(long, env = "BUILDSYS_PACKAGES_DIR")]
+    pub(crate) packages_dir: PathBuf,
+
+    #[arg(long, env = "BUILDSYS_VARIANT")]
```
ecpullen, line 95: This goes away once we no longer have conditional compilation right?

ecpullen (approved):
LGTM

tools/buildsys/src/builder.rs
```diff
@@ -86,269 +85,355 @@ lazy_static! {
...
-    let twoliter_tools_dir =
-        env::var(TOOLS_DIR).context(error::EnvironmentSnafu { var: TOOLS_DIR })?;
-    let dockerfile = PathBuf::from(twoliter_tools_dir).join("Dockerfile");
+        // Clean up any previous outputs we have tracked.
+        clean_build_files(&marker_dir, &self.artifacts_dir)?;
 
-    let mut build = format!(
-        "build . \
+        let mut build = format!(
+            "build {context} \
```
yeazelm, line 360: nit: this intending looks strange to me?

tools/buildsys/src/builder.rs
```diff
@@ -358,6 +443,8 @@ fn docker(args: &[String], retry: Retry) -> Result<Output> {
         retry_messages = messages;
     }
 
+    println!("Running command:\ndocker {}", args.join(" /\n"));
```
yeazelm: This is only one place out of two we `println!()` in this file. Is that intentional? I think so but wanted to ensure this isn't a hold over from debugging and this output is doing something useful for logs/output for the user?

yeazelm (approved):

tools/buildsys/src/builder.rs
webern: > This is only one place out of two we println!() in this file. Is that intentional? I think so but wanted to ensure this isn't a hold over from debugging and this output is doing something useful for logs/output for the user?

Good catch. Fixed.

tools/buildsys/src/builder.rs
webern: > nit: the spacing on these lines seems off?

Yeah fmt is no help here. I've now aligned the multiline string line starts with the variable list below it.

tools/buildsys/src/builder.rs
webern: > nit: this intending looks strange to me?

Yeah fmt is no help here. I've now aligned the multiline string line starts with the variable list below it.

tools/buildsys/src/args.rs
webern: > This goes away once we no longer have conditional compilation right?

That should be the goal for packages, yes.

tools/buildsys/src/args.rs
webern: > To check my understanding - we don't want this in `Common` because kits won't need this variable?

That's the idea, yeah. And I think our goal should be that package builds also don't need to know the variant name.

tools/buildsys/src/args.rs
webern: > nit: it feels like a (pre-existing) mistake to not track `BUILDSYS_SOURCES_DIR` if we are tracking `BUILDSYS_PACKAGES_DIR` - if we override the build to point to a completely different sources tree then that should invalidate any builds that used those sources.

I'm going to defer this because I'm concerned that the symlink change when changing variant could trigger something. I'll track this as a GitHub issue and play around with adding it. https://github.com/bottlerocket-os/twoliter/issues/152

Edit: no that's dumb. I can safely add this.

Edit, edit: Actually I'm still squeamish... deferring.

PR #137, opened 2024-01-23
testsys: Add test type for cluster templating

ecpullen:
**Issue number:**

Closes #N/A

**Description of changes:**

Adds 2 additional fields for templating cluster config files. This allows us to provision separate clusters for each test type.

```
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig
metadata:
  name: "{{kube-arch}}-{{kube-variant}}-ipv6-{{test-flavor}}"
  version: "{{version}}"
  region: "us-west-2"
kubernetesNetworkConfig:
  ipFamily: IPv6
```

**Testing done:**

TBA

**Terms of contribution:**

### DIFF
--- a/tools/testsys/src/crds.rs
+++ b/tools/testsys/src/crds.rs
@@ -486,8 +486,14 @@ pub(crate) trait CrdCreator: Sync {
                         &mut self
                             .additional_fields(&test_type.to_string())
                             .into_iter()
-                            // Add the image id in case it is needed for cluster creation
+                            // Add the image id, original test type, and resolved test type in case
+                            // it is needed for cluster creation
                             .chain(Some(("image-id".to_string(), image_id.clone())))
+                            .chain(Some((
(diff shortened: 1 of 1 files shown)


webern (approved):

tools/testsys/src/crds.rs
```diff
@@ -488,6 +488,11 @@ pub(crate) trait CrdCreator: Sync {
                             .into_iter()
                             // Add the image id in case it is needed for cluster creation
```
arnaldo2792: Comment should reflect the change

arnaldo2792 (commented):
Folks might be already familiar with how this change affects testsys, but it will be nice to add a line or two on the PR description to help reviewers understand why this is needed or how this will improve the testsys experience.

arnaldo2792 (approved):

PR #141, opened 2024-01-25
enable lookaside cache arguments

webern:
**Issue number:**

Closes #136 
Closes #135 

**Description of changes:**

    twoliter: accept lookaside cache arguments

    Allow the user to change the lookaside cache URL and upstream sources
    fallback in the twoliter build variant command.

    buildsys: get lookaside cache from the environment

    Out-of-tree builders will want their own lookaside cache for sources.
    This makes it possible to specify the location of the lookaside cache
    instead of hard-coding it in buildsys.

**Testing done:**

- [x] Built `tests/projects/project1` with `twoliter build variant`. This does not test lookaside cache logic, but checks for regression.

Built Bottlerocket

- [x] `cargo make` succeeds
- [x] `BUILDSYS_LOOKASIDE_CACHE=https:://example.com cargo make` fails as expected
- [x] `BUILDSYS_LOOKASIDE_CACHE=https:://example.com BUILDSYS_UPSTREAM_SOURCE_FALLBACK=true` succeeds

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #146, opened 2024-01-29
buildsys: revert removal of force-upstream

webern:
46fe1b8 in #141 incorrectly assumed the force-upstream field was unused. Revert the change.


bcressey (approved):

yeazelm (approved):

webern:
I was able to build an nvidia variant, which uses this feature, so it's good. Merging.

PR #156, opened 2024-02-21
buildsys: fix image features

### LINKED ISSUE #153: image features aren't passed to package builds correctly
In what looks like fallout from 92bc3bbe40b816b554020f90aa37542f1ca3ec51, I built `aws-dev` and ended up with a system using cgroups v1 (instead of v2) and wicked (instead of networkd).

`systemd` and `os` are the only packages that rely on the `package-features` capability that lets them opt into tracking certain variant features.

Notably, the variant did have an XFS data partition and Secure Boot enabled, so it seems like it's just the packages that aren't being built correctly.

webern:
**Issue number:**

Closes #153 

**Description of changes:**

This commt fixes a couple of regressions introduced in 92bc3bb. Most notably, the wrong set of image_features was being used to set package build args. In this commit we correctly use the image features found in the downstream variants Cargo.toml.

Also fixes an oversight where an error constructing the Builder object would have resulted in a panic.

**Testing done:**

- @bcressey built an `aws-dev` image and observed that the problem was fixed.

**Terms of contribution:**

### DIFF
--- a/tools/buildsys/src/builder.rs
+++ b/tools/buildsys/src/builder.rs
@@ -114,7 +114,12 @@ impl CommonBuildArgs {
 }
 
 struct PackageBuildArgs {
-    image_features: Option<HashSet<ImageFeature>>,
+    /// The package might need to know what the `image_features` are going to be for the variant
+    /// it is going to be used in downstream. This is because certain packages will be built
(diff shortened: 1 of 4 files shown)


bcressey (approved):
LGTM. The fix worked for me locally when testing an `aws-dev` build.

ecpullen (approved):
LGTM

PR #159, opened 2024-02-22
buildsys: avoid trailing slash for cache lookup

### LINKED ISSUE #158: blobs can't be fetched from the lookaside cache
I'm getting the following error when attempting to build Bottlerocket with the current tip of develop:
```
  ExternalFileFetch: Failed to fetch 'https://cache.bottlerocket.aws//acl-2.3.2.tar.gz/31aeb39958d8af5d08933dd3a89333a41025c3eb49fc461fa3c291caca51dad575ec13faeb7deba9b3c2ebf7615be7d45e2b78e50d4f83d8ec933c95931a7682/acl-2.3.2.tar.gz': 403 Forbidden
```

Note the doubled `//` in the URL path, which S3 won't like.

bcressey:
**Issue number:**

Closes #158

**Description of changes:**
The lookaside cache URL is now a `Url` and may have a trailing slash when formatted as a string. Appending a second slash could produce an invalid URL. Avoid this by using `Url::join` instead.

**Testing done:**
Built `aws-dev` and was able to retrieve missing blobs.

**Terms of contribution:**

### DIFF
--- a/tools/buildsys/src/cache.rs
+++ b/tools/buildsys/src/cache.rs
@@ -68,11 +68,20 @@ impl LookasideCache {
                 }
             }
 
-            let name = path.display();
+            let name = &path.display().to_string();
             let tmp = PathBuf::from(format!(".{}", name));
 
             // first check the lookaside cache
-            let url = format!("{}/{}/{}/{}", self.lookaside_cache, name, hash, name);
+            let mut url = self.lookaside_cache.clone();
+            url.path_segments_mut()
+                .map_err(|_| {
+                    error::UrlPathSegmentsSnafu {
+                        url: self.lookaside_cache.clone(),
+                    }
+                    .build()
+                })?
+                .extend([name, hash, name]);
+            let url = url.to_string();
             match self.fetch_file(&url, &tmp, hash) {
                 Ok(_) => {
                     fs::rename(&tmp, path)
--- a/tools/buildsys/src/cache/error.rs
+++ b/tools/buildsys/src/cache/error.rs
@@ -50,6 +50,9 @@ pub(crate) enum Error {
 
     #[snafu(display("Failed to delete file '{}': {}", path.display(), source))]
     ExternalFileDelete { path: PathBuf, source: io::Error },
+
+    #[snafu(display("Failed to get path segments from URL '{}'", url))]
+    UrlPathSegments { url: String },
 }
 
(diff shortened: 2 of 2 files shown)


webern (approved):
Ugh, that bug again (it took forever to get it right in tough and tuftool)

tools/buildsys/src/cache.rs
```diff
@@ -72,7 +72,15 @@ impl LookasideCache {
...
-            let url = format!("{}/{}/{}/{}", self.lookaside_cache, name, hash, name);
+            let cache_key = format!("{}/{}/{}", name, hash, name);
+            let url = self
+                .lookaside_cache
+                .join(&cache_key)
+                .context(error::UrlJoinSnafu {
+                    path: cache_key,
+                    base: self.lookaside_cache.to_owned(),
+                })?
+                .to_string();
```
ecpullen: Was there a reason for this implementations vs the one below other than the extra error management?

```suggestion
           let url = self
                .lookaside_cache
                .join(&name)
                .context(error::UrlJoinSnafu {
                    path: name,
                    base: self.lookaside_cache.to_owned(),
                })?.join(&hash)
                .context(error::UrlJoinSnafu {
                    path: hash,
                    base: self.lookaside_cache.to_owned(),
                })?.join(&name)
                .context(error::UrlJoinSnafu {
                    path: name,
                    base: self.lookaside_cache.to_owned(),
                })?
                .to_string();
```

ecpullen (approved):

tools/buildsys/src/cache.rs
bcressey: > Was there a reason for this implementations vs the one below other than the extra error management?

Mostly that, yeah. I poked around the url crate a bit and came up with what looks like a more idiomatic approach.

bcressey:
Tested the above with the following:
```
# current setting works
BUILDSYS_LOOKASIDE_CACHE = "https://cache.bottlerocket.aws"

# trailing slash works
BUILDSYS_LOOKASIDE_CACHE = "https://cache.bottlerocket.aws/"

# repeated slash doesn't work
BUILDSYS_LOOKASIDE_CACHE = "https://cache.bottlerocket.aws//"
```

This seems like the right behavior to me - the first two are reasonable values that should be treated the same way, and the third is odd but might be wanted for a backend where every key is prefixed with a `/`.

ecpullen (approved):

PR #166, opened 2024-03-04
switch to the unified SDK

bcressey:
**Issue number:**
Related: https://github.com/bottlerocket-os/bottlerocket-sdk/pull/156

**Description of changes:**
Adapt `Dockerfile` and `Makefile.toml` to the breaking changes in the unified SDK - merge the right architecture's macros prior to building RPMs, and add a subset of the target's binutils to `$PATH` so that `find-debuginfo.sh` works.

Update `twoliter` to remove the need to suffix the SDK image with the target's architecture, and to get rid of the toolchain image entirely.

Fix the alpha SDK script to generate two images rather than the previous four.

**Testing done:**
Used the updated `twoliter` to build Bottlerocket images with this `Twoliter.toml`:
```
schema-version = 1
release-version = "1.19.3"

[sdk]
registry = "public.ecr.aws/c2b5m1s5"
repo = "thar-be-a-beta-sdk"
tag = "v0.40.0
```

**Terms of contribution:**

### DIFF
--- a/twoliter/Cargo.toml
+++ b/twoliter/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "twoliter"
-version = "0.0.7"
+version = "0.1.0"
 edition = "2021"
 description = "A command line tool for creating custom builds of Bottlerocket"
 authors = ["Matthew James Briggs <brigmatt@amazon.com>"]
--- a/twoliter/src/cmd/make.rs
+++ b/twoliter/src/cmd/make.rs
@@ -39,7 +39,7 @@ impl Make {
         let toolsdir = project.project_dir().join("build/tools");
         install_tools(&toolsdir).await?;
         let makefile_path = toolsdir.join("Makefile.toml");
-        CargoMake::new(&project, &self.arch)?
+        CargoMake::new(&project)?
             .env("CARGO_HOME", self.cargo_home.display().to_string())
             .env("TWOLITER_TOOLS_DIR", toolsdir.display().to_string())
             .env("BUILDSYS_VERSION_IMAGE", project.release_version())
--- a/twoliter/src/docker/image.rs
+++ b/twoliter/src/docker/image.rs
@@ -1,17 +1,17 @@
+use serde::{Deserialize, Serialize};
 use std::fmt::{Display, Formatter};
 
 /// Represents a docker image URI such as `public.ecr.aws/myregistry/myrepo:v0.1.0`. The registry is
 /// optional as it is when using `docker`. That is, it will be looked for locally first, then at
 /// `dockerhub.io` when the registry is absent.
-#[derive(Debug, Default, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
-#[allow(unused)]
+#[derive(Debug, Default, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Deserialize, Serialize)]
 pub(crate) struct ImageUri {
     /// e.g. public.ecr.aws/bottlerocket
-    registry: Option<String>,
+    pub(crate) registry: Option<String>,
     /// e.g. my-repo
-    repo: String,
+    pub(crate) repo: String,
     /// e.g. v0.31.0
-    tag: String,
+    pub(crate) tag: String,
 }
 
 impl ImageUri {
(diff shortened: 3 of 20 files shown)


twoliter/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "twoliter"
-version = "0.0.6"
+version = "0.1.0"
```
bcressey, line 3: I can back out this commit if you'd prefer to do the version bump elsewhere. I'd like to leave a gap between `0.0.6` and the new version if possible, in case we need to do a point release that works with the older SDKs.

twoliter/src/test/data/Twoliter-1.toml
```diff
@@ -3,10 +3,5 @@ release-version = "1.0.0"
...
 registry = "a.com/b"
-name = "my-bottlerocket-sdk"
-version = "v1.2.3"
-
-[toolchain]
-registry = "c.co/d"
-name = "toolchainz"
-version = "v3.4.5"
+repo = "my-bottlerocket-sdk"
+tag = "v1.2.3"
```
bcressey, line 7: @webern we're taking a breaking change for this, so if this isn't the format you actually want, let me know and I can  change it.

twoliter/Cargo.toml
webern: Seems like a good idea to tie it to this PR.

twoliter/src/cmd/make.rs
```diff
@@ -52,7 +52,7 @@ impl Make {
 
 #[test]
 fn test_trailing_args_1() {
-    let args = Make::try_parse_from(&[
```
webern: Clippy?

twoliter/src/test/data/Twoliter-1.toml
webern: > we're taking a breaking change for this, so if this isn't the format you actually want, let me know and I can change it.

The only question I have is whether we would prefer `sdk = "a.com/b/my-bottlerocket-sdk:v1.2.3"` since we no longer need to do any processing on this. I can go either way, doesn't seem important.

twoliter/src/docker/image.rs
```diff
@@ -1,132 +1,67 @@
...
 /// Represents a docker image URI such as `public.ecr.aws/myregistry/myrepo:v0.1.0`. The registry is
 /// optional as it is when using `docker`. That is, it will be looked for locally first, then at
 /// `dockerhub.io` when the registry is absent.
-#[derive(Debug, Default, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
-#[allow(unused)]
+#[derive(Debug, Default, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Deserialize, Serialize)]
 pub(crate) struct ImageUri {
     /// e.g. public.ecr.aws/bottlerocket
-    registry: Option<String>,
+    pub(crate) registry: Option<String>,
```
webern, line 10: Are these `pub(crate)`s needed?

webern (commented):
I'm surprised not to see a change in rpm2kmodkit. Is it because the SDK has the toolchains in the same locations that rpm2kmodkit was expecting them?

webern:
> I'm surprised not to see a change in rpm2kmodkit. Is it because the SDK has the toolchains in the same locations that rpm2kmodkit was expecting them?

Ok, I see the relevant changes in `FROM repobuild as kmodkitbuild`.

webern (approved):

bcressey:
> Ok, I see the relevant changes in `FROM repobuild as kmodkitbuild`.

I have more changes in mind for kmod kit and migrations - merging them into `rpm2img` most likely. But that can wait for another PR since I expect it'll involve packaging changes in the main repo too.

twoliter/src/test/data/Twoliter-1.toml
ecpullen: I think a single sdk string would create a clearer interface.

ecpullen (approved):

twoliter/src/cmd/make.rs
bcressey: > Clippy?

Yes - specifically, `cargo clippy --tests`. We may want to add that to the `Makefile` clippy target.

twoliter/src/test/data/Twoliter-1.toml
bcressey: > The only question I have is whether we would prefer sdk = "a.com/b/my-bottlerocket-sdk:v1.2.3" since we no longer need to do any processing on this.

@webern would you want this to become a top-level key rather than a nested map? E.g.
```
❯ cat Twoliter.toml
schema-version = 1
release-version = "1.19.2"
sdk = "a.com/b/my-bottlerocket-sdk:v1.2.3"
```

bcressey:
Per discussion with @webern, going to leave the SDK structure as it is here and not go all the way to a single string yet. We can revisit as part of the kits integration.

bcressey:
⬆️ force push to rebase and fix the conflicts in `Cargo.lock` and `Cargo.toml`

ginglis13 (approved):
LGTM, changes worked in testing https://github.com/bottlerocket-os/bottlerocket-sdk/pull/165

PR #179, opened 2024-03-25
twoliter: use Twoliter.toml for cache layers

bcressey:
**Issue number:**
N/A

**Description of changes:**
Twoliter.toml should always exist, while .gitignore might not.

**Testing done:**
Built the main repo with a custom Twoliter build.

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/Dockerfile
+++ b/twoliter/embedded/Dockerfile
@@ -31,7 +31,7 @@ ARG TOKEN
 # We can't create directories via RUN in a scratch container, so take an existing one.
 COPY --chown=1000:1000 --from=sdk /tmp /cache
 # Ensure the ARG variables are used in the layer to prevent reuse by other builds.
-COPY --chown=1000:1000 .gitignore /cache/.${PACKAGE}.${ARCH}.${TOKEN}
+COPY --chown=1000:1000 Twoliter.toml /cache/.${PACKAGE}.${ARCH}.${TOKEN}
 
 # =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^=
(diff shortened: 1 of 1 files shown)


twoliter/embedded/Dockerfile
```diff
@@ -31,7 +31,7 @@ ARG TOKEN
 # We can't create directories via RUN in a scratch container, so take an existing one.
 COPY --chown=1000:1000 --from=sdk /tmp /cache
 # Ensure the ARG variables are used in the layer to prevent reuse by other builds.
-COPY --chown=1000:1000 .gitignore /cache/.${PACKAGE}.${ARCH}.${TOKEN}
+COPY --chown=1000:1000 Twoliter.toml /cache/.${PACKAGE}.${ARCH}.${TOKEN}
```
jmt-lab, line 34: This change looks fine to me, but for my educational purposes is there some sort of corelation between .gitignore and Twoliter.toml (duplicate data being stored)

jmt-lab (approved):

webern (approved):

webern:
`.gitignore` is, to some extent, required in order to ignore `build`, `.cargo` and such. But this makes sense anyway since Twoliter.toml really is constrained to exist in the root.

twoliter/embedded/Dockerfile
bcressey: No relationship really, they are both standing in for "a file we're sure will exist". Except with Twoliter.toml we are now guaranteed it will exist.

Scratch containers are kind of hard to work with - there's no shell, so you can only copy files from the build context or other images.

Strictly speaking the cache layer also doesn't have to be a totally empty layer; we could use the SDK itself and just add a file. But I liked the idea that the cache layer would only have files that some later stage added, so that you could `find /home/builder/.cache` and not get back a giant list of files.

PR #181, opened 2024-03-28
add FIPS-related functionality

bcressey:
**Issue number:**
Related: https://github.com/bottlerocket-os/bottlerocket/issues/1667

**Description of changes:**
The overall aim of this series is to expose the variant's image features as (RPM) package-level metadata, so that packages can add sub-packages that affinitize (or anti-affinitize) to those features. This avoids the need to hardcode the exact list of packages required to support a particular feature. In the case of FIPS, a hardcoded list is unworkable, since most Go binaries need to be built to use FIPS and non-FIPS crypto, and different variants install a different set of Go binaries - most obviously, `kubelet` or `ecs-agent`.

I've added support for bootconfig drop-ins as well: any such drop-ins will be collected and used to populate the default bootconfig file, which was previously empty. In combination with the metadata feature, this can be used so that the bootconfig snippet `kernel.fips = 1` is installed if the `fips` image feature is set. We can also use it to eliminate the conditional compilation of systemd, by adding the bootconfig snippet `init.systemd.unified_cgroup_hierarchy = 1` if the `unified-cgroup-hierarchy` feature is set.

I plan to build on the image feature metadata functionality in follow-up changes, for example to support feature flags like `in-place-updates`, `host-containers`, and `kmod-development` that control whether the related packages are installed. These may also involve changes to `rpm2img`, but my hope is to do that sparingly. For example, `in-place-updates` might control whether the image has one or two partition banks, and whether a migrations archive is produced, but otherwise all the functionality would come through the installation of optional packages.

**Testing done:**
My [fips-and-non-fips](https://github.com/bcressey/bottlerocket/commits/fips-and-non-fips/) branch shows how this feature is used to arrange for the correct set of Go binaries to be installed depending on the image feature.

These two commits demonstrate how bootconfig drop-ins will work:
* [systemd: use bootconfig for unified cgroup hierarchy](https://github.com/bcressey/bottlerocket/commit/2af0c7bfb33bc5c7a6ee88ef3bdb8cd28fb509b1)
* [release: add subpackage for fips](https://github.com/bcressey/bottlerocket/commit/d42d8bceae4e380b9052b4f72480537aea325580)

**Terms of contribution:**

### DIFF
--- a/tools/buildsys/src/args.rs
+++ b/tools/buildsys/src/args.rs
@@ -134,6 +134,18 @@ pub(crate) struct BuildVariantArgs {
     #[arg(long, env = "BUILDSYS_VARIANT")]
     pub(crate) variant: String,
 
+    #[arg(long, env = "BUILDSYS_VARIANT_PLATFORM")]
+    pub(crate) variant_platform: String,
+
+    #[arg(long, env = "BUILDSYS_VARIANT_RUNTIME")]
+    pub(crate) variant_runtime: String,
+
+    #[arg(long, env = "BUILDSYS_VARIANT_FAMILY")]
+    pub(crate) variant_family: String,
+
+    #[arg(long, env = "BUILDSYS_VARIANT_FLAVOR")]
+    pub(crate) variant_flavor: String,
+
     #[arg(long, env = "BUILDSYS_VERSION_BUILD")]
     pub(crate) version_build: String,
 
--- a/tools/buildsys/src/manifest.rs
+++ b/tools/buildsys/src/manifest.rs
@@ -233,6 +233,15 @@ when the platform supports it.
 uefi-secure-boot = true
 ```
 
+`fips` means that FIPS-certified modules will be used for cryptographic operations. This affects
+the kernel at runtime. It also causes alternate versions of Go and Rust programs that use
+FIPS-compliant ciphers to be included in the image.
+
+```ignore
+[package.metadata.build-variant.image-features]
+fips = true
+```
+
 */
 
 mod error;
@@ -529,6 +538,7 @@ pub enum ImageFeature {
     UnifiedCgroupHierarchy,
     XfsDataPartition,
     UefiSecureBoot,
+    Fips,
 }
 
 impl TryFrom<String> for ImageFeature {
@@ -540,6 +550,7 @@ impl TryFrom<String> for ImageFeature {
             "unified-cgroup-hierarchy" => Ok(ImageFeature::UnifiedCgroupHierarchy),
             "xfs-data-partition" => Ok(ImageFeature::XfsDataPartition),
             "uefi-secure-boot" => Ok(ImageFeature::UefiSecureBoot),
+            "fips" => Ok(ImageFeature::Fips),
             _ => error::ParseImageFeatureSnafu { what: s }.fail()?,
         }
     }
@@ -553,6 +564,7 @@ impl fmt::Display for ImageFeature {
             ImageFeature::UnifiedCgroupHierarchy => write!(f, "UNIFIED_CGROUP_HIERARCHY"),
             ImageFeature::XfsDataPartition => write!(f, "XFS_DATA_PARTITION"),
             ImageFeature::UefiSecureBoot => write!(f, "UEFI_SECURE_BOOT"),
+            ImageFeature::Fips => write!(f, "FIPS"),
         }
     }
 }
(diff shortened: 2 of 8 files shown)


tools/buildsys/src/args.rs
```diff
@@ -134,6 +134,18 @@ pub(crate) struct BuildVariantArgs {
...
+    pub(crate) variant_platform: String,
+
+    #[arg(long, env = "BUILDSYS_VARIANT_RUNTIME")]
+    pub(crate) variant_runtime: String,
+
+    #[arg(long, env = "BUILDSYS_VARIANT_FAMILY")]
+    pub(crate) variant_family: String,
+
+    #[arg(long, env = "BUILDSYS_VARIANT_FLAVOR")]
+    pub(crate) variant_flavor: String,
```
bcressey, line 147: Here I opted to add these to `BuildVariantArgs`, even though they're also present in `BuildPackageArgs`, because my goal is to eventually eliminate them from `BuildPackageArgs` when conditional package builds are eliminated.

tools/buildsys/src/manifest.rs
```diff
@@ -233,6 +233,15 @@ when the platform supports it.
 uefi-secure-boot = true
 ```
 
+`fips` means that FIPS-certified modules will be used for cryptographic operations. This affects
+the kernel at runtime. It also causes alternate versions of Go and Rust programs that use
```
bcressey, line 237: I am kind of over-promising here since I'm still not 100% sure what the Rust integration will look like.

twoliter/embedded/Dockerfile
```diff
@@ -158,18 +176,35 @@ FROM sdk AS repobuild
...
 RUN --mount=target=/host \
     mkdir -p /local/rpms ./rpmbuild/RPMS \
     && ln -s /host/build/rpms/*.rpm ./rpmbuild/RPMS \
+    && ln -s /home/builder/rpmbuild/RPMS/*/*.rpm ./rpmbuild/RPMS \
     && createrepo_c \
         -o ./rpmbuild/RPMS \
         -x '*-debuginfo-*.rpm' \
         -x '*-debugsource-*.rpm' \
         --no-database \
-        /host/build/rpms \
```
bcressey, line 172: Previously it didn't matter whether we used `/host/build/rpms` or `./rpmbuild/RPMS` to generate the repo, but now that we build an RPM in the course of building the variant, we have to use the latter because it's writable.

twoliter/embedded/Dockerfile
```diff
@@ -179,7 +214,7 @@ RUN --mount=target=/host \
         --downloadonly \
         --downloaddir . \
         --forcearch "${ARCH}" \
-        install $(printf "bottlerocket-%s\n" ${PACKAGES}) \
+        install $(printf "bottlerocket-%s\n" metadata ${PACKAGES}) \
```
bcressey, line 217: `printf` is prefixing everything here with `bottlerocket-`, so the package is actually called `bottlerocket-metadata`.

twoliter/embedded/metadata.spec
```diff
@@ -0,0 +1,64 @@
+%global cross_generate_attribution %{nil}
```
bcressey, line 1: I didn't want licenses for the metadata package to show up on the image - it's kind of an implementation quirk - which is why this is set to `%{nil}`.

I thought about force-erasing the RPM so it didn't show up in the generated application-inventory.json, or filtering it out, but that seemed less important and a bit too "magical".

tools/buildsys/src/manifest.rs
jmt-lab: Just for arguement's sake and clarification, is this below Cargo.toml add for the variant/rpm package or is it for rust code, and if for rust code is there a reason not to translate it to a standardized feature flag?

jmt-lab (approved):

tools/buildsys/src/manifest.rs
bcressey: > Just for arguement's sake and clarification, is this below Cargo.toml add for the variant/rpm package or is it for rust code, and if for rust code is there a reason not to translate it to a standardized feature flag?

It's for the buildsys-related metadata for variants. I'd like to keep it separate from Cargo's own functionality because - as distinctive and delightful as Bottlerocket's use of `cargo` is - eventually I expect to create our own tool to do what it does for us now (tracking changes, traversing the dependency graph, driving builds).

arnaldo2792 (approved):

webern (approved):

Issue #185, opened 2024-04-05
kits: implement recursive dependency search


webern:
** Description **

When building kits and packages, `buildsys` is going to need information from the entire cargo dependency tree for the kit or package that it is building (the algorithm this is being used for will be introduced elsewhere). We need an algorithm that can recurse through Cargo.toml files to create a dependency graph. This graph should distinguish between nodes of the graph that are a kit and those that are a package.

** Definition of Done **

The algorithm exists and can be demonstrated in a unit test on `kits` and `packages` directories.

Needs
- [x] #189

PR #190, opened 2024-04-09
update local-kit test for kit development

### LINKED ISSUE #189: enhance local-kits test project
Add a few packages to the local-kits test project. These should exercise the following types of dependencies:

* Kits that depend on packages
* A kit that depends on another kit
* A package that depends on a kit
* A package that depends on another package

Related to #17 and #185

Example:

- `pkga` depends on nothing
- `pkgb` requires `pkga`
- `pkgc` requires `pkga`
- `pkgd` requires `pkgc`

`core-kit` contains
* `pkga`

`extra-kit-1` contains
* `pkgb`
* `pkgd`

`extra-kit-2` contains
* `pkgc`

webern:
**Issue number:**

Closes #189

**Description of changes:**

It's hard to know if I did all of this correctly, but this provides a test bed for #185 and beyond.

**Testing done:**

I built all of the packages before changing the dependency tree over to using kits (which don't build yet!)

**Terms of contribution:**


tests/projects/local-kit/kits/core-kit/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "hello-kit"
+name = "core-kit"
```
jmt-lab, line 2: Are we still planning on modifying the Cargo.toml file for Kits, My impression was that twoliter would have its own project toml that would also include kit config stuff.

tests/projects/local-kit/kits/core-kit/Cargo.toml
```diff
@@ -9,14 +9,11 @@ exclude = ["README.md"]
 
 [package.metadata.build-kit]
 included-packages = [
-    # in-tree packages
-    "hello-agent",
-    "hello-go",
+    "pkg-a",
```
jmt-lab, line 12: Are we going to allow users to build a kit from segments of another kit (As in a user wants everything in core kit cept the kernel which they want to add their own). If so this should probably be a table of some sort so that version requirements could be specified.

tests/projects/local-kit/kits/core-kit/Cargo.toml
webern: Kits will have Cargo.toml files like Variants and Packages do because of the way we are using the Cargo build system to drive our builds.

tests/projects/local-kit/kits/core-kit/Cargo.toml
webern: We aren't considering adding versioning at the level of package name right now. The versioning will take place at the level of kit metadata (we think).

tests/projects/local-kit/packages/pkg-a/pkg-a.spec
```diff
@@ -0,0 +1,28 @@
+%global _cross_first_party 1
+#%undefine _debugsource_packages
```
bcressey: You have to escape macros in comments or else they still have an effect:
```suggestion
#%%undefine _debugsource_packages
```

But you may not want that here since there's no actual build happening, so I would just remove the `#`.

tests/projects/local-kit/packages/pkg-a/pkg-a.spec
```diff
@@ -0,0 +1,28 @@
...
+Name: %{_cross_os}pkg-a
+Version: 0.0
+Release: 0%{?dist}
+Summary: pkg-a
+License: Apache-2.0 OR MIT
+URL: https://github.com/bottlerocket-os/bottlerocket
+
+Source100: pkg-a.txt
+
+Requires: %{_cross_os}pkg-a
```
bcressey: package shouldn't require itself:
```suggestion
```

tests/projects/local-kit/packages/pkg-b/pkg-b.spec
```diff
@@ -0,0 +1,29 @@
...
+Name: %{_cross_os}pkg-b
+Version: 0.0
+Release: 0%{?dist}
+Summary: pkg-b
+License: Apache-2.0 OR MIT
+URL: https://github.com/bottlerocket-os/bottlerocket
+
+Source100: pkg-b.txt
+
+Requires: %{_cross_os}pkg-b
```
bcressey:
```suggestion
```

tests/projects/local-kit/packages/pkg-b/Cargo.toml
```diff
@@ -1,20 +1,19 @@
...
-source-groups = ["hello-agent"]
+source-groups = []
 
 [lib]
 path = "../packages.rs"
 
 # RPM BuildRequires
 [build-dependencies]
-# None
+core-kit = { path = "../../kits/core-kit" }
```
bcressey, line 16: This is one way of having a kit dependency come in: by direct dependency of a package crate on a kit crate.

The other way would be indirectly: a package crate that is only built for a kit crate that depends on another kit crate. Then we could have a rule like "all packages in the k8s kit have access to packages in the core kit because the k8s kit depends on core kit".

But I think that's not supported by the current direction - we need each package to list its kit dependencies.

tests/projects/local-kit/packages/pkg-c/pkg-c.spec
```diff
@@ -0,0 +1,29 @@
...
+Name: %{_cross_os}pkg-c
+Version: 0.0
+Release: 0%{?dist}
+Summary: pkg-c
+License: Apache-2.0 OR MIT
+URL: https://github.com/bottlerocket-os/bottlerocket
+
+Source100: pkg-c.txt
+
+Requires: %{_cross_os}pkg-c
```
bcressey:
```suggestion
```

tests/projects/local-kit/kits/extra-1-kit/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+    "pkg-b",
+    "pkg-d",
+]
+
+[lib]
+path = "../kit.rs"
+
+[build-dependencies]
+pkg-b = { path = "../../packages/pkg-b" }
+pkg-d = { path = "../../packages/pkg-d" }
```
bcressey: Might be worth an `extra-3-kit` that includes `pkg-e` which:
* depends on `extra-1-kit` and `extra-2-kit` in its Cargo.toml
* BuildRequires `pkg-a`, `pkg-b`, `pkg-c`, `pkg -d` in its spec

That will test the implicit kit dependency we actually care about: `pkg-e` should have access to `core-kit` as an indirect dependency of `extra-1-kit` and `extra-2-kit`.

tests/projects/local-kit/variants/hello-ootb/Cargo.toml
```diff
@@ -15,12 +15,14 @@ xfs-data-partition = true
...
 kernel-parameters = []
 
 [lib]
 path = "../variants.rs"
 
 [build-dependencies]
-hello-kit = { path = "../../kits/hello-kit" }
+core-kit = { path = "../../kits/core-kit" }
+extra-1-kit = { path = "../../kits/extra-1-kit" }
+extra-2-kit = { path = "../../kits/extra-2-kit" }
```
bcressey: This should work with only one kit dependency right? Like if we add `extra-3-kit` that depends on one and two, and implicitly depends on core, then they should all be available for variant builds.

webern:
Combined changes addressing Ben's comments https://github.com/bottlerocket-os/twoliter/compare/94e39cd07a9a71a1737daca12de84b82eabbc307..0be2bfd9d235216d0046c4efc702a89b9bf1ee5c

tests/projects/local-kit/kits/extra-1-kit/Cargo.toml
webern: Good idea to add this.

tests/projects/local-kit/variants/hello-ootb/Cargo.toml
webern: Yes. I think it would be weird, but supported, to include a package from core-kit in the variant's list of packages, then rely on core-kit being a transient dependency of another kit. But it would work.

webern:
So I made some changes as best as I could. It might be hard to get this just right until the actual build starts taking place, so PRs for actual build code may tweak these dependency relationships.

webern:
These are some issues I found while working on the dependency graph algorithm https://github.com/bottlerocket-os/twoliter/compare/e186398188702135dacc6b1916c672c4adba8c3e..271add0a2a412bf3699f90160173fdb31b9cd07d

webern:
@sam-berning, @jmt-lab, @bcressey - I'm pretty sure this is ready to merge without a ton of follow-up tweaking in subsequent PRs.

bcressey (approved):

jmt-lab (approved):
LGTM

PR #198, opened 2024-04-13
implement a dependency graph algorithm for package and kit builds

### LINKED ISSUE #185: kits: implement recursive dependency search

webern:
**Issue number:**

Closes #185 

**Description of changes:**

For kit builds we are going to need to understand the dependency graph in `buildsys build-kit`. We may need it for other operations as well.

This PR changes buildsys so that it requires a cargo-metadata json file which is uses to understand the dependency graph and send the list of package dependencies and kit dependencies to the Dockerfile in build args.

This implementation requires that Twoliter projects be organized as in a single workspace with all packages, kits and variants in the same Cargo.toml (which feels like an improvement overall).

**Testing done:**

- [x] unit-testing on the in-tree test project `local-kit`
- [x] Build Bottlerocket (with changes to make it into one workspace): https://github.com/webern/bottlerocket/compare/8933289a4952a17f08f5b2b8f8547e23a853bd4c...webern:bottlerocket:workspace
- [x] Build an AMI
- [x] Rebuild the image to check incremental compilation (only os , release and the variant should be rebuilt) - (aws-iam-authenticator was also rebuilt, but that doesn't seem show-stopping)

**Terms of contribution:**


tests/projects/project1/packages/hello-agent/Cargo.toml
```diff
@@ -1,20 +1,19 @@
 [package]
-name = "hello-agent"
+name = "pkg-b"
```
bcressey: Maybe this file needs to be deleted?
```suggestion
name = "hello-agent"
```

tests/projects/project1/packages/hello-go/Cargo.toml
```diff
@@ -1,12 +1,12 @@
 [package]
-name = "hello-go"
+name = "pkg-a"
```
bcressey: This file seems like it should be deleted also, or if not:
```suggestion
name = "hello-go"
```

tools/buildsys/src/deps.rs
```diff
@@ -0,0 +1,241 @@
+// TODO - fix all clippy issues
+#![allow(clippy::all)]
+
+use crate::manifest::ManifestInfo;
+use std::collections::{HashMap, HashSet};
+use std::error::Error;
+use std::fmt::{Debug, Display, Formatter};
+use std::path::{Path, PathBuf};
+
+/// TODO - replace with Snafu even though I hate it.
```
bcressey: 🔥

tools/buildsys/src/deps.rs
```diff
@@ -0,0 +1,241 @@
...
+
+impl Error for TempError {}
+
+type Result<T> = std::result::Result<T, TempError>;
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Ord, PartialOrd)]
+pub enum ManifestType {
+    Package,
+    Kit,
+}
```
bcressey: Will this be integrated with the `Package` / `Metadata` structs in `manifest.rs` eventually?

tools/buildsys/src/deps.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// Whether this is a kit or a package.
+    _manifest_type: ManifestType,
+    /// The path to the manifest.
+    _manifest_path: PathBuf,
+    // // TODO - keep only the subset of info that is required.
+    // manifest: ManifestInfo,
+}
+
+pub struct Graph {
+    /// The starting point of this graph, i.e., the name of the kit or package that was the starting
```
bcressey: Curious whether you looked at the `guppy` crate.

Based on [this example](https://github.com/guppy-rs/guppy/blob/main/guppy/examples/remove_dev_only.rs), it looks like we could take the `cargo metadata` output (possibly generated once, at the start of the build), build a graph, find our crate in it, then iterate over it with a custom resolver.

webern:
https://github.com/bottlerocket-os/twoliter/compare/0e06bb6c06f354db5816821b215641906755982d..c150a5584c3fe542705869d4d7cd8a16864315d7

Rebase

tests/projects/project1/packages/hello-agent/Cargo.toml
webern: Errant change.

tests/projects/project1/packages/hello-go/Cargo.toml
webern: Errant change.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/c150a5584c3fe542705869d4d7cd8a16864315d7..665c8598b6ed8c9deb4fb58f5a5c9d54e29565b5

Now using Guppy and getting a bit closer to where we are going with this.

tests/projects/local-kit/packages/pkg-d/Cargo.toml
```diff
@@ -1,3 +1,6 @@
+[workspace]
+members = []
```
bcressey: Why is this workspace entry needed?

tools/buildsys/src/deps.rs
```diff
@@ -0,0 +1,196 @@
...
+/// gives a list of all the packages that are required when we are build a package, or all of the
+/// packages that should be included when building a kit.
+pub fn package_dependencies(manifest_path: impl AsRef<Path>) -> Result<Vec<String>> {
+    let manifest_path = manifest_path.as_ref();
+    let (name, manifest_type) = get_name_and_type(&manifest_path).unwrap();
+    // TODO - deserialize instead of using a MetadataCommand.
+    let graph = MetadataCommand::new()
+        .manifest_path(&manifest_path)
+        .build_graph()
+        .unwrap();
```
bcressey: Seems like this might address the workspace / cargo.lock issue - if we run `cargo metadata` and point it at the `variants` workspace and just use that file here.

tools/buildsys/src/deps.rs
```diff
@@ -0,0 +1,196 @@
...
+    Ok(kits)
+}
+
+fn is_manifest_type(package_metadata: &PackageMetadata, manifest_type: ManifestType) -> bool {
+    let metadata_table = package_metadata.metadata_table();
+    match manifest_type {
+        ManifestType::Package => metadata_table.get("build-package").is_some(),
+        ManifestType::Kit => metadata_table.get("build-kit").is_some(),
+    }
+}
```
bcressey: Not sure where this fix would go, but we have the problem today that the manifest is very loose and will allow both `build-package` and `build-variant` tables to exist in the same file without throwing an error. So we might end up confused here if there's a malformed Cargo.toml.

tools/buildsys/src/deps.rs
```diff
@@ -0,0 +1,196 @@
...
+
+/// Because the PackageGraph uses PackagedID, and it is not easy for us to calculate the PackageID
+/// ourselves, we need to know the name of our manifest so that we can look up its PackageID in the
+/// graph. The simplest way to do that seems to be to deserialize into TOML and look for the
+/// `package.name` key.
+fn get_name_and_type(manifest_path: impl AsRef<Path>) -> Result<(String, ManifestType)> {
+    let path = manifest_path.as_ref();
+    let contents = fs::read_to_string(&path).unwrap();
+    let table: Table = toml::from_str(&contents).unwrap();
+    let package_table = table.get("package").unwrap().as_table().unwrap();
```
bcressey: You could turn this into a `ManifestInfo` - probably we should have a helper method there that returns the preferred package name based on the crate name or metadata override, and another helper to return the type.

tests/projects/local-kit/packages/pkg-d/Cargo.toml
webern: > Why is this workspace entry needed?

Because Twoliter has a workspace above it causing an error along the lines of "the package thinks its in a workspace when it isn't" when running a command (e.g. cargo metadata) in the pkg-d directory.

The fix is probably to require a global workspace for the whole of the Twoliter project, but that might interact poorly with a `sources` workspace(TBD). (And you would definitely want a separate `sources` workspace so that you can build your first-party Rust code, i.e. for development work, without all of the buildsys machinery in build.rs files taking place.

tools/buildsys/src/deps.rs
webern: > Not sure where this fix would go, but we have the problem today that the manifest is very loose and will allow both build-package and build-variant tables to exist in the same file without throwing an error. So we might end up confused here if there's a malformed Cargo.toml.

I'm thinking I want to defer this tightening just due to the amount of code I'm about to bomb.

tools/buildsys/src/deps.rs
webern: Done!

tests/projects/local-kit/packages/pkg-d/Cargo.toml
webern: Now this PR is using "One Workspace to Rule them All"

tools/buildsys/src/deps.rs
webern: Kinda ended up doing this although I introduced another type named `Manifest` that includes the graph, and left `ManifestInfo` unchanged. (This is because `pubsys` and `bottlerocket-variant` are using `ManifestInfo` in a way where the graph isn't needed.)

webern:
Current status: encountered during a Bottlerocket build at the `filesystem` package:

```
  --- stderr
  Unable to instantiate the builder: Failed to create build arguments due to a dependency error: Expected to find one of build-package, build-kit, or build-variant in package.metadata.
```

Turns out that Cargo.toml doesn't have `build-package` in it. I will fix it and continue on Monday

```toml
[package]
name = "filesystem"
version = "0.1.0"
edition = "2021"
publish = false
build = "../build.rs"

[lib]
path = "../packages.rs"
```

webern:
Over the weekend I made some changes to the Bottlerocket diff and the testing passed. This is ready for review.

webern:
> https://github.com/bottlerocket-os/twoliter/compare/596002d362144e6b941596234301c927ad38f061..e5ee4e19e43c322a92613c73fd1cfe7e8bf54d40

Improvement to the algorithm. For package build, for the top-level manifest, we should only consider `build-dependencies`. When lower in the tree/graph, we should consider both `build-dependencies` and `dev-dependencies`.

tests/projects/local-kit/kits/extra-1-kit/Cargo.toml
```diff
@@ -4,14 +4,8 @@ version = "0.1.0"
...
 publish = false
 build = "../build.rs"
-# Don't rebuild crate just because of changes to README.
-exclude = ["README.md"]
 
 [package.metadata.build-kit]
-included-packages = [
-    "pkg-b",
-    "pkg-d",
-]
```
bcressey, line 14: Are we not doing `included-packages` for kits any more?

It might be more natural to support an optional `excluded-packages` instead, if the default should be that all packages go into the kit.

twoliter/embedded/Makefile.toml
```diff
@@ -1471,16 +1493,20 @@ dependencies = [
...
-for ws in sources; do
-  [ -d "${ws}" ] || continue
-  cargo clean --manifest-path ${ws}/Cargo.toml
-done
+
+if [ -f "${BUILDSYS_SOURCES_DIR}/Cargo.toml" ]; then
+    cargo clean --manifest-path "${BUILDSYS_SOURCES_DIR}/Cargo.toml"
+fi
+
+cargo clean --manifest-path "${BUILDSYS_ROOT_DIR}/Cargo.toml"
```
bcressey: nit: this should go into the `clean-packages` stage

tools/buildsys/src/args.rs
```diff
@@ -157,65 +161,79 @@ pub(crate) struct BuildVariantArgs {
...
 
 #[test]
 fn build_type_includes_test() {
     // true
-    assert!(BuildType::Package.includes(PACKAGE | VARIANT));
-    assert!(BuildType::Variant.includes(VARIANT));
-    assert!(BuildType::Variant.includes(VARIANT | PACKAGE));
+    assert!(BuildFlags::Package.includes(PACKAGE | VARIANT));
+    assert!(BuildFlags::Variant.includes(VARIANT));
+    assert!(BuildFlags::Variant.includes(VARIANT | PACKAGE));
```
bcressey, line 217: I think we need kits to include the same build flags that packages do, except we should rebuild on a change to `BUILDSYS_KITS_DIR` vs `BUILDSYS_PACKAGES_DIR`.

tools/buildsys/src/manifest.rs
```diff
@@ -358,16 +488,73 @@ impl ManifestInfo {
...
+/// We do not want this, we want only a list of things that it depends on. Here we convert
+/// `PackageMetadata` objects to the `String` name, and filter out the "top-level manifest".
+fn filter_map_to_name(top_manifest_name: &str, pkg_metadata: &PackageMetadata) -> Option<String> {
+    if pkg_metadata.name() == top_manifest_name {
+        None
+    } else {
+        // Return the package override name, if it exists, or else the Cargo manifest name.
+        Some(get_buildsys_package_name(pkg_metadata))
+    }
+}
```
bcressey, line 529: Are we guaranteed that the first entry in the traversal is the node we started from? Naively, that's what I would expect from a graph iterator. If so, we could just skip the first element from the iterator.

tools/buildsys/src/manifest.rs
```diff
@@ -259,6 +281,89 @@ use std::path::{Path, PathBuf};
...
+            is_valid_dep(name, &link) && is_manifest_type(&to, BuildType::Package)
+        });
+        let mut packages: Vec<String> = package_set
+            .packages(DependencyDirection::Forward)
+            .filter_map(|pkg_metadata| filter_map_to_name(name, &pkg_metadata))
+            .collect();
+
+        // Sort so that this function has consistent, dependable output regardless of graph internals.
+        packages.sort();
+        Ok(packages)
```
bcressey, line 338: Does this make a difference? Seems like we'd already be getting topologically-sorted output, which should be reasonable consistent. This final sort ends up throwing away the ordering information.

bcressey (approved):

jmt-lab (approved):
LGTM

tools/buildsys/src/args.rs
```diff
@@ -57,6 +58,9 @@ pub(crate) struct Common {
     #[arg(long, env = "BUILDSYS_ARCH")]
     pub(crate) arch: SupportedArch,
 
+    #[arg(long, env = "BUILDSYS_CARGO_METADATA_PATH")]
```
sam-berning, line 61: does this need to be added to the rebuild vars?

sam-berning (approved):

webern:
> https://github.com/bottlerocket-os/twoliter/compare/5824391d09f6343f27c181f71d4ca98d6c30078a..654e237e8af43776e4fe07daa7b5ff8f65aab0a8

Fix a nit by moving a cargo clean in the Makefile to the clean-packages section.

tools/buildsys/src/manifest.rs
webern: > Are we guaranteed that the first entry in the traversal is the node we started from? Naively, that's what I would expect from a graph iterator. If so, we could just skip the first element from the iterator.

The documentation is silent on order of returned results. Seems safer and cheap to not make the assumption.

tools/buildsys/src/manifest.rs
webern: > Does this make a difference? Seems like we'd already be getting topologically-sorted output, which should be reasonable consistent. This final sort ends up throwing away the ordering information.

The documentation is silent on order of returned results (though topo-sort seems likely). My most common interaction with this list is "is the package I care about in the list?", whereby alphabetical is more convenient.

tools/buildsys/src/args.rs
webern: > I think we need kits to include the same build flags that packages do, except we should rebuild on a change to BUILDSYS_KITS_DIR vs BUILDSYS_PACKAGES_DIR.

True! I'll get it in the next one when adding the `build-kit` command and getting it to work.

tools/buildsys/src/args.rs
webern: > does this need to be added to the rebuild vars?

I think we're covered by the fact that the package/variant dirs are being watched for changes, and the cargo metadata output should be the same unless something has changed about those packages. In other words, I think if anything important changes, we will be rebuilding anyway.

tests/projects/local-kit/kits/extra-1-kit/Cargo.toml
webern: An `included-packages` would be sort of gnarly to implement since it would need to inform the graph algorithm. When we talked about it we were thinking we would just go with Cargo dependencies defining the package list. `excluded-packages` could be a feature add fairly easily if anyone ends up needing that.

PR #199, opened 2024-04-16
rpm2img: include variant name to binaries in an ELF note

sam-berning:
**Issue number:**

N/A

**Description of changes:**

Adds an ELF note to binaries under `/usr/bin` that contains the name of the variant. This change is needed to enable us to package each variant's model into a giant enum and serialize and deserialize the correct model by reading the variant name from the ELF note.

**Testing done:**

Built an `aws-dev` variant from the current bottlerocket repo at https://github.com/sam-berning/bottlerocket/tree/megamodel using this version of `rpm2img`. That branch contains a proof-of-concept of the combined model-of-models and of reading the variant from the binary's ELF notes to serialize and deserialize the correct model.

When I ran the resulting AMI, the correct settings were present when calling `apiclient get` and `set`

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/rpm2img
+++ b/twoliter/embedded/rpm2img
@@ -240,6 +240,68 @@ INVENTORY_DATA="$(jq --arg PKG_PREFIX "bottlerocket-" \
 INVENTORY_DATA="$(jq --slurp 'sort_by(.Name)' <<< "${INVENTORY_DATA}" | jq '{"Content": .}')"
 printf "%s\n" "${INVENTORY_DATA}" > "${ROOT_MOUNT}/usr/share/bottlerocket/application-inventory.json"
 
+elf_note() {
+  local input output name
+  local input_len input_pad input_fmt
+  local name_len name_pad name_fmt
+  local var val len pad fmt n sz
+  input="${1:?}"
+  output="${2:?}"
+  name="Bottlerocket"
+
+  # Handle null-termination, padding, and formatting.
+  for var in name input ; do
+    val="${!var}"
+    len="${var}_len"
+    pad="${var}_pad"
+    fmt="${var}_fmt"
+    # Add one byte for null terminator.
+    printf -v "${len}" "$(( "${#val}" + 1 ))"
+    # Pad to the nearest 4 byte boundary.
+    printf -v "${pad}" "$(( (4 - ("${!len}" % 4)) % 4 ))"
+    # Generate a format string for the variable.
+    str="%%s\\\x00"
+    for p in $(seq 1 "${!pad}") ; do
+      str+="\\\x00"
+    done
+    printf -v "${fmt}" "${str}"
+  done
+
+  # Clean up the output file.
+  rm -f "${output}"
+
+  # Write the sizes for the ELF note name and description.
+  for n in "${name_len}" "${input_len}" 0 ; do
+    printf -v sz '\\x%02x\\x%02x\\x%02x\\x%02x' \
+      $((n & 255)) $((n >> 8 & 255)) $((n >> 16 & 255)) $((n >> 24 & 255))
+    printf "${sz}" >> "${output}"
+  done
+
+  # Write the ELF note name and description.
+  printf "${name_fmt}" "${name}" >> "${output}"
+  printf "${input_fmt}" "${input}" >> "${output}"
+}
+
+is_elf() {
+  local some_file elf_magic
+  some_file="${1:?}"
+  elf_magic="\x7fELF"
+  [ "$(head -c 4 ${some_file})" == "$(echo -e ${elf_magic})" ]
+}
+
+# add elf note to relevant binaries
+elf_note "${VARIANT}" "note.elf"
+
+for some_file in "${ROOT_MOUNT}"/usr/bin/*; do
+  # ensure that the file is not a link to avoid adding duplicate notes
+  if ! readlink "${some_file}" >/dev/null; then
+    # check that the file is an ELF binary
+    if is_elf ${some_file}; then
+      objcopy --add-section .note.bottlerocket.variant="note.elf" "${some_file}"
+    fi
+  fi
+done
+
 # install licenses
 mksquashfs \
   "${ROOT_MOUNT}"/usr/share/licenses \


twoliter/embedded/rpm2img
```diff
@@ -240,6 +240,37 @@ INVENTORY_DATA="$(jq --arg PKG_PREFIX "bottlerocket-" \
...
+  done
+  echo -en "${name}\x00" >> "${output}"
+  echo -en "${input}\x00" >> "${output}"
+}
+
+# add elf note to relevant binaries
+elf_note "${VARIANT}" "note.elf"
+
+ELF_MAGIC="\x7fELF"
+for file in "${ROOT_MOUNT}"/usr/bin/*; do
```
jpculp:
```suggestion
for some_file in "${ROOT_MOUNT}"/usr/bin/*; do
```
Can we avoid `file` as a variable name? There is `file` utility present in the Bottlerocket SDK.

jpculp (changes requested):

jpculp (approved):

jmt-lab (approved):
LGTM

twoliter/embedded/rpm2img
```diff
@@ -240,6 +240,37 @@ INVENTORY_DATA="$(jq --arg PKG_PREFIX "bottlerocket-" \
 INVENTORY_DATA="$(jq --slurp 'sort_by(.Name)' <<< "${INVENTORY_DATA}" | jq '{"Content": .}')"
 printf "%s\n" "${INVENTORY_DATA}" > "${ROOT_MOUNT}/usr/share/bottlerocket/application-inventory.json"
 
+elf_note() {
+  local input output size offset uint32_t
+  input="${1:?}"
+  output="${2:?}"
+  name="Bottlerocket"
+  name_len="$(( ${#name} + 1 ))"
+  input_len="$(( ${#input} + 1 ))"
```
bcressey: This isn't quite right because the name has to be padded to a four-byte boundary ([source](https://www.netbsd.org/docs/kernel/elf-notes.html)) including the null terminator.

If I try this as-is then readelf complains:
```
Displaying notes found in: .note.bottlerocket.variant
  Owner                Data size 	Description
readelf: ./target/release/legolas: Warning: note with invalid namesz and/or descsz found at offset 0
readelf: ./target/release/legolas: Warning:  type: 0, namesize: 0xd, descsize: 0x14, alignment: 4
```

twoliter/embedded/rpm2img
```diff
@@ -240,6 +240,37 @@ INVENTORY_DATA="$(jq --arg PKG_PREFIX "bottlerocket-" \
...
+ELF_MAGIC="\x7fELF"
+for some_file in "${ROOT_MOUNT}"/usr/bin/*; do
+  # ensure that the file is not a link to avoid adding duplicate notes
+  if ! readlink "${some_file}" >/dev/null; then
+    # check that the file is an ELF binary
+    if [[ "$(head -c 4 ${some_file})" == "$(echo -e ${ELF_MAGIC})" ]]; then
+      objcopy --add-section .note.bottlerocket.variant="note.elf" "${some_file}"
+    fi
+  fi
+done
```
bcressey: nit: consider factoring the ELF test into an `is_elf()` helper function.

How long does it take to add the note to all binaries? We might need to be smarter about which binaries we touch if it's longer than a couple seconds.

twoliter/embedded/rpm2img
bcressey: A bit tricky because we need to write out runs of null bytes to achieve the padding, which is hard in bash.

I ended up with this:
```
elf_note() {
  local input output name
  local input_len input_pad input_fmt
  local name_len name_pad name_fmt
  local var val len pad fmt n sz
  input="${1:?}"
  output="${2:?}"
  name="Bottlerocket"

  # Handle null-termination, padding, and formatting.
  for var in name input ; do
    val="${!var}"
    len="${var}_len"
    pad="${var}_pad"
    fmt="${var}_fmt"
    # Add one byte for null terminator.
    printf -v "${len}" "$(( "${#val}" + 1 ))"
    # Pad to the nearest 4 byte boundary.
    printf -v "${pad}" "$(( (4 - ("${!len}" % 4)) % 4 ))"
    # Generate a format string for the variable.
    str="%%s\\\x00"
    for p in $(seq 1 "${!pad}") ; do
      str+="\\\x00"
    done
    printf -v "${fmt}" "${str}"
  done

  # Clean up the output file.
  rm -f "${output}"

  # Write the sizes for the ELF note name and description.
  for n in "${name_len}" "${input_len}" 0 ; do
    printf -v sz '\\x%02x\\x%02x\\x%02x\\x%02x' \
      $((n & 255)) $((n >> 8 & 255)) $((n >> 16 & 255)) $((n >> 24 & 255))
    printf "${sz}" >> "${output}"
  done

  # Write the ELF note name and description.
  printf "${name_fmt}" "${name}" >> "${output}"
  printf "${input_fmt}" "${input}" >> "${output}"
}
```

twoliter/embedded/rpm2img
sam-berning: just tested and it came out to 3 seconds on `aws-dev`, I think that should be ok

twoliter/embedded/rpm2img
sam-berning: good catch! just published a new version that uses this script with one modification: after we calculate the padding size I added that to the overall size of the section. without this, the parser can't properly find the start of the `desc` section

bcressey: > good catch! just published a new version that uses this script with one modification: after we calculate the padding size I added that to the overall size of the section. without this, the parser can't properly find the start of the desc section

The linked reference says:
> The "Name Size" and "Desc Size" fields are integers (in the byte order specified by the binary's ELF header) which specify the size of the "Name" and "Desc" fields (excluding padding).

So I don't think it's correct to add the padding to these size fields.

bcressey: I expect this is the right fix on the parsing side:
```
let mut shdr: SectionHeader = ... ;
shdr.sh_addralign = 4;
```

It looks like the address alignment defaults to one byte, which doesn't seem correct for note sections ([source](https://docs.oracle.com/cd/E23824_01/html/819-0690/chapter6-18048.html)):
> For 64–bit objects and 32–bit objects, each entry is an array of 4-byte words in the format of the target processor.
> ...
> Padding is present, if necessary, to ensure 4-byte alignment for the descriptor. Such padding is not included in namesz.

sam-berning: ah ok, noted. removed my change and will fix the parsing logic

twoliter/embedded/rpm2img
```diff
@@ -240,6 +240,68 @@ INVENTORY_DATA="$(jq --arg PKG_PREFIX "bottlerocket-" \
...
+  # Write the ELF note name and description.
+  printf "${name_fmt}" "${name}" >> "${output}"
+  printf "${input_fmt}" "${input}" >> "${output}"
+}
+
+is_elf() {
+  local some_file elf_magic
+  some_file="${1:?}"
+  elf_magic="\x7fELF"
+  [[ "$(head -c 4 ${some_file})" == "$(echo -e ${elf_magic})" ]]
```
bcressey: nit: don't need `[[` for this, `[` would work fine

bcressey (approved):

webern (approved):
Is doubling-down on "variant name" for this model-selection plan the right approach? Is it compatible with the other idea of a plugin dylib for selecting a different model? My thought was that the name of the model could be used instead of variant-name logic.

bcressey:
> Is it compatible with the other idea of a plugin dylib for selecting a different model?

If we go with the settings plugin approach then I don't think we need this. We'd just need a default `dev-dependencies` plugin crate for use with unit tests, and then the final image would have only one plugin installed.

sam-berning:
Canceling this PR as we've discussed that we prefer dynamically-loaded settings plugins to the [megamodel](https://github.com/sam-berning/bottlerocket/tree/megamodel) approach to remove conditional compilation as it also allows builders to specify settings out of tree.

PR #203, opened 2024-04-26
makefile: add fetch-variant

### LINKED ISSUE #176: Tooling to import images from the TUF repo
**What I'd like:**
bottlerocket-os/bottlerocket#2486 covers the modify-image flow where modifying a pre-built image is faster or more advantageous than building from source. There are really two distinct pieces here, the modification of images vs downloading built images from the TUF repo and placing them in the working directory for modification. This issue covers importing those images. 

Ideally this would be a make target such as `cargo make import-images` that takes a variant and version and downloads from a configured TUF repo. There is likely work to also have the `kmod-kit` and migrations also downloaded but I think we could scope that to yet a different issue if needed since modifying the images is really the only thing strictly needed.

[I wrote a bash version](https://github.com/bottlerocket-os/bottlerocket/pull/3136/files#diff-9375fd04332c86472d7be397ef09428cb86babd8826880a5835bd1d1c1bdbc08R1050) of this target but ideally, this should go in `pubsys` to keep all the logic around where these images should be in one place.

Building from source, which sort of defeats the point of bottlerocket-os/bottlerocket#2486

jpculp:
**Issue number:**

Closes #176

**Description of changes:**

Adds a target to pull a pre-built variant using `tuftool` to the typical build directory. Supports both 'unified' and 'split' partitioned variants.

**Testing done:**

- [x] Downloaded artifacts with default `root.json`.
- [x] Downloaded artifacts with local `root.json`.
- [x] Attempted to download artifacts with a non-empty output directory.
- [x] Validated `tuftool` command when `FETCH_VARIANT_UNIFIED=true`.

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/Makefile.toml
+++ b/twoliter/embedded/Makefile.toml
@@ -186,6 +186,15 @@ BUILDSYS_OVF_TEMPLATE = "${BUILDSYS_ROOT_DIR}/variants/${BUILDSYS_VARIANT}/templ
 # The default name of uploaded OVAs; override by setting VMWARE_VM_NAME
 VMWARE_VM_NAME_DEFAULT = "${BUILDSYS_NAME}-${BUILDSYS_VARIANT}-${BUILDSYS_ARCH}-v${BUILDSYS_VERSION_IMAGE}-${BUILDSYS_VERSION_BUILD}"
 
+# Depends on ${BUILDSYS_ARCH} and ${BUILDSYS_VARIANT}.
+FETCH_VARIANT_LOCAL_ROOT_JSON = { script = ['echo "${FETCH_VARIANT_LOCAL_ROOT_JSON:-""}"'] }
+FETCH_VARIANT_ROOT_JSON = { script = ['echo "${FETCH_VARIANT_ROOT_JSON:-https://cache.bottlerocket.aws/root.json}"'] }
+FETCH_VARIANT_ROOT_JSON_SHA512 = { script = ['echo "${FETCH_VARIANT_ROOT_JSON_SHA512:-a3c58bc73999264f6f28f3ed9bfcb325a5be943a782852c7d53e803881968e0a4698bd54c2f125493f4669610a9da83a1787eb58a8303b2ee488fa2a3f7d802f}"'] }
+
+FETCH_VARIANT_UNIFIED = { script = ['echo "${FETCH_VARIANT_UNIFIED:-false}"'] }
+FETCH_VARIANT_METADATA_URL = { script = ['echo "${FETCH_VARIANT_METADATA_URL:-https://updates.bottlerocket.aws/2020-07-07/"${BUILDSYS_VARIANT}"/"${BUILDSYS_ARCH}"/}"'] }
+FETCH_VARIANT_TARGETS_URL = { script = ['echo "${FETCH_VARIANT_TARGETS_URL:-https://updates.bottlerocket.aws/targets}"'] }
+
 # Config file for Boot Configuration initrd generation
 BOOT_CONFIG_INPUT = "${BUILDSYS_ROOT_DIR}/bootconfig-input"
 # Boot Configuration initrd
@@ -1412,6 +1421,42 @@ pubsys \
 '''
 ]
 
+[tasks.fetch-variant]
+script_runner = "bash"
+script = [
+'''
+set -e
+
+export PATH="${TWOLITER_TOOLS_DIR}:${PATH}"
+
+if [ -z "${FETCH_VARIANT_LOCAL_ROOT_JSON}" ]; then
+   fetch_variant_root="./fetch-variant-root.json"
+   echo "Fetching root.json..."
+   curl -s -L -C - "${FETCH_VARIANT_ROOT_JSON}" -o "${fetch_variant_root#./}"
+   sha512sum -c <<<"${FETCH_VARIANT_ROOT_JSON_SHA512}  ${fetch_variant_root#./}"
+else
+   fetch_variant_root="${FETCH_VARIANT_LOCAL_ROOT_JSON}"
+fi
+
+if [ "${FETCH_VARIANT_UNIFIED}" == "true" ]; then
+   tuftool download \
+      "${BUILDSYS_VARIANT_DIR}" \
+      --target-name "${BUILDSYS_NAME_FRIENDLY}.img.lz4" \
+      --root "${fetch_variant_root:?}" \
+      --metadata-url "${FETCH_VARIANT_METADATA_URL}" \
+      --targets-url "${FETCH_VARIANT_TARGETS_URL}"
+else
+   tuftool download \
+      "${BUILDSYS_VARIANT_DIR}" \
+      --target-name "${BUILDSYS_NAME_FRIENDLY}.img.lz4" \
+      --target-name "${BUILDSYS_NAME_FRIENDLY}-data.img.lz4" \
+      --root "${fetch_variant_root:?}" \
+      --metadata-url "${FETCH_VARIANT_METADATA_URL}" \
+      --targets-url "${FETCH_VARIANT_TARGETS_URL}"
+fi
+'''
+]
+
 [tasks._upload-ova-base]
 # Rather than depend on "build", which currently rebuilds images each run, we
 # depend on publish-tools and check for the image files below to save time.
@@ -1643,7 +1688,7 @@ script = [
 
 # This task will call watch on the `status` testsys command to show the results of all tests and
 # resources.
-# To see all incomplete crds use `cargo make watch-test-all --running` 
+# To see all incomplete crds use `cargo make watch-test-all --running`
 [tasks.watch-test-all]
 script = [
    '''


twoliter/embedded/Makefile.toml
```diff
@@ -186,6 +186,15 @@ BUILDSYS_OVF_TEMPLATE = "${BUILDSYS_ROOT_DIR}/variants/${BUILDSYS_VARIANT}/templ
 # The default name of uploaded OVAs; override by setting VMWARE_VM_NAME
 VMWARE_VM_NAME_DEFAULT = "${BUILDSYS_NAME}-${BUILDSYS_VARIANT}-${BUILDSYS_ARCH}-v${BUILDSYS_VERSION_IMAGE}-${BUILDSYS_VERSION_BUILD}"
 
+# Depends on ${BUILDSYS_ARCH} and ${BUILDSYS_VARIANT}.
+FETCH_VARIANT_LOCAL_ROOT_JSON = { script = ['echo "${FETCH_VARIANT_LOCAL_ROOT_JSON:-""}"'] }
+FETCH_VARIANT_ROOT_JSON = { script = ['echo "${FETCH_VARIANT_ROOT_JSON:-https://cache.bottlerocket.aws/root.json}"'] }
+FETCH_VARIANT_ROOT_JSON_SHA512 = { script = ['echo "${FETCH_VARIANT_ROOT_JSON_SHA512:-a3c58bc73999264f6f28f3ed9bfcb325a5be943a782852c7d53e803881968e0a4698bd54c2f125493f4669610a9da83a1787eb58a8303b2ee488fa2a3f7d802f}"'] }
```
rpkelly, line 192: This seems like an easy place to forget to update when we update the root.json

rpkelly (approved):

twoliter/embedded/Makefile.toml
```diff
@@ -1412,6 +1421,42 @@ pubsys \
...
+   curl -s -L -C - "${FETCH_VARIANT_ROOT_JSON}" -o "${fetch_variant_root#./}"
+   sha512sum -c <<<"${FETCH_VARIANT_ROOT_JSON_SHA512}  ${fetch_variant_root#./}"
+else
+   fetch_variant_root="${FETCH_VARIANT_LOCAL_ROOT_JSON}"
+fi
+
+if [ "${FETCH_VARIANT_UNIFIED}" == "true" ]; then
+   tuftool download \
+      "${BUILDSYS_VARIANT_DIR}" \
+      --target-name "${BUILDSYS_NAME_FRIENDLY}.img.lz4" \
```
yeazelm, line 1444: This doesn't work for VMware variants which have different suffixes.

jpculp: I planned on adding some VMware-specific logic in a later PR.

yeazelm: Makes sense it's just calling out that this doesn't work for those variants.

yeazelm (approved):

twoliter/embedded/Makefile.toml
webern: We could also specify a specific root.json, like `root-4.json`, which would keep matching the sha and would allow us to pull future root.jsons in the chain.

I agree this is a foot-gun as is.

jpculp: I think we just have to be diligent in updating. Even if there was a lapse between updates, the work around would be to pass `-e  FETCH_VARIANT_ROOT_JSON_SHA512=<NEW_SHA>`. An alternative is we just have the hash somewhere in the bottlerocket repo and then this logic and any quickstart guides we have can just point to that hashes file. We can even keep track of older roots in said file.

twoliter/embedded/Makefile.toml
```diff
@@ -186,6 +186,15 @@ BUILDSYS_OVF_TEMPLATE = "${BUILDSYS_ROOT_DIR}/variants/${BUILDSYS_VARIANT}/templ
...
 VMWARE_VM_NAME_DEFAULT = "${BUILDSYS_NAME}-${BUILDSYS_VARIANT}-${BUILDSYS_ARCH}-v${BUILDSYS_VERSION_IMAGE}-${BUILDSYS_VERSION_BUILD}"
 
+# Depends on ${BUILDSYS_ARCH} and ${BUILDSYS_VARIANT}.
+FETCH_VARIANT_LOCAL_ROOT_JSON = { script = ['echo "${FETCH_VARIANT_LOCAL_ROOT_JSON:-""}"'] }
+FETCH_VARIANT_ROOT_JSON = { script = ['echo "${FETCH_VARIANT_ROOT_JSON:-https://cache.bottlerocket.aws/root.json}"'] }
+FETCH_VARIANT_ROOT_JSON_SHA512 = { script = ['echo "${FETCH_VARIANT_ROOT_JSON_SHA512:-a3c58bc73999264f6f28f3ed9bfcb325a5be943a782852c7d53e803881968e0a4698bd54c2f125493f4669610a9da83a1787eb58a8303b2ee488fa2a3f7d802f}"'] }
+
+FETCH_VARIANT_UNIFIED = { script = ['echo "${FETCH_VARIANT_UNIFIED:-false}"'] }
+FETCH_VARIANT_METADATA_URL = { script = ['echo "${FETCH_VARIANT_METADATA_URL:-https://updates.bottlerocket.aws/2020-07-07/"${BUILDSYS_VARIANT}"/"${BUILDSYS_ARCH}"/}"'] }
+FETCH_VARIANT_TARGETS_URL = { script = ['echo "${FETCH_VARIANT_TARGETS_URL:-https://updates.bottlerocket.aws/targets}"'] }
```
bcressey, line 196: No. This needs to integrate with `Infra.toml` and take a named repo. We have `PUBLISH_REPO` for the target of a new repo build, so you can have `FETCH_REPO` for the source of an import.

twoliter/embedded/Makefile.toml
```diff
@@ -186,6 +186,15 @@ BUILDSYS_OVF_TEMPLATE = "${BUILDSYS_ROOT_DIR}/variants/${BUILDSYS_VARIANT}/templ
 # The default name of uploaded OVAs; override by setting VMWARE_VM_NAME
 VMWARE_VM_NAME_DEFAULT = "${BUILDSYS_NAME}-${BUILDSYS_VARIANT}-${BUILDSYS_ARCH}-v${BUILDSYS_VERSION_IMAGE}-${BUILDSYS_VERSION_BUILD}"
 
+# Depends on ${BUILDSYS_ARCH} and ${BUILDSYS_VARIANT}.
+FETCH_VARIANT_LOCAL_ROOT_JSON = { script = ['echo "${FETCH_VARIANT_LOCAL_ROOT_JSON:-""}"'] }
+FETCH_VARIANT_ROOT_JSON = { script = ['echo "${FETCH_VARIANT_ROOT_JSON:-https://cache.bottlerocket.aws/root.json}"'] }
+FETCH_VARIANT_ROOT_JSON_SHA512 = { script = ['echo "${FETCH_VARIANT_ROOT_JSON_SHA512:-a3c58bc73999264f6f28f3ed9bfcb325a5be943a782852c7d53e803881968e0a4698bd54c2f125493f4669610a9da83a1787eb58a8303b2ee488fa2a3f7d802f}"'] }
+
+FETCH_VARIANT_UNIFIED = { script = ['echo "${FETCH_VARIANT_UNIFIED:-false}"'] }
```
bcressey, line 194: You should parse this from the local variant definition. If we don't have the variant definition, I am very skeptical that the larger operation (fetch, resign, publish) will be correct.

twoliter/embedded/Makefile.toml
```diff
@@ -1412,6 +1421,42 @@ pubsys \
...
+export PATH="${TWOLITER_TOOLS_DIR}:${PATH}"
+
+if [ -z "${FETCH_VARIANT_LOCAL_ROOT_JSON}" ]; then
+   fetch_variant_root="./fetch-variant-root.json"
+   echo "Fetching root.json..."
+   curl -s -L -C - "${FETCH_VARIANT_ROOT_JSON}" -o "${fetch_variant_root#./}"
+   sha512sum -c <<<"${FETCH_VARIANT_ROOT_JSON_SHA512}  ${fetch_variant_root#./}"
+else
+   fetch_variant_root="${FETCH_VARIANT_LOCAL_ROOT_JSON}"
+fi
```
bcressey, line 1439: These values need to come from `Infra.toml`.

If you have `FETCH_REPO` you should be able to use its `root_role_url` and `root_role_sha512` fields for this. `pubsys-setup` could probably be extended for this.

twoliter/embedded/Makefile.toml
```diff
@@ -1412,6 +1421,42 @@ pubsys \
...
+      --targets-url "${FETCH_VARIANT_TARGETS_URL}"
+else
+   tuftool download \
+      "${BUILDSYS_VARIANT_DIR}" \
+      --target-name "${BUILDSYS_NAME_FRIENDLY}.img.lz4" \
+      --target-name "${BUILDSYS_NAME_FRIENDLY}-data.img.lz4" \
+      --root "${fetch_variant_root:?}" \
+      --metadata-url "${FETCH_VARIANT_METADATA_URL}" \
+      --targets-url "${FETCH_VARIANT_TARGETS_URL}"
+fi
```
bcressey, line 1456: Why do you need the `-data.img` file? I'd expect you  could do all the operations on just the boot / root / hash part of the main `.img` file, whether or not it contains the data partition.

As a practical matter, you end up depending on the variant's manifest in order to correctly do Secure Boot registration, and (for VMware) to pick the correct OVF template.

bcressey (changes requested):

twoliter/embedded/Makefile.toml
bcressey: > Why do you need the `-data.img` file?

Answering my own question - `-data.img` would be needed to carry out a publish operation after the resigning, so it's better to pull the artifact down at the same time (if needed).

But again, unified vs. split is in the variant's manifest, so you can just read it from there instead.

jpculp:
Superseded by https://github.com/bottlerocket-os/twoliter/pull/236.

PR #204, opened 2024-04-30
design: build algorithm for kits

webern:
**Issue number:**

Related to #198 

**Description of changes:**

Describe a build algorithm that understands the difference between kits and packages and can put the right packages into a kit.

**Testing done:**

N/A

**Terms of contribution:**

### DIFF
--- a/docs/design/README.md
+++ b/docs/design/README.md
@@ -142,27 +142,27 @@ This will be a Cargo workspace with one Cargo package for each kit.
 
 ```
 kits
-├── aws
+├── aws-kit
 │   ├── build.rs
 │   ├── Cargo.toml
 │   └── lib.rs
-├── core
+├── core-kit
 │   ├── build.rs
 │   ├── Cargo.toml
 │   └── lib.rs
-├── ecs
+├── ecs-kit
 │   ├── build.rs
 │   ├── Cargo.toml
 │   └── lib.rs
-├── k8s
+├── k8s-kit
 │   ├── build.rs
 │   ├── Cargo.toml
 │   └── lib.rs
-├── metal
+├── metal-kit
 │   ├── build.rs
 │   ├── Cargo.toml
 │   └── lib.rs
-└── vmware
+└── vmware-kit
     ├── build.rs
     ├── Cargo.toml
     └── lib.rs
@@ -441,7 +441,7 @@ The Cargo package versions will be locked using Cargo's `=` operator for depende
 bottlerocket-sdk = "0.50.0"
 ```
 
-At this point Twoliter will build a package that represents the maintainers top level dependencies.
+At this point, Twoliter will build a package that represents the maintainer's top level dependencies.
 
 ```toml
 [dependencies]
@@ -471,7 +471,7 @@ This means we will have the following items and dependencies:
 The user kicks things off with:
 
 ```sh
-twoliter build --variant example-dev --arch aarch64
+twoliter build variant example-dev --arch aarch64
 ```
 
 ![build sequence diagram](diagrams/build-sequence.svg "Build Sequence")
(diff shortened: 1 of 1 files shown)


docs/design/README.md
```diff
@@ -534,6 +534,149 @@ After installing packages, the result of `dnf list installed` will be output to
...
+
+When creating a kit, we won't have the yum install command to help us figure out all the right packages.
+We need to change the build system so that the dependency graph can be used to put the right packages into a kit.
+
+For starters, we will no longer rely on a flat structure in `build/rpms`.
+We will put the output of each `package` build into a subdirectory using the Cargo package's name.
+
+For example, when building `glibc/Cargo.toml` we will put the output in `build/rpms/glibc`.
+Note that more than one RPM file can result from a package build.
+The subdirectory maps the resultant RPM files to the Cargo package that produced them.
```
jmt-lab, line 551: For knowledge sharing why is this needed to have a directory per package instead of say a directory per kit?

docs/design/README.md
```diff
@@ -534,6 +534,149 @@ After installing packages, the result of `dnf list installed` will be output to
...
+
+At each `build-package` invocation, a distinct, flat RPM directory will be created containing the packages needed for the current `build-package` invocation.
+This will be done by traversing the Cargo dependency graph and making a list of packages and kits that are needed.
+The kits will be unpacked and the direct package dependencies will be copied from their `build/rpms` subdirectories.
+This will require us to pass a list of the packages needed for the build to the dockerfile, for example:
```
jmt-lab, line 556: Wouldn't this cause potentially a lot of pack-unpack-pack-unpack cycle, if we have the directory per kit and the directory is already built in our build/rpms could we not sufficiently utilize the pre-extracted set of rpms?

docs/design/README.md
```diff
@@ -534,6 +534,149 @@ After installing packages, the result of `dnf list installed` will be output to
...
+The `build-kit` requirements are slightly simplified.
+Only the list of "pure" package dependencies is required.
+When constructing a kit, buildsys does not need to know what other kits the current kit depends on.
+
+#### Dependency Traversal
+
+The simplest way to get a list of "pure" package dependencies and a list of kits is to traverse the dependency graph twice.
+
+In the first traversal, kit dependencies are ignored and not followed, resulting in only the "pure" package dependencies.
+In the second traversal, all nodes are followed and package dependencies are filtered out from the final list, resulting in a list of all required kits.
```
jmt-lab, line 584: I am still confused on why the two traversals are needed, can we clarify this need and why we can't just ensure we have the kit metadata as pregenerated. (As in when the kit is built it also generates its metadata.json which would list any dependent kits. Then gathering the total set of needed kits wouldn't be a scan each package problem, potentially.

docs/design/README.md
```diff
@@ -534,6 +534,149 @@ After installing packages, the result of `dnf list installed` will be output to
...
+
+```
+ARG PACKAGES
+ARG KITS
+
+mkdir big-rpm-dir
+
+# PACKAGES could be empty for packages like glibc with no deps
+for pkg in ${PACKAGES} ; do
+  cp /host/build/rpms/${pkg}/* big-rpm-dir
```
jmt-lab, line 611: This syncs up with above question on why the pkg dirs are needed especially if the end result is just copying them to the kit anyway. Seems like creating duplicate storage in build folder for an unknown reason right now

docs/design/README.md
```diff
@@ -534,6 +534,149 @@ After installing packages, the result of `dnf list installed` will be output to
...
+buildsys:
+
+```
+// load the kit's manifest
+let (packages, _) = manifest.walk_dependency_graph();
+
+// no need for kits, they're some other crate's problem
+// when calling build_kit:
+  build-arg PACKAGES="{packages}"
+```
```
jmt-lab, line 657: Just a note talking with Ben the other day we don't necessarily think that buildsys should care about packing the repository into a container. This is still accurate here right, as in the dockerfile is more for orchestrating the kit build for creating the "local-kit"?

docs/design/README.md
bcressey: > For knowledge sharing why is this needed to have a directory per package instead of say a directory per kit?

Kits may have "internal" packages that are needed to build other packages in that kit, but not intended to propagate onward when the kit is published. An example would be a package to build a host tool that is then used to build other packages for the target. To support this it's necessary to have precise targeting for which RPMs belong to which package, so that the resulting kit doesn't contain more than intended.

However, the more fundamental reason is that this is the wrong direction to think about the problem from. Packages don't "know" that they are part of a kit, any more than they should "know" they are part of a variant. `cargo` will trigger package crate builds if the kit crate depends on them, but the package doesn't know why it is being built unless we build some mechanism to pass that data in the environment. That's similar to how we pass variant data to packages today, and we're trying hard to eliminate that mechanism.

docs/design/README.md
bcressey: We could leave kits in their unpacked form with loose RPMs + a repo definition, either instead of an archive or in addition to it.

It's sort of a trade-off between how complex we want the publish side to be vs. how complex we want the build side to be, since what's convenient for one is not for the other. But we could also just create both forms in the `build-kit` operation.

docs/design/README.md
bcressey: > I am still confused on why the two traversals are needed, can we clarify this need

The algorithm design could clarify that this is "conceptually" the simplest way. Obviously you can optimize it for a single traversal, but that would require some gnarlier state tracking since you would have to switch modes when crossing a kit boundary and put those packages in a different collection.

PR #205, opened 2024-05-01
rpm2img: regenerate kernel module info if possible

bcressey:
**Issue number:**
Related: https://github.com/bottlerocket-os/bottlerocket/pull/3932

**Description of changes:**
With conditional compilation being eliminated, it is no longer always the case that all of the modules built by the kernel package will be installed on all variants. For example, some network device drivers may be built as modules and only installed on bare metal variants, if the corresponding hardware is not expected to show up in virtualized environments.

This requires some care in packaging so that any module dependencies that are not provided by the kernel will be provided by other modules that are also installed. By regenerating module info at image build time, it's guaranteed to be present and correct, or otherwise the build will fail.

**Testing done:**
While testing the related pull request, I verified that modules in the new `modules-metal` package were not present in the regenerated modinfo.

I also verified that removing a dependency of a module resulted in a build failure. (`depmod` does not have an option to exit non-zero after emitting a warning, which is why `rpm2img` captures the output and inspects it with `grep`.)

**Terms of contribution:**


yeazelm (approved):

jmt-lab (approved):
LGTM

PR #206, opened 2024-05-02
add repack-variant (OLD)

jpculp:
**Description of changes:**

This PR is a work in progress to introduce functionality to repackage a variant.

```
    buildsys: add repack-variant target

    Add the the initial target for repackaging a Bottlerocket variant. This
    target will skip the traditional workflow of building packages and
    using `rpm2img`. Instead the build system will point to a yet to be
    implemented `img2img` tool.
```

```
    buildsys: preliminary housekeeping

    Adjusts some things in the build system for preliminary support for an
    additional build target. Removes some unnecessary checks, references,
    and dependencies. Renames Dockerfile to build.Dockerfile and adds a
    common argument for future build targets to skip cleanup if desired.
```

```
    buildsys: add support for custom CA certificates

    Rather than having the CA certificate bundle installed as a package, the
    build system will pull the bundle from the Bottlerocket SDK. This can be
    overridden with a local CA bundle by passing
    BUILDSYS_CACERTS_BUNDLE_OVERRIDE.
```

**Testing done:**
- [x] Build Bottlerocket with default CA certificates in Bottlerocket SDK.
- [x] ~~Build Bottlerocket with custom CA certificates in `/certs`.~~
- [x] Build Bottlerocket with custom CA certificates using **BUILDSYS_CACERTS_BUNDLE_OVERRIDE**.
- [x] Build `aws-k8s-1.27` _(no secure boot)_.
- [x] Build `aws-k8s-1.28` _(secure boot)_.
- [x] Repack `aws-k8s-1.27` _(no secure boot)_.
- [x] Repack `aws-k8s-1.28` _(secure boot)_.
- [x] Repack `vmware-k8s-1.28`.

**Terms of contribution:**


tools/buildsys/src/args.rs
```diff
@@ -156,6 +158,37 @@ pub(crate) struct BuildVariantArgs {
...
+    pub(crate) variant_family: String,
+
+    #[arg(long, env = "BUILDSYS_VARIANT_FLAVOR")]
+    pub(crate) variant_flavor: String,
+
+    #[arg(long, env = "BUILDSYS_VERSION_BUILD")]
+    pub(crate) version_build: String,
+
+    #[arg(long, env = "BUILDSYS_VERSION_IMAGE")]
+    pub(crate) version_image: String,
```
bcressey: If you don't actually need these then don't collect them as arguments.

tools/buildsys/src/args.rs
```diff
@@ -165,17 +198,23 @@ fn sensitive_env_vars(build_type: BuildType) -> impl Iterator<Item = &'static st
...
+            }
+        }
     }
 }
 
 /// The thing that buildsys is building.
 #[repr(u8)]
 #[derive(Debug, Clone, Copy)]
 pub(crate) enum BuildType {
+    Repack = 0b00000000,
```
bcressey, line 205: Should extend the unit test for this - no flags should be set since nothing should trigger a rebuild.

tools/buildsys/src/args.rs
```diff
@@ -165,17 +198,23 @@ fn sensitive_env_vars(build_type: BuildType) -> impl Iterator<Item = &'static st
...
 pub(crate) fn rerun_for_envs(build_type: BuildType) {
-    for var in sensitive_env_vars(build_type) {
-        println!("cargo:rerun-if-env-changed={}", var)
+    match build_type {
+        BuildType::Repack => (),
+        _ => {
+            for var in sensitive_env_vars(build_type) {
+                println!("cargo:rerun-if-env-changed={}", var)
+            }
+        }
```
bcressey, line 197: This seems unnecessary since your `BuildType::Repack` doesn't have any flags set.

tools/buildsys/src/builder.rs
```diff
@@ -202,17 +202,57 @@ impl VariantBuildArgs {
...
+    name: String,
+    partition_plan: String,
+    variant: String,
+    variant_family: String,
+    variant_flavor: String,
+    variant_platform: String,
+    variant_runtime: String,
+    version_build: String,
+    version_image: String,
+}
```
bcressey: Please cut this down to just the values you need.

tools/buildsys/src/builder.rs
```diff
@@ -464,8 +557,12 @@ fn docker(args: &[String], retry: Retry) -> Result<Output> {
             .run()
             .context(error::CommandStartSnafu)?;
 
+        // Print `docker` output to stdout for non-cleanup tasks.
         let stdout = String::from_utf8_lossy(&output.stdout);
-        println!("{}", &stdout);
+        if !args[0].starts_with("rm") {
+            println!("{}", &stdout);
+        }
```
bcressey: This is probably better as a list of strings we want to filter from the output - checking for the prefix of the command we're running is a bit of a hack, and what if it returns some other error?

jpculp:
- Move `symlink_image` to `imghelper`.
- Prevent `buildsys` from deleting repacked images if repacking a second time.

tools/buildsys/src/builder.rs
```diff
@@ -367,8 +459,10 @@ impl DockerBuild {
             &self.state_dir,
         )?;
 
-        // Clean up any previous outputs we have tracked.
-        clean_build_files(&marker_dir, &self.artifacts_dir)?;
+        // Clean up any previous outputs we have tracked, unless we are repacking.
+        if self.target != "repack" {
+            clean_build_files(&marker_dir, &self.artifacts_dir)?;
+        }
```
jpculp: I'm not sure if I should skip this when repacking, or if it would be more correct to do something with the markers instead. The conditional was added because I noticed if you run `repack-variant` a second time the images disappear prior to `img2img` and the repackaging fails.

bcressey (commented):
Partial review, I'll try to look at the rest soon.

tools/buildsys/src/builder.rs
jpculp: I could regex the output if we want to suppress the specific error I'm trying to avoid, but I thought we might not actually care about the output from `rm` or `rmi`.

```
Error response from daemon: No such container: buildsys-repack-aws-k8s-1.27-aarch64-b4041b87905a

Error response from daemon: No such image: buildsys-repack-aws-k8s-1.27-aarch64-b4041b87905a:latest
```

tools/buildsys/src/args.rs
jpculp: I think I might need a couple, but my plan was to trim the list from this commit once I have everything working as expected in img2img.

tools/buildsys/src/args.rs
jpculp: That's what I would have thought, but without it I got a wall of output.

```
cargo:rerun-if-env-changed=BUILDSYS_ARCH
cargo:rerun-if-env-changed=BUILDSYS_NAME
cargo:rerun-if-env-changed=BUILDSYS_OUTPUT_DIR
cargo:rerun-if-env-changed=BUILDSYS_PACKAGES_DIR
cargo:rerun-if-env-changed=BUILDSYS_PRETTY_NAME
cargo:rerun-if-env-changed=BUILDSYS_ROOT_DIR
cargo:rerun-if-env-changed=BUILDSYS_STATE_DIR
cargo:rerun-if-env-changed=BUILDSYS_TIMESTAMP
cargo:rerun-if-env-changed=BUILDSYS_VARIANT
cargo:rerun-if-env-changed=BUILDSYS_VERSION_BUILD
cargo:rerun-if-env-changed=BUILDSYS_VERSION_IMAGE
cargo:rerun-if-env-changed=TLPRIVATE_SDK_IMAGE
```

tools/buildsys/src/builder.rs
bcressey: Filtering out the messages that start with this is what I meant:
```
Error response from daemon: No such container:
Error response from daemon: No such image:
```

Can be done like how `DOCKER_BUILD_FRONTEND_ERROR` is handled

tools/buildsys/src/args.rs
bcressey: > That's what I would have thought, but without it I got a wall of output.

That's unexpected - it'd be worth root-causing that rather than sidestepping the machinery here, which feels like a bit of a hack.

tools/buildsys/src/builder.rs
bcressey: It might be more aligned with the new structure to have a `CommonBuildArgs` with a `clean_output_before_build` flag, which could then be set to false. Otherwise, better to use match on `TargetBuildArgs` to control this, vs string matching on targets which could break if the `Dockerfile` is refactored.

Can you verify that the right marker files are created under `build/state` for the files generated by your build? We want the property that a resign operation followed by a clean variant build with remove the files created during resign, which is what the markers accomplish.

tools/buildsys/src/main.rs
```diff
@@ -257,6 +258,21 @@ fn build_variant(args: BuildVariantArgs) -> Result<()> {
     Ok(())
 }
 
+fn repack_variant(args: RepackVariantArgs) -> Result<()> {
+    let manifest_file = "Cargo.toml";
+
+    let manifest = ManifestInfo::new(args.common.cargo_manifest_dir.join(manifest_file))
+        .context(error::ManifestParseSnafu)?;
+
+    supported_arch(&manifest, args.common.arch)?;
```
bcressey, line 261: For consistency, this should also include the `if manifest.included_packages().is_some()` check from `build_variant` (because otherwise how could an image have been built?)

Although that check is kind of annoying so you could instead remove it in a preliminary refactoring commit.

tools/buildsys/src/builder.rs
```diff
@@ -511,6 +610,15 @@ fn secrets_args() -> Result<Vec<String>> {
...
     }
 
+    let ca_bundle_var = "BUILDSYS_CACERTS_BUNDLE";
+    let ca_bundle_path = PathBuf::from(
+        env::var(ca_bundle_var).context(error::EnvironmentSnafu { var: ca_bundle_var })?,
+    );
+
+    if ca_bundle_path.exists() {
+        args.build_secret("file", "ca-bundle.crt", &ca_bundle_path.to_string_lossy());
+    }
```
bcressey: I don't agree with these checks - if the environment variable exists and points to a non-existent path, that should trigger a build failure. Otherwise it's too easy to silently omit a vital piece of configuration and build the wrong thing.

What you want instead is the inverse - if the environment variable is unset or empty, skip the secret; otherwise, assert that the file exists and then pass it as a secret.

twoliter/embedded/Makefile.toml
```diff
@@ -16,6 +16,8 @@ BUILDSYS_TOOLS_DIR = "${BUILDSYS_ROOT_DIR}/tools"
 BUILDSYS_SOURCES_DIR = "${BUILDSYS_ROOT_DIR}/sources"
 BUILDSYS_SBKEYS_DIR = "${BUILDSYS_ROOT_DIR}/sbkeys"
 BUILDSYS_SBKEYS_PROFILE = { script = ['echo "${BUILDSYS_SBKEYS_PROFILE:-local}"'] }
+BUILDSYS_CACERTS_DIR = "${BUILDSYS_ROOT_DIR}/certs"
+BUILDSYS_CACERTS_BUNDLE = { script = ['echo "${BUILDSYS_CACERTS_BUNDLE:-${BUILDSYS_CACERTS_DIR}/ca-bundle.crt}"'] }
```
bcressey: Per my feedback in `buildsys` - I don't think this should be unconditionally set to a real path.

twoliter/embedded/Makefile.toml
```diff
@@ -789,6 +791,29 @@ ln -snf "${BUILDSYS_VERSION_FULL}" "${BUILDSYS_OUTPUT_DIR}/latest"
...
+eval "$(bottlerocket-variant)"
+export BUILDSYS_VARIANT_PLATFORM="${BUILDSYS_VARIANT_PLATFORM:?}"
+export BUILDSYS_VARIANT_RUNTIME="${BUILDSYS_VARIANT_RUNTIME:?}"
+export BUILDSYS_VARIANT_FAMILY="${BUILDSYS_VARIANT_FAMILY:?}"
+export BUILDSYS_VARIANT_FLAVOR="${BUILDSYS_VARIANT_FLAVOR}"
+export PUBLISH_REPO_ROOT_JSON="${PUBLISH_REPO_ROOT_JSON}"
+
+# Save built artifacts for each architecture.  We don't set this everywhere
+# because we build host tools with cargo as well, like buildsys and pubsys.
+export CARGO_TARGET_DIR=${BUILDSYS_ROOT_DIR}/variants/target/${BUILDSYS_ARCH}
```
bcressey: This isn't needed because you're not invoking via `cargo`.

twoliter/embedded/Makefile.toml
```diff
@@ -789,6 +791,29 @@ ln -snf "${BUILDSYS_VERSION_FULL}" "${BUILDSYS_OUTPUT_DIR}/latest"
...
+script = [
+'''
+export PATH="${TWOLITER_TOOLS_DIR}:${PATH}"
+
+# Parse the variant into its components and set additional variables.
+eval "$(bottlerocket-variant)"
+export BUILDSYS_VARIANT_PLATFORM="${BUILDSYS_VARIANT_PLATFORM:?}"
+export BUILDSYS_VARIANT_RUNTIME="${BUILDSYS_VARIANT_RUNTIME:?}"
+export BUILDSYS_VARIANT_FAMILY="${BUILDSYS_VARIANT_FAMILY:?}"
+export BUILDSYS_VARIANT_FLAVOR="${BUILDSYS_VARIANT_FLAVOR}"
```
bcressey: I expect all these could be omitted.

twoliter/embedded/build.Dockerfile.dockerignore
```diff
@@ -8,5 +8,6 @@
 /build/rpms/*-debugsource-*.rpm
 !/build/tools/*
 **/target/*
+/certs
```
bcressey: Can drop this

twoliter/src/tools.rs
```diff
@@ -104,14 +104,17 @@ async fn test_install_tools() {
     // Assert that the expected files exist in the tools directory.
 
     // Check that non-binary files were copied.
-    assert!(toolsdir.join("Dockerfile").is_file());
     assert!(toolsdir.join("Makefile.toml").is_file());
+    assert!(toolsdir.join("build.Dockerfile").is_file());
```
bcressey, line 108: might be good to add the two `.dockerignore` files while you're at it

twoliter/src/tools.rs
```diff
@@ -122,10 +125,18 @@ async fn test_install_tools() {
     assert!(toolsdir.join("tuftool").is_file());
 
     // Check that the mtimes match.
-    let dockerfile_metadata = fs::metadata(toolsdir.join("Dockerfile")).await.unwrap();
+    let build_dockerfile_metadata = fs::metadata(toolsdir.join("build.Dockerfile"))
+        .await
+        .unwrap();
+    let repack_dockerfile_metadata = fs::metadata(toolsdir.join("repack.Dockerfile"))
+        .await
+        .unwrap();
```
bcressey: nit: you don't need to duplicate this check - all the files should have the same mtime, `Dockerfile` was just standing in for one that would always be present.

twoliter/embedded/rpm2img
```diff
@@ -240,6 +206,9 @@ INVENTORY_DATA="$(jq --arg PKG_PREFIX "bottlerocket-" \
 INVENTORY_DATA="$(jq --slurp 'sort_by(.Name)' <<< "${INVENTORY_DATA}" | jq '{"Content": .}')"
 printf "%s\n" "${INVENTORY_DATA}" > "${ROOT_MOUNT}/usr/share/bottlerocket/application-inventory.json"
 
+# install CA certificates
+install_ca_certs
```
bcressey: This should take the arguments it needs, rather than relying on implicit global state:
```suggestion
install_ca_certs "${ROOT_MOUNT}"
```

`partyplanner` is a good example of how to approach this extraction - the end result should be less coupled to the original file.

twoliter/embedded/rpm2img
```diff
@@ -271,166 +240,11 @@ mv "${ROOT_MOUNT}/boot/efi"/* "${EFI_MOUNT}"
...
-
-  # For signing the grub config, we need to embed the GPG public key in binary
-  # form, which is similarly awkward to store in source control.
-  gpg --import "${SBKEYS}/config-sign.key"
-  if [ "${SB_KEY_SOURCE}" == "aws" ] ; then
-    gpg --card-status
-  fi
-  gpg --export > "${SBKEYS}/config-sign.pubkey"
-  gpg --list-keys
+  sbsetup
```
bcressey: Pass these by reference:
```
sbsetup shim_sign_key code_sign_key
```

And then accept them as parameters defined using `local -n` - see `set_partition_labels` in `partyplanner` and its usage.

bcressey (commented):
second round of feedback

tools/buildsys/src/args.rs
```diff
@@ -165,17 +198,23 @@ fn sensitive_env_vars(build_type: BuildType) -> impl Iterator<Item = &'static st
...
     }
 }
 
 /// The thing that buildsys is building.
 #[repr(u8)]
 #[derive(Debug, Clone, Copy)]
 pub(crate) enum BuildType {
+    Repack = 0b00000000,
     Package = 0b00000001,
     Variant = 0b00000010,
```
webern, line 207: Seems like this is an issue. Zero is supposed to mean "no flags are set" but you have assigned one of the flags to zero. Maybe this would fix the cargo rerun output problem.

```suggestion
    Repack = 0b00000001,
    Package = 0b00000010,
    Variant = 0b00000100,
```

twoliter/embedded/repack.Dockerfile.dockerignore
```diff
@@ -0,0 +1,11 @@
+/.git
+/.gomodcache
+/build/*
+!/build/images/*
+!/build/tools/*
```
webern: I don't think these need to go into the docker build context.

```suggestion
```

jpculp: That's actually one of the only directories I _do_ need (for `img2img` and `imghelper`), although I might be able to dial it in with:
```
*
!/build/images/*
!/build/tools/img*
```

jpculp:
Addressed most feedback, but still need to investigate some things related to standard output and other variants.

twoliter/embedded/repack.Dockerfile.dockerignore
webern: Sorry I forgot about the scripts.

jpculp: No worries! It helped me realize that I probably don't need anything besides the scripts and images anyway.

jpculp:
Added support for qcow2 and ova/vmdk.

tools/buildsys/src/builder.rs
```diff
@@ -90,10 +99,16 @@ struct CommonBuildArgs {
     sdk: String,
     nocache: String,
     token: String,
+    clean_output_before_build: bool,
```
bcressey, line 102: nit: boolean parameters tend to create readability issues at call sites.

You could have an enum instead:
```
enum OutputCleanup {
  BeforeBuild,
  None,
}
```

And then it will be more obvious what the flag being passed is for.

tools/buildsys/src/builder.rs
```diff
@@ -264,6 +317,7 @@ impl DockerBuild {
                 &args.common.root_dir,
                 args.common.sdk_image,
                 args.common.arch,
+                true,
```
bcressey, line 320: Following on the above note - this is what I mean about readability issues. There's no way to infer what this "true" does without checking the function signature. If we passed `OutputCleanup::BeforeBuild` it would be clear.

tools/buildsys/src/builder.rs
```diff
@@ -354,6 +409,54 @@ impl DockerBuild {
         })
     }
 
+    /// Create a new `DockerBuild` that can repackage a variant image.
+    pub(crate) fn repack_variant(args: RepackVariantArgs, manifest: &ManifestInfo) -> Result<Self> {
+        let image_layout = manifest.image_layout().cloned().unwrap_or_default();
+
+        let (os_image_publish_size_gib, data_image_publish_size_gib) =
+            image_layout.publish_image_sizes_gib();
```
bcressey, line 417: I expect you also need `os_image_size_gib` , `data_image_size_gib`, and `partition_plan` to do more robust error checking in `img2img`. Basically this should look like the destructuring in `new_variant`.

In particular, it's important that we don't switch from `data_image_size_gib` over to `data_image_publish_size_gib` for AMI targets during a repack operation, because then the data volume becomes unnecessarily large.

Meanwhile the other values are helpful to assert that the input you got matches the variant definition, and that the output you write conforms to it as well.

tools/buildsys/src/builder.rs
```diff
@@ -464,8 +570,17 @@ fn docker(args: &[String], retry: Retry) -> Result<Output> {
...
         let stdout = String::from_utf8_lossy(&output.stdout);
-        println!("{}", &stdout);
+        match args[0].as_str() {
+            "rm" | "rmi" => {
+                if !DOCKER_BUILD_NO_SUCH_THING_ERROR.is_match(&stdout) {
+                    println!("{}", &stdout)
+                }
+            }
+            _ => println!("{}", &stdout),
+        }
```
bcressey, line 582: For a forthcoming fix - to silence the excessive tracebacks on error - I would prefer that we iterate over lines in the output, and stop when we get to a matching line.

Just do it unconditionally; I don't like the `rm` and `rmi` argument checking.

tools/buildsys/src/builder.rs
```diff
@@ -511,7 +626,20 @@ fn secrets_args() -> Result<Vec<String>> {
         );
     }
 
-    for var in &[
+    let ca_bundle_var = "BUILDSYS_CACERTS_BUNDLE_OVERRIDE";
+    let ca_bundle_value =
+        env::var(ca_bundle_var).context(error::EnvironmentSnafu { var: ca_bundle_var })?;
```
bcressey, line 631: This is an interesting case - we also want to support this for the initial `build-variant` since we may wish to override the cert bundle at that time too. So this probably should come in via `args` and we should emit the right cargo change detection variables for it in the build variant case but not the repack variant case (using the flags mechanism).

tools/buildsys/src/builder.rs
```diff
@@ -511,7 +626,20 @@ fn secrets_args() -> Result<Vec<String>> {
...
+
+    if !ca_bundle_value.is_empty() {
+        let ca_bundle_path = PathBuf::from(&ca_bundle_value);
+        if ca_bundle_path.exists() {
+            args.build_secret("file", "ca-bundle.crt", &ca_bundle_path.to_string_lossy());
+        } else {
+            return error::BadCaBundleSnafu { ca_bundle_path }.fail();
+        }
+    }
```
bcressey, line 640: nit: better style is to invert the check and early return to reduce excessive indentation

```suggestion
        if !ca_bundle_path.exists() {
            return error::BadCaBundleSnafu { ca_bundle_path }.fail();
        }
        
        args.build_secret("file", "ca-bundle.crt", &ca_bundle_path.to_string_lossy());
    }
```

twoliter/embedded/repack.Dockerfile.dockerignore
```diff
@@ -0,0 +1,4 @@
+*
+!/build/images/*
+!/build/tools/img*
```
bcressey, line 3: Just pass in all the tools; you don't want to rely on an `img*` prefix for all the shell scripts you need.
```suggestion
!/build/tools/*
```

twoliter/embedded/repack.Dockerfile.dockerignore
```diff
@@ -0,0 +1,4 @@
+*
```
bcressey, line 1:
```suggestion
/
```

twoliter/embedded/repack.Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+WORKDIR /root
+
+USER root
+RUN --mount=target=/host \
+    --mount=type=secret,id=PK.crt,target=/root/sbkeys/PK.crt \
+    --mount=type=secret,id=KEK.crt,target=/root/sbkeys/KEK.crt \
+    --mount=type=secret,id=db.crt,target=/root/sbkeys/db.crt \
+    --mount=type=secret,id=vendor.crt,target=/root/sbkeys/vendor.crt \
+    --mount=type=secret,id=shim-sign.key,target=/root/sbkeys/shim-sign.key \
+    --mount=type=secret,id=shim-sign.crt,target=/root/sbkeys/shim-sign.crt \
```
bcressey, line 30: If `repack.Dockerfile` and `build.Dockerfile` are symlinks to the same `Dockerfile`, does the custom `.dockerignore` still work?

Asking because this is a lot of copy/paste. The way I would refactor it in the shared Dockerfile is to add a new stage that takes the right files and populates some output directory, and then subsequent stages can just copy that output directory wholesale. `rpm-macros-and-bconds` is an example of this.

So if we could reuse the same Dockerfile then you could do that refactor, and then it would be obvious that `rpm2img` and `img2img` got all the Secure Boot related secrets they needed.

Not doing the renames would also avoid some of the looming rebase pain that's otherwise heading your way. And ultimately it's going to be more maintainable than the copy/paste.

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+    exit 1
+  fi
+
+  if [[ "${UEFI_SECURE_BOOT}" == "yes" ]]; then
+    if ! grep -Fq '{{DB_CERT_DER_HEX}}' "${OVF_TEMPLATE}"; then
+      echo "Missing CA certificate field in OVF template, which is required for Secure Boot support." >&2
+      exit 1
+    fi
+  fi
+fi
```
bcressey, line 37: This should be factored out into a common check.

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+      "${INPUT_DIR}/${IMAGE_NAME}-${VARIANT}-${ARCH}-v${VERSION_ID}.ova" \
+      "${OS_IMAGE_NAME%-dirty}.vmdk"
+    qemu-img convert -f vmdk -O raw \
+      "${OS_IMAGE_NAME%-dirty}.vmdk" "${OS_IMAGE}"
+    rm "${OS_IMAGE_NAME%-dirty}.vmdk"
+  fi
+else
+  echo "unexpected image output format '${OUTPUT_FMT}'" >&2
+  exit 1
+fi
```
bcressey, line 75: This can be handled with the check that `rpm2img` does, so we don't have to deal with it here.

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+
+# import the image helper functions and variables
+# shellcheck source=imghelper
+. "${0%/*}/imghelper"
+
+# step 1: decompress the input image
+OS_IMAGE="${OS_IMAGE_NAME}.img"
+if [[ "${OUTPUT_FMT}" == "raw" ]]; then
+  DATA_IMAGE="${INPUT_DIR}/${DATA_IMAGE_NAME}.img.lz4"
+  unlz4 -f "${INPUT_DIR}/${OS_IMAGE_NAME}.img.lz4" "${OS_IMAGE}"
```
bcressey, line 54: It would be nice to have `compress_image` and `decompress_image` functions in the same file, where the adjacency makes it easier to compare.

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+    qemu-img convert -f vmdk -O raw \
+      "${INPUT_DIR}/${OS_IMAGE_NAME}.vmdk" "${OS_IMAGE}"
+  else
+    tar xvf \
+      "${INPUT_DIR}/${IMAGE_NAME}-${VARIANT}-${ARCH}-v${VERSION_ID}.ova" \
+      "${OS_IMAGE_NAME%-dirty}.vmdk"
+    qemu-img convert -f vmdk -O raw \
+      "${OS_IMAGE_NAME%-dirty}.vmdk" "${OS_IMAGE}"
+    rm "${OS_IMAGE_NAME%-dirty}.vmdk"
+  fi
```
bcressey, line 71: I'm not sure entirely what's going on here, or why the `-dirty` tag gets discarded.

That part seems to omit vital information - we are "sanitizing" an image built from a dirty checkout and marking it as clean.

IMO we should treat the OVA as the canonical form and only deal with that, rather than with loose VMDKs.

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+  exit 1
+fi
+
+# step 2: extract the root filesystem
+echo -e "Starting extract of root for ${OS_IMAGE}"
+ROOT_IMAGE="${WORKDIR}/root.ext4"
+ROOT_START="$(sgdisk --print "${OS_IMAGE}" | awk '/BOTTLEROCKET-ROOT-A/{ print $2 }')"
+ROOT_END="$(sgdisk --print "${OS_IMAGE}" | awk '/BOTTLEROCKET-ROOT-A/{ print $3 }')"
+ROOT_SIZE="$((ROOT_END - ROOT_START + 1))"
+dd if="${OS_IMAGE}" of="${ROOT_IMAGE}" count="${ROOT_SIZE}" skip="${ROOT_START}"
```
bcressey, line 83: We are making a lot of assumptions that the partition and filesystem sizes match what the variant's partition plan says they should. Need to avoid that and just check and assert on mismatch instead. This is a sensitive operation - working with images coming from a different trust domain - and we need to be careful.

You could extend `partyplanner` with "get_" methods instead that work on input images and populate similar maps, and then have a compare function that compares all the entries in both maps to see that they match up.

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+mapfile -t new_root_artifacts <<<"$(find "${ROOT_MOUNT}" -type f)"
+if [[ -z "${new_root_artifacts[0]}" ]]; then
+  echo "Nothing to write, skipping..."
+else
+  for artifact in "${new_root_artifacts[@]}"; do
+    cat <<EOF | debugfs -w -f - "${ROOT_IMAGE}"
+rm ${artifact#"${ROOT_MOUNT}"}
+write ${artifact} ${artifact#"${ROOT_MOUNT}"}
+ea_set ${artifact#"${ROOT_MOUNT}"} security.selinux system_u:object_r:os_t:s0
+EOF
```
bcressey, line 102: I think it'd be more readable to write this to a temp file outside of a heredoc, and then pass that file as input to `debugfs`.

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+  done
+fi
+
+# step 4: write the root filesystem back
+echo "Getting root image written back"
+OLD_ROOT_SIZE="$((ROOT_SIZE * 512))"
+NEW_ROOT_SIZE="$(stat --format='%s' "${ROOT_IMAGE}")"
```
bcressey, line 109: It's annoying but any time you have a "size" variable, please suffix it with the unit - sectors, bytes, gibibytes, etc. Here I am inferring that `ROOT_SIZE` is in sectors and `OLD_ROOT_SIZE` is now in bytes, but having that in the variable names will make it less error prone.

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+if [[ ${NEW_ROOT_SIZE} -gt ${OLD_ROOT_SIZE} ]]; then
```
bcressey, line 110: nit: I prefer `[` over `[[` for non-complex conditions, and please follow the usual style of exuberantly double-quoting all variables.

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+  echo "new root image size of ${NEW_ROOT_SIZE} bytes is greater than old size of ${OLD_ROOT_SIZE} bytes" >&2
+  exit 1
+fi
+dd if="${ROOT_IMAGE}" of="${OS_IMAGE}" conv=notrunc seek="${ROOT_START}"
```
bcressey, line 114: Let's continue using `bs=1M` for these operations.
```suggestion
dd if="${ROOT_IMAGE}" of="${OS_IMAGE}" conv=notrunc bs=1M seek="${ROOT_START}"
```

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+echo "Wrote files, recomputing verity"
+
+# step 5: regenerate the verity image
+VERITY_IMAGE="${WORKDIR}/root.verity"
+VERITY_START="$(sgdisk --print "${OS_IMAGE}" | awk '/BOTTLEROCKET-HASH-A/{ print $2 }')"
+VERITY_END="$(sgdisk --print "${OS_IMAGE}" | awk '/BOTTLEROCKET-HASH-A/{ print $3 }')"
```
bcressey, line 120: Having `get_*` functions in `partyplanner` would make this more robust. Here we don't even know at this point that we have these partitions in the input images. It'd be better to collect the data into a map, assert that it looks right, and then use the recorded sizes.

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+VERITY_SIZE="$((VERITY_END - VERITY_START + 1))"
+veritysetup_output="$(veritysetup_format "${ROOT_IMAGE}" "${VERITY_IMAGE}")"
+VERITY_DATA_4K_BLOCKS="$(grep '^Data blocks:' <<<"${veritysetup_output}" | awk '{ print $NF }')"
+VERITY_DATA_512B_BLOCKS="$((VERITY_DATA_4K_BLOCKS * 8))"
+VERITY_ROOT_HASH="$(grep '^Root hash:' <<<"${veritysetup_output}" | awk '{ print $NF }')"
+VERITY_SALT="$(grep '^Salt:' <<<"${veritysetup_output}" | awk '{ print $NF }')"
```
bcressey, line 126: It seems like much or all of this could be a shared helper function - if you were working with the same partsize / partoff maps.

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+mkdir -p "${BOOT_MOUNT}"
+debugfs -R "rdump / ${BOOT_MOUNT}" "${BOOT_IMAGE}"
+
+# step 8: maybe secure boot
+if [[ "${UEFI_SECURE_BOOT}" == "yes" ]]; then
+  EFI_IMAGE="${WORKDIR}/efi.vfat"
+  EFI_START="$(sgdisk --print "${OS_IMAGE}" | awk '/EFI-SYSTEM/{ print $2 }')"
+  EFI_END="$(sgdisk --print "${OS_IMAGE}" | awk '/EFI-SYSTEM/{ print $3 }')"
+  EFI_SIZE="$((EFI_END - EFI_START + 1))"
+  dd if="${OS_IMAGE}" of="${EFI_IMAGE}" count="${EFI_SIZE}" skip="${EFI_START}"
```
bcressey, line 153: I would recommend doing all of these "extract filesystem images from partition" steps in the same place at the outset (after verification).

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+  mcopy -i "${EFI_IMAGE}" -sv ::/EFI/BOOT "${EFI_MOUNT}/EFI/"
+
+  # Do the setup required for `pesign` and `gpg` signing and
+  # verification to "just work" later on, regardless of which
+  # type of signing profile we have.
+  declare -a shim_sign_key code_sign_key
+  sbsetup shim_sign_key code_sign_key
+
+  # remove the original signing certificate to avoid confusion
+  rm "${EFI_MOUNT}/EFI/BOOT/db.cer" "${EFI_MOUNT}/EFI/BOOT/db.crt"
```
bcressey, line 165:
```suggestion
  # remove the original signing certificate to avoid confusion
  rm -f "${EFI_MOUNT}/EFI/BOOT/db.cer" "${EFI_MOUNT}/EFI/BOOT/db.crt"
```

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+
+  # replace artifacts in efi filesystem
+  pushd "${EFI_MOUNT}/EFI/BOOT" >/dev/null
+  replace_vendor_certs shim_sign_key code_sign_key "${UEFI_SECURE_BOOT}"
+  mcopy -i "${EFI_IMAGE}" -ov "${EFI_MOUNT}/EFI/BOOT"/*.efi ::/EFI/BOOT
+  mcopy -i "${EFI_IMAGE}" -ov "${SBKEYS}"/db.{crt,cer} ::/EFI/BOOT
+  popd >/dev/null
```
bcressey, line 172: It's weird to change the directory for `replace_vendor_certs` to work properly - better for it to take a directory argument. Also, why pass `UEFI_SECURE_BOOT`? We know it's true in this branch.

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+# step 8: maybe secure boot
+if [[ "${UEFI_SECURE_BOOT}" == "yes" ]]; then
+  EFI_IMAGE="${WORKDIR}/efi.vfat"
+  EFI_START="$(sgdisk --print "${OS_IMAGE}" | awk '/EFI-SYSTEM/{ print $2 }')"
+  EFI_END="$(sgdisk --print "${OS_IMAGE}" | awk '/EFI-SYSTEM/{ print $3 }')"
+  EFI_SIZE="$((EFI_END - EFI_START + 1))"
+  dd if="${OS_IMAGE}" of="${EFI_IMAGE}" count="${EFI_SIZE}" skip="${EFI_START}"
+  EFI_MOUNT="${WORKDIR}/efi"
+  mkdir -p "${EFI_MOUNT}/EFI/BOOT"
+  mcopy -i "${EFI_IMAGE}" -sv ::/EFI/BOOT "${EFI_MOUNT}/EFI/"
```
bcressey, line 156: Can we just extract all of it? It's weird dealing with `EFI` and `EFI/BOOT` subdirectories.

```suggestion
  mcopy -i "${EFI_IMAGE}" -sv ::/EFI "${EFI_MOUNT}"
```

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+  mcopy -i "${EFI_IMAGE}" -ov "${SBKEYS}"/db.{crt,cer} ::/EFI/BOOT
+  popd >/dev/null
+
+  # write the efi image back
+  dd if="${EFI_IMAGE}" of="${OS_IMAGE}" conv=notrunc seek="${EFI_START}"
+
+  pushd "${BOOT_MOUNT}" >/dev/null
+
+  # resign the kernel
+  vmlinuzsetup vmlinuz
```
bcressey, line 180: Definitely needs a name that makes it more obvious this is a Secure Boot signing related function.

Also, these functions should operate on a directory parameter rather than current working directory.

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+  exit 1
+fi
+
+# step 10: write the boot filesystem back
+dd if="${BOOT_IMAGE}" of="${OS_IMAGE}" conv=notrunc seek="${BOOT_START}"
+
+# now that the images have been written byte for byte back,
+# we can resize them before compressing
+resize2fs -M "${ROOT_IMAGE}"
+resize2fs -M "${BOOT_IMAGE}"
```
bcressey, line 244: I don't like this - it means that the hashes of the partition contents extracted from a disk image would not match the hashes of the files used for in-place updates.

Also you must be getting lucky here and the resize of the root filesystem image isn't doing anything; otherwise you would invalidate the dm-verity hash. Or maybe you are and then filesystem reads to certain contents would randomly trigger a kernel panic.

bcressey (commented):
Still haven't really dug into `imghelper` or the `rpm2img` refactor.

twoliter/embedded/imghelper
```diff
@@ -0,0 +1,398 @@
...
+  if [[ -s "${data_vmdk}" ]] ; then
+     tar -rf "${OUTPUT_DIR}/${ova}" -C "${OUTPUT_DIR}" "${data_vmdk}"
+  fi
+
+  symlink_image "ova" "os_image"
+  rm -rf "${ova_dir}"
+}
+
+symlink_image() {
+  local ext what
```
bcressey, line 389: It works in this context, but I'd still avoid defining a function after the places where it's called.

twoliter/embedded/imghelper
```diff
@@ -0,0 +1,398 @@
...
+  local ovf_template uefi_secure_boot
+  os_image_publish_size_gib="${1:?}"
+  data_image_publish_size_gib="${2:?}"
+  ovf_template="${3:?}"
+  uefi_secure_boot="${4:?}"
+
+  os_vmdk="${OS_IMAGE_NAME}.vmdk"
+  data_vmdk="${DATA_IMAGE_NAME}.vmdk"
+  ovf="${OS_IMAGE_NAME}.ovf"
+  ova_dir="$(mktemp -d)"
```
bcressey, line 323: It would be good to make all of these variables `local` now. The previous style wasn't amazing but the risk was lower because the execution in `rpm2img` only flowed one way.

As a separate helper script, the execution order is up to the caller, so we should be more rigorous about leaking variables into the global scope where they could clobber a previous assignment unexpectedly, or create subtle dependencies that would break on refactor.

twoliter/embedded/imghelper
```diff
@@ -0,0 +1,398 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC2034,SC2154,SC2164,SC2312
+
+# Store output artifacts in a versioned directory.
+OUTPUT_DIR="${OUTPUT_DIR}/${VERSION_ID}-${BUILD_ID}"
```
bcressey, line 5: It's very confusing to clobber a global like this when the file is sourced.

twoliter/embedded/imghelper
```diff
@@ -0,0 +1,398 @@
...
+
+FILENAME_PREFIX="${IMAGE_NAME}-${VARIANT}-${ARCH}-${VERSION_ID}-${BUILD_ID}"
+SYMLINK_PREFIX="${IMAGE_NAME}-${VARIANT}-${ARCH}"
+VERSIONED_SYMLINK_PREFIX="${IMAGE_NAME}-${VARIANT}-${ARCH}-${VERSION_ID}"
+FRIENDLY_VERSIONED_SYMLINK_PREFIX="${IMAGE_NAME}-${VARIANT}-${ARCH}-v${VERSION_ID}"
+
+OS_IMAGE_NAME="${FILENAME_PREFIX}"
+OS_IMAGE_SYMLINK="${SYMLINK_PREFIX}"
+OS_IMAGE_VERSIONED_SYMLINK="${VERSIONED_SYMLINK_PREFIX}"
+OS_IMAGE_FRIENDLY_VERSIONED_SYMLINK="${FRIENDLY_VERSIONED_SYMLINK_PREFIX}"
```
bcressey, line 15: You should approach this refactor with the concept of a single owner for global variables in mind. If this file "owns" them, all its functions are allowed to use them. If not, all its functions must receive the variable as a parameter.

Same for the files that source this - `rpm2img` and `img2img`. It is OK to move ownership here but then you can't use the variables there. You either have to provide a function that takes some context from the caller, or you have to change your functions to take the global as a parameter so you don't transfer ownership.

twoliter/embedded/imghelper
```diff
@@ -0,0 +1,398 @@
...
+ROOT_IMAGE_FRIENDLY_VERSIONED_SYMLINK="${FRIENDLY_VERSIONED_SYMLINK_PREFIX}-root.ext4.lz4"
+
+CROSS_FACTORYDIR="/usr/share/factory"
+CROSS_SYSCONFIGDIR="/etc"
+CROSS_TMPFILESDIR="/usr/lib/tmpfiles.d"
+
+CA_CERTS_BUNDLE="${HOME}/certs/ca-bundle.crt"
+CA_CERTS_DEFAULT_BUNDLE="/usr/share/bottlerocket/ca-certificates/ca-bundle.crt"
+CA_CERTS_DEFAULT_LICENSES="/usr/share/licenses/ca-certificates"
+CA_CERTS_INSTALLDIR="${CROSS_FACTORYDIR}${CROSS_SYSCONFIGDIR}/pki/tls/certs"
```
bcressey, line 44: These shouldn't be globals; they can be defined in `install_ca_certs`.

twoliter/embedded/imghelper
```diff
@@ -0,0 +1,398 @@
...
+  # For an AWS profile, we expect a config file for the PKCS11
+  # helper. Otherwise, there should be a local key and cert.
+  if [[ -s "${HOME}/.config/aws-kms-pkcs11/config.json" ]]; then
+    # Set AWS environment variables from build secrets, if present.
+    for var in AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN; do
+      val="${var,,}"
+      val="${HOME}/.aws/${val//_/-}.env"
+      [[ -s "${val}" ]] || continue
+      declare -x "${var}=$(cat "${val}")"
+    done
```
bcressey, line 136: Let's split this into two helper functions - `sbsetup_aws_profile` and `sbsetup_local_profile` - that we call depending on which branch we take. Effectively they both have the same interface - populate the `shim_sign_key` and `code_sign_key` args - so they can just take those by reference.

twoliter/embedded/imghelper
```diff
@@ -0,0 +1,398 @@
...
+  local uefi_secure_boot
+  shim_sign_key="${1:?}"
+  code_sign_key="${2:?}"
+  uefi_secure_boot="${3:?}"
+  shims=(boot*.efi)
+  shim="${shims[0]}"
+  grubs=(grub*.efi)
+  grub="${grubs[0]}"
+  mokms=(mm*.efi)
+  mokm="${mokms[0]}"
```
bcressey, line 212: These should all be locals.

twoliter/embedded/imghelper
```diff
@@ -0,0 +1,398 @@
...
+CROSS_FACTORYDIR="/usr/share/factory"
+CROSS_SYSCONFIGDIR="/etc"
+CROSS_TMPFILESDIR="/usr/lib/tmpfiles.d"
+
+CA_CERTS_BUNDLE="${HOME}/certs/ca-bundle.crt"
+CA_CERTS_DEFAULT_BUNDLE="/usr/share/bottlerocket/ca-certificates/ca-bundle.crt"
+CA_CERTS_DEFAULT_LICENSES="/usr/share/licenses/ca-certificates"
+CA_CERTS_INSTALLDIR="${CROSS_FACTORYDIR}${CROSS_SYSCONFIGDIR}/pki/tls/certs"
+
+SBKEYS="${HOME}/sbkeys"
```
bcressey, line 46: I am torn on where ownership of `SBKEYS` should live but it's one of the globals that's crossing the boundary in a way that I don't like.

twoliter/embedded/imghelper
```diff
@@ -0,0 +1,398 @@
...
+
+VERITY_VERSION=1
+VERITY_HASH_ALGORITHM=sha256
+VERITY_DATA_BLOCK_SIZE=4096
+VERITY_HASH_BLOCK_SIZE=4096
```
bcressey, line 51: The verity stuff should probably just be a `do_verity_for_image` function that takes parameters. Unlike the `ca-cert` ones, I like these as globals because they're effectively constants that can't be changed without a massive compatibility break.

Please also look to `partyplanner` for inspiration on comments on both function calls and important constants.

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+  # type of signing profile we have.
+  declare -a shim_sign_key code_sign_key
+  sbsetup shim_sign_key code_sign_key
+
+  # remove the original signing certificate to avoid confusion
+  rm "${EFI_MOUNT}/EFI/BOOT/db.cer" "${EFI_MOUNT}/EFI/BOOT/db.crt"
+
+  # replace artifacts in efi filesystem
+  pushd "${EFI_MOUNT}/EFI/BOOT" >/dev/null
+  replace_vendor_certs shim_sign_key code_sign_key "${UEFI_SECURE_BOOT}"
```
bcressey, line 169: This smells like an error - we need to call this even if we are not doing Secure Boot for consistency with `rpm2img`, because this is what does the no-op replacements in shim and grub. But because we know it's a Secure Boot image, we know that doesn't apply to the artifacts we've extracted.

I'd recommend splitting the function into two, like `replace_vendor_certs` and `replace_dummy_certs`, and then `rpm2img` can call one or the other depending on whether Secure Boot is active.

twoliter/embedded/imghelper
```diff
@@ -0,0 +1,398 @@
...
+  shim_sign_key="${1:?}"
+  code_sign_key="${2:?}"
+  uefi_secure_boot="${3:?}"
+  shims=(boot*.efi)
+  shim="${shims[0]}"
+  grubs=(grub*.efi)
+  grub="${grubs[0]}"
+  mokms=(mm*.efi)
+  mokm="${mokms[0]}"
+  if [[ "${uefi_secure_boot}" == "yes" ]]; then
```
bcressey, line 213: See other note - split this function to remove the need for this branch.

twoliter/embedded/imghelper
```diff
@@ -0,0 +1,398 @@
...
+    mv "${shim}.signed" "${shim}"
+    pesign -i "${shim}" -l
+    pesigcheck -i "${shim}" -n 0 -c "${SBKEYS}/db.cer"
+
+    # Remove the original key if necessary, then sign the MOK manager as well.
+    if [[ "$(pesign -i "${mokm}" -l)" == "No signatures found." ]]; then
+      mv "${mokm}" "${mokm}.unsigned"
+    else
+      pesign -i "${mokm}" -o "${mokm}.unsigned" -u 0 -r
+    fi
```
bcressey, line 231: Why is this necessary for mokm and grub but not for shim? It would be cleaner to treat all the same way.

twoliter/embedded/imghelper
```diff
@@ -0,0 +1,398 @@
...
+  # form, which is similarly awkward to store in source control.
+  gpg --batch --no-tty --import "${SBKEYS}/config-sign.key"
+  if [[ "${sb_key_source}" == "aws" ]]; then
+    gpg --batch --no-tty --card-status
+  fi
+  gpg --batch --no-tty --export >"${SBKEYS}/config-sign.pubkey"
+  gpg --batch --no-tty --list-keys
+}
+
+replace_vendor_certs() {
```
bcressey, line 201: Naming isn't great here - this isn't just doing "replace vendor certs", it's doing the entire signing operation for multiple artifacts.

twoliter/embedded/imghelper
```diff
@@ -0,0 +1,398 @@
...
+  | awk -v vmlinuz="${vmlinuz}" '{ print $2 "  " vmlinuz }' \
+  > ".${vmlinuz}.hmac"
+}
+
+sign_grub(){
+  local grub_cfg
+  grub_cfg="${1:?}"
+  gpg --batch --no-tty --detach-sign "${grub_cfg}"
+  gpg --batch --no-tty --verify "${grub_cfg}.sig" "${grub_cfg}"
+}
```
bcressey, line 297: These functions implicitly assume that the current working directory is the same as the `vmlinuz` or `grub.cfg` that they're operating. It would be good to correct that - take a full path to those artifacts and then use `"${vmlinuz##*/}"` to compute the target directory.

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+VERITY_OPTS_PATTERN+=" [0-9a-f]\+ [0-9a-f]\+"
+
+VERITY_OPTS="${VERITY_DATA_BLOCK_SIZE} ${VERITY_HASH_BLOCK_SIZE} ${VERITY_DATA_4K_BLOCKS}"
+VERITY_OPTS+=" 1 ${VERITY_HASH_ALGORITHM}"
+VERITY_OPTS+=" ${VERITY_ROOT_HASH} ${VERITY_SALT}"
+
+sed -i \
+  -e "s,\(.*\) ${DM_MOD_OPTS_PATTERN} \(.*\),\1 ${DM_MOD_OPTS} \2," \
+  -e "s,\(.*\) ${VERITY_OPTS_PATTERN} \(.*\),\1 ${VERITY_OPTS} \2," \
+  "${BOOT_MOUNT}/grub/grub.cfg"
```
bcressey, line 213: This would be better as a helper function.

twoliter/embedded/img2img
```diff
@@ -0,0 +1,288 @@
...
+
+if [[ "${UEFI_SECURE_BOOT}" == "yes" ]]; then
+  rm -f "${BOOT_MOUNT}/grub/grub.cfg.sig"
+  sign_grub "${BOOT_MOUNT}/grub/grub.cfg"
```
bcressey, line 217: `sign_grub` can take care of removing previous signature files if they exist.

twoliter/embedded/rpm2img
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# shellcheck disable=SC2034
+# shellcheck disable=SC2034,SC2154,SC2249,SC2312
```
bcressey, line 2: nit: can you list out which lints you're disabling and why?

twoliter/embedded/rpm2img
```diff
@@ -79,40 +79,13 @@ if [ "${OUTPUT_FMT}" == "vmdk" ] ; then
   fi
 fi
 
+# import the imager helper functions and variables
```
bcressey, line 82: "variables" 🤨

bcressey (commented):
Finally made it all the way through. Overall I like the direction, just had a lot of suggestions for code clarity and maintainability that will help keep the code readable.

Part of why I consider the `partyplanner` refactor a success is that it's very rare to need to modify it along with `rpm2img`. What we want to create is enough of separation of concerns that we can modify `imghelper` by itself and not also coordinate that with `img2img` and `rpm2img`.

tools/buildsys/src/builder.rs
jpculp: At the moment the publish sizes are only being used for generating the OVA. Aside from that (and generating a symlink) the data partition is left alone. Since we're replacing files in the existing images rather than creating new ones the size shouldn't fluctuate meaningfully. We could add checks to verify the input matches the variant definition, but since the build id is passed along the only way I think you could get a mismatch is if you built off a dirty branch, changed the definition, then repacked it. Should I add the checks regardless?

tools/buildsys/src/builder.rs
jpculp: We could, but I think we would actually want to see this error if we're not doing a cleanup task since that would indicate something is seriously off. We also wouldn't have to add logic to iterate through lines in the string as the cleanup tasks only return the single line.

tools/buildsys/src/builder.rs
jpculp: Already ahead of you! The cert installation is a helper function in `imghelper` so that the code could be shared between `rpm2img` and `img2img`. I did not add cargo change detection variables though so I'll add that to the to-do list.

twoliter/embedded/img2img
jpculp: For testing I have to add:
```toml
[tasks.repack-variant]
run_task = "run-twoliter"
```
which makes my branch `-dirty`, but I plan to remove it before pulling out of draft.

As far as loose VMDKs, yeah we can remove that since the worst we'd do is untar on top of it.

twoliter/embedded/img2img
jpculp: Sure. I just removed them since we're doing an `int` comparison here.

twoliter/embedded/img2img
jpculp: I saw a lot of really weird behavior when I was setting the blocksize manually. I might have to show you offline.

twoliter/embedded/img2img
jpculp: In the case that UEFI_SECURE_BOOT is "no" this block never gets called and we don't modify EFI at all. My hope is we only touch the bare minimum when repacking. If we do want to touch both though, I wonder if we should just generate a fresh EFI image each time.

twoliter/embedded/img2img
jpculp: It's true in `img2img`, but not in `rpm2img`. However, if we split up the function into `replace_vendor_certs` and `replace_dummy_certs` we can drop it.

twoliter/embedded/img2img
jpculp: That's fine. I was mostly just mimicking `rpm2img`'s current behavior, but I agree that working with absolute paths is more straightforward.

twoliter/embedded/img2img
jpculp: It must be the former. The differences in the certs and signatures haven't been enough to actually require a resize thus far.

twoliter/embedded/imghelper
jpculp: I agree! This was just another oddity carried over from `rpm2img`, but I would prefer having this be a `VERSIONED_OUTPUT_DIR` instead.

twoliter/embedded/imghelper
jpculp: After running `objdumpcopy` the signature disappears, but I can add the remove before the `objdumpcopy` to be safe and consistent.

twoliter/embedded/imghelper
jpculp: It looks like we do this clobber in multiple scripts. Instead I wonder if we should just pass the version output directory from the start.

twoliter/embedded/repack.Dockerfile
jpculp: Gave this a shot, but no dice.

twoliter/embedded/imghelper
bcressey, line 44: I don't love the systemd / tmpfiles specifics leaking in here. I'd prefer that we add the tmpfiles snippet to the `release` package and have it copy from the `/usr/share/bottlerocket` path where we know it will be written, with a comment there on where it's coming from and why it's unusual.

jpculp (commented):
Since this is a bit hard to review I'm going to split this up into several bite-sized PRs. I'll leave this open in draft for a tiny bit longer just to provide some big picture context.

jpculp:
Superseded by #231.

PR #208, opened 2024-05-08
Fix tmp files being created in project directory

### LINKED ISSUE #180: twoliter build variant creates a random directory in the project
## Description

`twoliter build variant` creates a directory with a name like `.tmpERPdi` in the root of the project directory. It shouldn't.

## Definition of Done

`twoliter build variant` no longer creates a random directory in the root of the project.

sumukhballal:
**Issue number:**

Closes #180

**Description of changes:**

This issue is probably reproducible on every variant (tested on 1) when SIGINT or SIGSTP (or SIGTERM) is called on the build process in a linux host, temp files are not cleared. This is expected behavior as per https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#resource-leaking as the destructors are never called. The build.rs where these files are created are [here](https://github.com/bottlerocket-os/twoliter/blob/develop/twoliter/src/cmd/build.rs#L60).

Instead we will create these files under /tmp as suggested by @webern . This allows us to keep the current build directory clean.

**Testing done:**

When a `SIGINT` or `SIGSTP` (ctrl+c and ctrl+z) is called, I see the file created in the tmp directory. For example:

```
/tmp/.tmpXYZ
```

When no such interruption is provided, watching a directory, I see a file created, then deleted eventually when it goes out of scope.

```
Every 2.0s: ls -a | grep "^\.tmp"                                                                                             Wed May  8 09:45:02 2024

.tmpi6qmGD
.tmpM1RQq1
```

But eventually it was cleared
```
Every 2.0s: ls -a | grep "^\.tmp"                                                                                             Wed May  8 09:45:02 2024

.tmpM1RQq1
```

**Terms of contribution:**

### DIFF
--- a/twoliter/src/cmd/build.rs
+++ b/twoliter/src/cmd/build.rs
@@ -57,7 +57,7 @@ impl BuildVariant {
         install_tools(&toolsdir).await?;
         let makefile_path = toolsdir.join("Makefile.toml");
         // A temporary directory in the `build` directory
-        let build_temp_dir = TempDir::new_in(project.project_dir())
+        let build_temp_dir = TempDir::new_in("/tmp")
             .context("Unable to create a tempdir for Twoliter's build")?;
         let packages_dir = build_temp_dir.path().join("sdk_rpms");
         fs::create_dir_all(&packages_dir).await?;


twoliter/src/cmd/build.rs
```diff
@@ -57,7 +57,7 @@ impl BuildVariant {
         install_tools(&toolsdir).await?;
         let makefile_path = toolsdir.join("Makefile.toml");
         // A temporary directory in the `build` directory
-        let build_temp_dir = TempDir::new_in(project.project_dir())
+        let build_temp_dir = TempDir::new_in("/tmp")
```
bcressey, line 60: /tmp can be a memory-backed tmpfs so it is unwise to leak files there - and can be pretty annoying to clean up.

Let's use a path like build/.tmp that we recursively delete and create fresh.

webern: > Let's use a path like build/.tmp that we recursively delete and create fresh.

Note that an `rm -rf "${BUILDSYS_BUILD_DR}/.tmp"` needs to be added to `Makefile.toml` in the appropriate `clean` task if we do this here. Also, a `create_dir_all` is probably required here for the `build` directory because we are doing these things before we go into `cargo make`.

webern:
This was fixed by #286

Issue #73, opened 2023-09-20
kits: integrate with Makefile.toml and Dockerfile


webern:
## Description

Integrate the `twoliter build kit` command into the Makefile.toml ecosystem. Make it possible to build a variant where the packages are built into a kit first. Possibly there should be a default mode that detects the absence of kits in the project and creates a single all-packages kit such that the Bottlerocket build still works without yet introducing Bottlerocket's kits (just an idea).

## Definition of Done

When `cargo make kit` works. And when `twoliter build kit` invokes the proper Makefilte.toml prerequisites before building the kit.

PR #209, opened 2024-05-10
update twoliter and buildsys to put RPMs into per-package sub-directories

webern:
**Issue number:**

Related #185
Related #73
Builds on #198

**Description of changes:**

In order to construct kits containing the correct RPMs, we need to stop using a flat `build/rpms` structure. Instead we need to put each RPM into a sub-directory. Then we can use the package-name to find the right RPMs to build repos with the correct RPMs on-the-fly when building packages, kits and variants.

**Testing done:**

- [ ] Build a Bottlerocket Variant
- [ ] Manifest unit tests are passing

**Terms of contribution:**


webern:
Sorry, closing this because I meant to push a branch to upstream for PR chaining.

PR #210, opened 2024-05-10
update twoliter and buildsys to put RPMs into per-package sub-directories

webern:
**Issue number:**

Related #185
Related #73
Builds on #198

**Description of changes:**

In order to construct kits containing the correct RPMs, we need to stop using a flat `build/rpms` structure. Instead we need to put each RPM into a sub-directory. Then we can use the package-name to find the right RPMs to build repos with the correct RPMs on-the-fly when building packages, kits and variants.

**Testing done:**

- [x] Build a Bottlerocket Variant
- [x] unit tests are passing
- [x] Built a Bottlerocket variant from develop (flat RPM structure) then tried to do an incremental build without the `twoliter: error if old rpm structure is discovered` commit. Buildsys failed as expected with `No matching package to install: 'bottlerocket-glibc-devel'`
- [x] Tried again with `twoliter: error if old rpm structure is discovered` , got the helpful message: `PLEASE RUN: cargo make clean`
- [x] Tried again after a `cargo make clean` and the build succeeded

**Terms of contribution:**

### DIFF
--- a/tools/buildsys/src/builder.rs
+++ b/tools/buildsys/src/builder.rs
@@ -302,12 +302,7 @@ impl DockerBuild {
         manifest: &Manifest,
         image_features: HashSet<ImageFeature>,
     ) -> Result<Self> {
-        let package = if let Some(name_override) = manifest.info().package_name() {
-            name_override.clone()
-        } else {
-            args.cargo_package_name
-        };
-
+        let package = manifest.info().package_name();
         Ok(Self {
             dockerfile: args.common.tools_dir.join("build.Dockerfile"),
             context: args.common.root_dir.clone(),
@@ -323,7 +318,7 @@ impl DockerBuild {
             root_dir: args.common.root_dir.clone(),
             artifacts_dir: args.packages_dir,
             state_dir: args.common.state_dir,
-            artifact_name: package.clone(),
+            artifact_name: package.to_string(),
             common_build_args: CommonBuildArgs::new(
                 &args.common.root_dir,
                 args.common.sdk_image,
@@ -332,7 +327,7 @@ impl DockerBuild {
             ),
             target_build_args: TargetBuildArgs::Package(PackageBuildArgs {
                 image_features,
-                package,
+                package: package.to_string(),
                 package_dependencies: manifest.package_dependencies().context(error::GraphSnafu)?,
                 kit_dependencies: manifest.kit_dependencies().context(error::GraphSnafu)?,
                 publish_repo: args.publish_repo,
--- a/tools/buildsys/src/builder.rs
+++ b/tools/buildsys/src/builder.rs
@@ -303,6 +303,8 @@ impl DockerBuild {
         image_features: HashSet<ImageFeature>,
     ) -> Result<Self> {
         let package = manifest.info().package_name();
+        let per_package_dir = format!("{}/{}", args.packages_dir.display(), package).into();
+
         Ok(Self {
             dockerfile: args.common.tools_dir.join("build.Dockerfile"),
             context: args.common.root_dir.clone(),
@@ -316,7 +318,7 @@ impl DockerBuild {
                 &args.common.root_dir,
             ),
             root_dir: args.common.root_dir.clone(),
-            artifacts_dir: args.packages_dir,
+            artifacts_dir: per_package_dir,
             state_dir: args.common.state_dir,
             artifact_name: package.to_string(),
             common_build_args: CommonBuildArgs::new(
--- a/twoliter/embedded/rpm2kit
+++ b/twoliter/embedded/rpm2kit
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+#
+# Create a kit from RPM package inputs.
+set -eu -o pipefail
+
+declare -a PACKAGES
+
+for opt in "$@"; do
+   optarg="$(expr "${opt}" : '[^=]*=\(.*\)')"
+   case "${opt}" in
+      --packages-dir=*) PACKAGES_DIR="${optarg}" ;;
+      --package=*) PACKAGES+=("${optarg}") ;;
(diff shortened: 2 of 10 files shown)


webern:
> https://github.com/bottlerocket-os/twoliter/compare/43ed803ea9183599a333687c97dd9fd3e5b5026e..3ded669d9a51d82e304a4555c75c71a2c6c1a5bf

Clippy Fix (boring)

tools/buildsys/src/builder.rs
```diff
@@ -247,11 +247,13 @@ impl DockerBuild {
...
-        let package = if let Some(name_override) = manifest.info().package_name() {
-            name_override.clone()
-        } else {
-            args.cargo_package_name
-        };
+        let package = manifest.info().package_name();
+        let per_package_dir = format!(
+            "{}/{}",
+            args.packages_dir.display(),
+            manifest.info().manifest_name()
```
bcressey: Are we able to use `package` here?
```suggestion
        let per_package_dir = format!(
            "{}/{}",
            args.packages_dir.display(),
            package,
```

twoliter/embedded/Dockerfile.dockerignore
```diff
@@ -3,9 +3,10 @@
 /build/*
 !/build/rpms/
 /build/rpms/*
-!/build/rpms/*.rpm
-/build/rpms/*-debuginfo-*.rpm
-/build/rpms/*-debugsource-*.rpm
+!/build/rpms/*/*.rpm
+/build/rpms/*/*-debuginfo-*.rpm
+/build/rpms/*/*-debugsource-*.rpm
+!/build/kits/
```
bcressey: Maybe keep the earlier exclusions for now, with a comment that we can deprecate them eventually. Otherwise the build context will get very large for anyone that has existing files in `build/rpms` until they clean their checkout.
```suggestion
!/build/rpms/*.rpm
/build/rpms/*-debuginfo-*.rpm
/build/rpms/*-debugsource-*.rpm
!/build/rpms/*/*.rpm
/build/rpms/*/*-debuginfo-*.rpm
/build/rpms/*/*-debugsource-*.rpm
!/build/kits/
```

twoliter/embedded/rpm2kit
```diff
@@ -0,0 +1,38 @@
...
+   optarg="$(expr "${opt}" : '[^=]*=\(.*\)')"
+   case "${opt}" in
+      --packages-dir=*) PACKAGES_DIR="${optarg}" ;;
+      --package=*) PACKAGES+=("${optarg}") ;;
+      --output-dir=*) OUTPUT_DIR="${optarg}" ;;
+   esac
+done
+
+# HACK: because we're not passing KIT from buildsys
+export KIT="${VARIANT}"
```
bcressey:
```suggestion
```

twoliter/embedded/rpm2kit
```diff
@@ -0,0 +1,38 @@
...
+
+# FIXME: should this be a versioned directory?
+# - needs to change whenever the kit changes?
+# - needs to be predictable for publishing to work?
+# - needs to avoid ambiguity if we have other local builds of the same kit?
```
bcressey: If we're emitting the right `cargo:rerun-if-changed` and `cargo:rerun-if-env-changed` directives in `buildsys` then we can drop this. Just need to make sure https://github.com/bottlerocket-os/twoliter/pull/198#discussion_r1594576970 is addressed.

bcressey (approved):
Looks good, just some minor notes.

twoliter/embedded/rpm2kit
webern: I think the answer is as simple as `rm -rf "${KIT_DIR}"` since, if this script is running, then we know we are re-creating the kit. That's what I've pushed.

twoliter/embedded/Dockerfile.dockerignore
webern: I've committed this suggestion, but don't we actually want to ignore `/build/rpms/*.rpm` to solve that problem?

webern:
> https://github.com/bottlerocket-os/twoliter/compare/6eaabe9a12d433f94b32747ba50ab05d2ee4c271..2108263fb3c1a34e932451132cc45e808fe0bb46

Rebase over #214

webern:
> https://github.com/bottlerocket-os/twoliter/compare/3183c93da746776a9605b2008f2f036b1c22ba6c..b9d7639d19fe35c0ba44dd98f13782a1b30f8356

Add a check to Makefile.toml to tell the user when they need to run `cargo make clean`.

webern:
@bcressey this is ready for another review after adding the flat RPMs dir check.

bcressey (approved):

webern:
> https://github.com/bottlerocket-os/twoliter/compare/b9d7639d19fe35c0ba44dd98f13782a1b30f8356..737ca5d2cca82b22a56e4647c037b72fbc2c8131

Rebase to fix Makefile.toml conflict

twoliter/embedded/rpm2kit
```diff
@@ -0,0 +1,32 @@
...
+    -mindepth 1 \
+    -maxdepth 1 \
+    ! -name '*-debuginfo-*' \
+    ! -name '*-debugsource-*' \
+    -size +0c \
+    -exec install -p -m 0644 -t "${KIT_DIR}/Packages" {} \+
+done
+
+createrepo_c "${KIT_DIR}"
+dnf --disablerepo '*' --repofrompath "kit,file:///${KIT_DIR}" repoquery --all
```
sam-berning, line 32: just for my understanding, what does this do? is it just disabling the kit dnf repo by default so that it has to be explicitly enabled in the variant build?

sam-berning (approved):

bcressey (approved):

PR #214, opened 2024-05-14
add repack-variant

jpculp:
**Description of changes:**

Add the the initial target for repackaging a Bottlerocket variant. This target will skip the traditional workflow of building packages and using `rpm2img`. Instead the build system will point to a yet to be implemented `img2img` tool.

Includes some preliminary housekeeping. Removes some unnecessary checks, references, and dependencies. Renames `Dockerfile` to `build.Dockerfile` and adds a common argument for future build targets to skip cleanup if desired.

**Testing done:**

- Built bottlerocket with the default make target
- Verified cleanup occurred when built a second time
- Attempted build with bottlerocket with the repack target
- Verified cleanup did not occur when repacked a second time

**Terms of contribution:**


tools/buildsys/src/args.rs
```diff
@@ -183,6 +204,7 @@ pub(crate) fn rerun_for_envs(build_type: BuildType) {
 #[repr(u8)]
 #[derive(Debug, Clone, Copy)]
 enum BuildFlags {
+    Repack = 0b00000000,
```
bcressey: At least one of the bits should be set to determine which build type it is
```suggestion
    Repack = 0b00001000,
```

tools/buildsys/src/manifest.rs
```diff
@@ -504,6 +513,7 @@ fn is_manifest_type(pkg_metadata: &PackageMetadata, manifest_type: BuildType) ->
         BuildType::Package => metadata_table.get("build-package").is_some(),
         BuildType::Kit => metadata_table.get("build-kit").is_some(),
         BuildType::Variant => metadata_table.get("build-variant").is_some(),
+        BuildType::Repack => metadata_table.get("repack-variant").is_some(),
```
bcressey: This shouldn't be necessary, it's only for builds and won't apply to repacking. There's no new "type" of crate, just the three existing types: package, kit, and variant.

tools/buildsys/src/manifest.rs
```diff
@@ -486,6 +488,13 @@ impl ManifestInfo {
             .as_ref()
             .and_then(|m| m.build_variant.as_ref())
     }
+
+    fn repack_variant(&self) -> Option<&RepackVariant> {
+        self.package
+            .metadata
+            .as_ref()
+            .and_then(|m| m.repack_variant.as_ref())
+    }
```
bcressey: not needed

tools/buildsys/src/manifest.rs
```diff
@@ -606,6 +617,16 @@ pub struct BuildVariant {
...
+#[derive(Deserialize, Debug)]
+#[serde(rename_all = "kebab-case")]
+pub struct RepackVariant {
+    pub image_format: Option<ImageFormat>,
+    #[serde(default)]
+    pub image_layout: ImageLayout,
+    pub supported_arches: Option<HashSet<SupportedArch>>,
+    pub image_features: Option<HashMap<ImageFeature, bool>>,
+}
+
```
bcressey: We should only need the BuildVariant struct.

tools/buildsys/src/manifest.rs
```diff
@@ -460,6 +460,8 @@ impl ManifestInfo {
             Ok(BuildType::Kit)
         } else if self.build_variant().is_some() {
             Ok(BuildType::Variant)
+        } else if self.repack_variant().is_some() {
+            Ok(BuildType::Repack)
```
bcressey: not needed
```suggestion
```

tools/buildsys/src/manifest.rs
```diff
@@ -556,6 +566,7 @@ struct Metadata {
     build_package: Option<BuildPackage>,
     build_kit: Option<BuildKit>,
     build_variant: Option<BuildVariant>,
+    repack_variant: Option<RepackVariant>,
```
bcressey:
```suggestion
```

tools/buildsys/src/builder.rs
```diff
@@ -369,6 +422,60 @@ impl DockerBuild {
         })
     }
 
+    /// Create a new `DockerBuild` that can repackage a variant image.
+    pub(crate) fn repack_variant(args: RepackVariantArgs, manifest: &Manifest) -> Result<Self> {
+        let image_layout = manifest.info().image_layout().cloned().unwrap_or_default();
+        let ImageLayout { partition_plan, .. } = image_layout;
```
bcressey: I really think we need the expected os and data image sizes as well in order to do proper input checking.

tools/buildsys/src/builder.rs
```diff
@@ -551,6 +664,7 @@ fn create_marker_dir(
         BuildType::Package => "packages",
         BuildType::Kit => "kits",
         BuildType::Variant => "variants",
+        BuildType::Repack => "repack",
```
bcressey: merge into the previous match arm:
```suggestion
        BuildType::Repack => "variant",
```

We want the repack markers to go into the same directory that the variant markers go into, so that cleanup works consistently for repacked images.

tools/buildsys/src/builder.rs
jpculp: Sure thing. I was just waiting to add it after I wrote something to use it in `img2img`.

tools/buildsys/src/args.rs
jpculp: I was hoping it might get treated as nothing, but it seems to have had the opposite effect. The good news is by switching to `0b00001000` I'm no longer seeing the extraneous cargo directives for the sensitive env vars that I was seeing before.

jpculp:
- Removed unnecessary items from `manifest.rs`.
- Set **Repack** to the typical _"variants"_ marker dir.
- Switched **Repack** from `0b00000000` to `0b00001000`.
- Added **OS_IMAGE_SIZE_GIB** and **DATA_IMAGE_SIZE_GIB** to `repack-variant`.

jpculp:
Rebased.

tools/buildsys/src/args.rs
```diff
@@ -212,11 +235,14 @@ const VARIANT: u8 = BuildFlags::Variant as u8;
 #[test]
 fn build_type_includes_test() {
     // true
+    assert!(BuildFlags::Repack.includes(8));
```
bcressey: This isn't a useful test - it is effectively just asserting that the enum variant still has the value `0b0001000`, which would not survive a refactor.

tools/buildsys/src/args.rs
```diff
@@ -212,11 +235,14 @@ const VARIANT: u8 = BuildFlags::Variant as u8;
...
     assert!(BuildFlags::Package.includes(PACKAGE | VARIANT));
     assert!(BuildFlags::Variant.includes(VARIANT));
     assert!(BuildFlags::Variant.includes(VARIANT | PACKAGE));
 
     // false
+    assert!(!BuildFlags::Repack.includes(PACKAGE));
+    assert!(!BuildFlags::Repack.includes(VARIANT));
```
bcressey:
```suggestion
    assert!(!BuildFlags::Repack.includes(PACKAGE | VARIANT));
```

tools/buildsys/src/manifest.rs
```diff
@@ -504,6 +504,7 @@ fn is_manifest_type(pkg_metadata: &PackageMetadata, manifest_type: BuildType) ->
         BuildType::Package => metadata_table.get("build-package").is_some(),
         BuildType::Kit => metadata_table.get("build-kit").is_some(),
         BuildType::Variant => metadata_table.get("build-variant").is_some(),
+        BuildType::Repack => metadata_table.get("build-variant").is_some(),
```
bcressey: This would be better as an error or an `unreachable!` - we shouldn't be taking this graph-walk path for a repack operation, and the longer we make it work before failing, the harder it will be to diagnose the logic error that let us reach this point.

bcressey (approved):

tools/buildsys/src/args.rs
jpculp: This was to avoid `REPACK` being flagged as dead code, but I can add a `clippy` exception.

tools/buildsys/src/manifest.rs
jpculp: Since this function returns a bool, `unreachable!` might be cleanest.

jpculp:
- Adjust `build_type_includes_test`.
- Set **Repack** to unreachable in `is_manifest_type`.

tools/buildsys/src/args.rs
```diff
@@ -186,11 +207,13 @@ enum BuildFlags {
...
+    Repack = 0b0001000,
 }
 
 impl From<BuildType> for BuildFlags {
     fn from(value: BuildType) -> Self {
         match value {
+            BuildType::Repack => BuildFlags::Repack,
             BuildType::Package => BuildFlags::Package,
             BuildType::Kit => BuildFlags::Kit,
             BuildType::Variant => BuildFlags::Variant,
```
webern: Nit: make the order of the enum variants consistent.

```suggestion
            BuildType::Package => BuildFlags::Package,
            BuildType::Kit => BuildFlags::Kit,
            BuildType::Variant => BuildFlags::Variant,
            BuildType::Repack => BuildFlags::Repack,
```

tools/buildsys/src/args.rs
```diff
@@ -206,17 +229,22 @@ impl BuildFlags {
     }
 }
 
+#[allow(dead_code)]
+const REPACK: u8 = BuildFlags::Repack as u8;
+
```
webern: nit

```suggestion
const REPACK: u8 = BuildFlags::Repack as u8;
```

tools/buildsys/src/builder.rs
```diff
@@ -211,17 +223,60 @@ impl VariantBuildArgs {
     }
 }
 
+struct VariantRepackArgs {
```
webern: nit: consistent naming (albeit long)

```suggestion
struct RepackVariantBuildArgs {
```

tools/buildsys/src/manifest.rs
```diff
@@ -504,6 +504,7 @@ fn is_manifest_type(pkg_metadata: &PackageMetadata, manifest_type: BuildType) ->
         BuildType::Package => metadata_table.get("build-package").is_some(),
         BuildType::Kit => metadata_table.get("build-kit").is_some(),
         BuildType::Variant => metadata_table.get("build-variant").is_some(),
+        BuildType::Repack => unreachable!("Repacking is not defined in manifests"),
```
webern, line 507: I don't like this. Let's just return false since that is an accurate answer to the question `is_manifest_type`. A function that can panic seems out of character for us.

Edit: I have learned that this was Ben's preference so go ahead and merge it like this. I hope to make this function go away at some point with some refactoring, but we'll see...

```suggestion
        BuildType::Repack => false,
```

twoliter/src/tools.rs
```diff
@@ -104,15 +104,19 @@ async fn test_install_tools() {
...
     assert!(toolsdir.join("docker-go").is_file());
+    assert!(toolsdir.join("img2img").is_file());
     assert!(toolsdir.join("imghelper").is_file());
+    assert!(toolsdir.join("metadata.spec").is_file());
     assert!(toolsdir.join("partyplanner").is_file());
+    assert!(toolsdir.join("repack.Dockerfile").is_file());
+    assert!(toolsdir.join("repack.Dockerfile.dockerignore").is_file());
     assert!(toolsdir.join("rpm2img").is_file());
     assert!(toolsdir.join("rpm2kmodkit").is_file());
     assert!(toolsdir.join("rpm2migrations").is_file());
```
webern, line 119: nit: it was sorted before

```suggestion
    assert!(toolsdir.join("build.Dockerfile").is_file());
    assert!(toolsdir.join("build.Dockerfile.dockerignore").is_file());
    assert!(toolsdir.join("docker-go").is_file());
    assert!(toolsdir.join("img2img").is_file());
    assert!(toolsdir.join("imghelper").is_file());
    assert!(toolsdir.join("Makefile.toml").is_file());
    assert!(toolsdir.join("metadata.spec").is_file());
    assert!(toolsdir.join("partyplanner").is_file());
    assert!(toolsdir.join("repack.Dockerfile").is_file());
    assert!(toolsdir.join("repack.Dockerfile.dockerignore").is_file());
    assert!(toolsdir.join("rpm2img").is_file());
    assert!(toolsdir.join("rpm2kmodkit").is_file());
    assert!(toolsdir.join("rpm2migrations").is_file());
```

webern (approved):

twoliter/src/tools.rs
jpculp: `Makefile.toml` is at the top due to the capitalization.

jpculp:
Adjusted some names and ordering.

webern (approved):

bcressey (approved):

PR #217, opened 2024-05-15
add image prep functions to imghelper

jpculp:
**Description of changes:**

Adds or moves functionality around image name generation, compression, and decompression to `imghelper`.

**Testing done:**

- Built `aws-k8s-1.28`
- Repacked `aws-k8s-1.28`
- Build `vmware-k8s-1.28`
- Repacked `vmware-k8s-1.28`

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/imghelper
+++ b/twoliter/embedded/imghelper
@@ -1,4 +1,20 @@
 #!/usr/bin/env bash
+# shellcheck disable=SC2034 # some variables are used through parameter substitution
+
+# shellcheck disable=SC2154  # some variables are defined externally by the environment
+FILENAME_PREFIX="${IMAGE_NAME:?}-${VARIANT:?}-${ARCH:?}-${VERSION_ID:?}-${BUILD_ID:?}"
+SYMLINK_PREFIX="${IMAGE_NAME:?}-${VARIANT:?}-${ARCH:?}"
+
+DATA_IMAGE_SUFFIX="-data"
+BOOT_IMAGE_SUFFIX="-boot"
+VERITY_IMAGE_SUFFIX="-root"
+ROOT_IMAGE_SUFFIX="-root"
+
+OS_IMAGE_NAME="${FILENAME_PREFIX}"
+DATA_IMAGE_NAME="${FILENAME_PREFIX}${DATA_IMAGE_SUFFIX}"
+BOOT_IMAGE_NAME="${FILENAME_PREFIX}${BOOT_IMAGE_SUFFIX}"
+VERITY_IMAGE_NAME="${FILENAME_PREFIX}${VERITY_IMAGE_SUFFIX}"
+ROOT_IMAGE_NAME="${FILENAME_PREFIX}${ROOT_IMAGE_SUFFIX}"
 
 sanity_checks() {
   local output_fmt partition_plan ovf_template uefi_secure_boot
@@ -52,24 +68,52 @@ sanity_checks() {
   fi
 }
 
-# shellcheck disable=SC2034 # variables are used through parameter substitution
+compress_image() {
+  local ext what output_dir input_image image_name
+  ext="${1:?}"
+  what="${2:?}"
+  output_dir="${3:?}"
+  input_image="${what^^}"
+  image_name="${what^^}_NAME"
+
+  case "${ext}" in
+  *lz4)
+    lz4 -9vc "${!input_image}" >"${output_dir}/${!image_name}${ext:+.${ext}}"
+    ;;
+  qcow2)
+    qemu-img convert -f raw -O "${ext}" \
+      "${!input_image}" "${output_dir}/${!image_name}${ext:+.${ext}}"
+    ;;
+  # stream optimization is required for creating Open Virtual Appliances (OVAs)
+  vmdk)
+    qemu-img convert -f raw -O "${ext}" -o subformat=streamOptimized \
+      "${!input_image}" "${output_dir}/${!image_name}${ext:+.${ext}}"
+    ;;
+  *)
+    echo "unexpected extension: ${ext}" >&2
+    exit 1
+    ;;
+  esac
+}
+
 symlink_image() {
-  local ext what target symlink_prefix version_id output_dir symlink_suffix
+  local ext what target output_dir symlink_prefix version_id symlink_suffix
   ext="$1"
   what="${2:?}"
-  symlink_prefix="${3:?}"
-  version_id="${4:?}"
-  output_dir="${5:?}"
+  output_dir="${3:?}"
+
+  symlink_prefix="${SYMLINK_PREFIX}"
+  version_id="${VERSION_ID:?}"
 
   ext="${ext:+.$ext}"
   target="${what^^}_NAME"
 
   case "${what}" in
   os_image) symlink_suffix="" ;;
-  data_image) symlink_suffix="-data" ;;
-  boot_image) symlink_suffix="-boot.ext4.lz4" ;;
-  verity_image) symlink_suffix="-root.verity.lz4" ;;
-  root_image) symlink_suffix="-root.ext4.lz4" ;;
+  data_image) symlink_suffix="${DATA_IMAGE_SUFFIX}" ;;
+  boot_image) symlink_suffix="${BOOT_IMAGE_SUFFIX}" ;;
+  verity_image) symlink_suffix="${VERITY_IMAGE_SUFFIX}" ;;
+  root_image) symlink_suffix="${ROOT_IMAGE_SUFFIX}" ;;
   *)
     echo "unknown image '${what}'" >&2
     exit 1
--- a/twoliter/embedded/imghelper
+++ b/twoliter/embedded/imghelper
@@ -68,6 +68,34 @@ sanity_checks() {
   fi
 }
 
+decompress_image() {
+  local ext what input_dir input_image image_name
+  ext="${1:?}"
+  what="${2:?}"
+  input_dir="${3:?}"
+
+  input_image="${what^^}"
+  image_name="${what^^}_NAME"
+
+  case "${ext}" in
+  *lz4)
+    unlz4 -f "${input_dir}/${!image_name}${ext:+.${ext}}" "${!image_name}.img"
+    ;;
+  qcow2)
+    qemu-img convert -f "${ext}" -O raw \
+      "${input_dir}/${!image_name}${ext:+.${ext}}" "${!image_name}.img"
+    ;;
+  vmdk)
+    qemu-img convert -f "${ext}" -O raw \
+      "${input_dir}/${!image_name}${ext:+.${ext}}" "${!image_name}.img"
+    ;;
+  *)
+    echo "unexpected extension: ${ext}" >&2
+    exit 1
+    ;;
+  esac
+}
+
 compress_image() {
   local ext what output_dir input_image image_name
   ext="${1:?}"
--- a/twoliter/embedded/imghelper
+++ b/twoliter/embedded/imghelper
@@ -1,5 +1,57 @@
 #!/usr/bin/env bash
 
+sanity_checks() {
+  local output_fmt partition_plan ovf_template uefi_secure_boot
+  output_fmt="${1:?}"
+  partition_plan="${2:?}"
+  ovf_template="${3:?}"
+  uefi_secure_boot="${4:?}"
+  case "${output_fmt}" in
+  raw | qcow2 | vmdk) ;;
+  *)
+    echo "unexpected image output format '${output_fmt}'" >&2
+    exit 1
+    ;;
+  esac
+
+  case "${partition_plan}" in
+  split | unified) ;;
+  *)
+    echo "unexpected partition plan '${partition_plan}'" >&2
+    exit 1
+    ;;
+  esac
+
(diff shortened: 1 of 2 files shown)


twoliter/embedded/imghelper
```diff
@@ -1,6 +1,150 @@
...
+    if [ "${partition_plan}" == "split" ]; then
+      if ! grep -Fq '{{DATA_DISK}}' "${ovf_template}"; then
+        echo "Missing data disk in OVF template, which is required for 'split' layout." >&2
+        exit 1
+      fi
+    fi
+
+    if [ "${partition_plan}" == "unified" ]; then
+      if grep -Fq '{{DATA_DISK}}' "${ovf_template}"; then
+        echo "Incorrect data disk in OVF template, which is not supported for 'unified' layout." >&2
```
webern: ?

```suggestion
        echo "Data disk included in OVF template, which is not supported for 'unified' layout." >&2
```

webern (approved):

twoliter/embedded/imghelper
```diff
@@ -1,6 +1,150 @@
...
+  local ext what output_dir input_image image_name
+  ext="$1"
+  what="${2:?}"
+  output_dir="${3:?}"
+  input_image="${what^^}"
+  image_name="${what^^}_NAME"
+
+  if [[ "${what}" == "os_image" ]] || [[ "${what}" == "data_image" ]]; then
+    if [[ "${ext}" == "img.lz4" ]]; then
+      lz4 -vc "${!input_image}" >"${output_dir}/${!image_name}.img.lz4"
```
bcressey: It looks like this would be more readable with a couple changes:
* always append `${ext}` to `${!image_name}` when writing the output (this would work for boot/root/hash also where the empty string is passed)
* always use `lz4 -9vc` for "img.lz4" - not sure if this really saves much time for the os image; it's important for the boot/root/hash images to stay small.

Then this section can be just a case statement:

```
case "${ext}" in
  ""|.img.lz4)
    lz4 -9vc "${!input_image}" \
      > "${output_dir}/${!image_name}${ext:+.${ext}}"
    ;;
  qcow2)
    ...
  vmdk)
    ...
  *)
    echo "unexpected extension: ${ext}" >&2
    exit 1
    ;;
esac
```

twoliter/embedded/imghelper
```diff
@@ -1,6 +1,150 @@
...
+    unlz4 -f "${input_dir}/${filename_prefix}.img.lz4" "${output_image}"
+  elif [[ "${fmt}" == "qcow2" ]]; then
+    qemu-img convert -f qcow2 -O raw \
+      "${input_dir}/${filename_prefix}.qcow2" "${output_image}"
+  fi
+}
+
+decompress_ova() {
+  local fmt input_ova filename_prefix output_image
+  fmt="${1:?}"
```
bcressey: It feels like for consistency with `compress_image`, this should be a supported extension in `decompress_image` rather than a separate function.

twoliter/embedded/imghelper
```diff
@@ -1,6 +1,150 @@
...
+  input_ova="${2:?}"
+  filename_prefix="${3:?}"
+  output_image="${4:?}"
+
+  tar xvf "${input_ova}" "${filename_prefix}.vmdk"
+  qemu-img convert -f vmdk -O raw "${filename_prefix}.vmdk" "${output_image}"
+  rm "${filename_prefix}.vmdk"
```
bcressey: With a split VMDK, we might end up with two images to extract.

If you take an output directory, rather than a filename, this becomes easier to do - just extract files matching `*.vmdk` from the OVA. It seems like the function caller already knows what images to expect (because of `set_image_names`).

twoliter/embedded/imghelper
```diff
@@ -1,6 +1,150 @@
 #!/usr/bin/env bash
-
 # shellcheck disable=SC2034 # variables are used through parameter substitution
+
+DATA_IMAGE_SUFFIX="-data"
+BOOT_IMAGE_SUFFIX="-boot.ext4.lz4"
+VERITY_IMAGE_SUFFIX="-root.verity.lz4"
+ROOT_IMAGE_SUFFIX="-root.ext4.lz4"
```
bcressey: It's weird that the data image variable doesn't have an extension but the other three do. It suggests they're not really being treated the same way and probably shouldn't share a `_SUFFIX` suffix.

twoliter/embedded/imghelper
```diff
@@ -1,6 +1,150 @@
...
+  filename_prefix="${3:?}"
+
+  if [[ "${fmt}" == "raw" ]]; then
+    echo "${input_dir}/${filename_prefix}${DATA_IMAGE_SUFFIX}.img.lz4"
+  elif [[ "${fmt}" == "qcow2" ]]; then
+    echo "${input_dir}/${filename_prefix}${DATA_IMAGE_SUFFIX}.qcow2"
+  elif [[ "${fmt}" == "vmdk" ]]; then
+    echo "${input_dir}/${filename_prefix}${DATA_IMAGE_SUFFIX}.vmdk"
+  fi
+}
```
bcressey: Without seeing the caller for this function, it's hard to review but it does seem a bit strange. The caller has almost all of the information (because `${filename_prefix}${DATA_IMAGE_SUFFIX}` is the same as the `data_image_name` we set for them.

twoliter/embedded/imghelper
```diff
@@ -1,6 +1,150 @@
...
+
+DATA_IMAGE_SUFFIX="-data"
+BOOT_IMAGE_SUFFIX="-boot.ext4.lz4"
+VERITY_IMAGE_SUFFIX="-root.verity.lz4"
+ROOT_IMAGE_SUFFIX="-root.ext4.lz4"
+
+set_image_names() {
+  local filename_prefix
+  local -n os_image_name data_image_name
+  local -n boot_image_name verity_image_name root_image_name
```
bcressey: I don't see why the caller needs to have these three names (boot/root/hash). `rpm2img` never interacts with files by these names because they're only created when compressed.

Also, `rpm2img` only needs `os_image_name` and `data_image_name` because the OVA creation change hasn't merged yet.

I would say, try to merge that one first, rebase, and remove the `set_image_names` and `get_data_image` functions until the caller is added, so I have more context to review.

twoliter/embedded/imghelper
jpculp: Another weird behavior inherited from `rpm2img`. It was probably written this way since the extension for data depends on the build args whereas the others are constant. We can drop them and add the extensions to the `symlink_image` calls, it just means a bigger delta from the original script.

twoliter/embedded/imghelper
jpculp: Once everything is moved over, only `imghelper` will use them, but it felt kind of wrong to have `FILENAME_PREFIX="${IMAGE_NAME}-${VARIANT}-${ARCH}-${VERSION_ID}-${BUILD_ID}"` there. These values come from the docker environment calling `rpm2img` or `img2img`, but if we're ok with FILENAME_PREFIX (and the similar SYMLINK_PREFIX) being set at the top of `imghelper` it would make the code cleaner on both sides.

twoliter/embedded/imghelper
jpculp: I could save this for the `img2img` PR, but weirdness comes from our choice of formats. "raw" == "img.lz4", so we can't just do `${INPUT_DIR}/${FILENAME_PREFIX}${DATA_IMAGE_SUFFIX}.${OUTPUT_FMT}`. I could do `${INPUT_DIR}/${FILENAME_PREFIX}${DATA_IMAGE_SUFFIX}.${OUTPUT_FMT/raw/img.lz4}`, but ideally filename prefix and data image suffix will be in imghelper and not the caller. This seemed the most readable when factoring in the above.

twoliter/embedded/imghelper
jpculp: I originally split this up because the we vend OVAs with the "friendly" name as opposed to the "versioned" name for our raw images. However, since we're going to generate missing symlinks in the end anyway, I can just create a symlink for the ova at the start.

jpculp:
- Split extensions from ***_SUFFIX**.
- Adjusted wording on OVF error.
- Cleaned up `compress_image`.
- Unified `decompress_image`.
- Dropped `get_data_image` for a later PR.

twoliter/embedded/imghelper
```diff
@@ -1,6 +1,140 @@
...
+    ;;
+  qcow2)
+    qemu-img convert -f "${ext}" -O raw \
+      "${input_dir}/${filename_prefix}${ext:+.${ext}}" "${output_image}"
+    ;;
+  vmdk)
+    tar xvf "${filename_prefix}.ova" "*${ext:+.${ext}}"
+    qemu-img convert -f "${ext}" -O raw "${filename_prefix}.vmdk" "${output_image}"
+    rm "${filename_prefix}${ext:+.${ext}}" "${filename_prefix}.ova"
+    ls -1
```
bcressey: omit debugging commands:
```suggestion
```

twoliter/embedded/imghelper
```diff
@@ -1,6 +1,140 @@
...
+    ext="img.lz4"
+    unlz4 -f "${input_dir}/${filename_prefix}${ext:+.${ext}}" "${output_image}"
+    lz4 -9vc "${!input_image}" >"${output_dir}/${!image_name}${ext:+.${ext}}"
+    ;;
+  qcow2)
+    qemu-img convert -f "${ext}" -O raw \
+      "${input_dir}/${filename_prefix}${ext:+.${ext}}" "${output_image}"
+    ;;
+  vmdk)
+    tar xvf "${filename_prefix}.ova" "*${ext:+.${ext}}"
```
bcressey: Why is the `*` needed?

twoliter/embedded/imghelper
```diff
@@ -1,6 +1,140 @@
...
+    qemu-img convert -f "${ext}" -O raw "${filename_prefix}.vmdk" "${output_image}"
+    rm "${filename_prefix}${ext:+.${ext}}" "${filename_prefix}.ova"
```
bcressey: This won't handle split image layouts correctly - you are only extracting one file then removing the archive.

twoliter/embedded/imghelper
```diff
@@ -1,6 +1,140 @@
...
+      if ! grep -Fq '{{DB_CERT_DER_HEX}}' "${ovf_template}"; then
+        echo "Missing CA certificate field in OVF template, which is required for Secure Boot support." >&2
+        exit 1
+      fi
+    fi
+  fi
+}
+
+decompress_image() {
+  local fmt input_dir filename_prefix output_image ext
```
bcressey: I still think you need to align this with the `compress_image` signature.

Compress:
```
compress_image "vmdk" "os_image" "${OUTPUT_DIR}"
symlink_image "vmdk" "os_image" "${SYMLINK_PREFIX}" "${VERSION_ID}" "${OUTPUT_DIR}"
if [ -s "${DATA_IMAGE}" ] ; then
  compress_image "vmdk" "data_image" "${OUTPUT_DIR}"
  symlink_image "vmdk" "data_image" "${SYMLINK_PREFIX}" "${VERSION_ID}" "${OUTPUT_DIR}"
fi
```

Decompress:
```
decompress_image "vmdk" "os_image" "${OUTPUT_DIR}"
symlink_image "vmdk" "os_image" "${SYMLINK_PREFIX}" "${VERSION_ID}" "${OUTPUT_DIR}"
if [ -s "${DATA_IMAGE}" ] ; then
  decompress_image "vmdk" "data_image" "${OUTPUT_DIR}"
  symlink_image "vmdk" "data_image" "${SYMLINK_PREFIX}" "${VERSION_ID}" "${OUTPUT_DIR}"
fi
```

For that to work, you will want to leave the compressed artifact alone, so that the second call can extract the data image for the split VMDK case. But for consistency you should also not remove it in the raw or qcow2 cases.

twoliter/embedded/imghelper
bcressey: > if we're ok with FILENAME_PREFIX (and the similar SYMLINK_PREFIX) being set at the top of `imghelper` it would make the code cleaner on both sides.

OK, let's try it like that.

twoliter/embedded/imghelper
jpculp: This will extract multiple vmdks in the image. We can also just extract the whole tarball though.

twoliter/embedded/imghelper
jpculp: We're decompressing only the OS vmdk, but extracting all with `*${ext:+.${ext}}` on L98.

twoliter/embedded/imghelper
jpculp: This piece will be moved out of the decompress function, but for clarity it will be `"*.vmdk"` in `img2img`.

jpculp:
- Move **FILENAME_PREFIX**, **SYMLINK_PREFIX**, and the **_IMAGE_NAME**s to `imghelper`
- Limit scope of `decompress_image` to make it more consistent with `compress_image`
- Remove `VERSION_ID` as an arg for `symlink_image` since we're already assuming it's in the environment and to make the inputs more consistent with `compress_image` and `decompress_image`.

jpculp:
Rebased.

bcressey (approved):

PR #219, opened 2024-05-15
rpm2img: move OVA generation to imghelper

jpculp:
**Description of changes:**

Relocates code-block that created the OVA to `imghelper`. Small changes include localizing of variables and adding `|| exit 1` to `pushd`/`popd` since `imghelper` doesn't assume `set -e` is set in the calling script. 

**Testing done:**

- Built and smoke tested `vmware-k8s-1.28`.

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/imghelper
+++ b/twoliter/embedded/imghelper
@@ -107,3 +107,89 @@ mkfs_data_xfs() {
   dd if=/dev/zero of="${bottlerocket_data}" bs=1M count="${size%?}"
   dd if="${bottlerocket_data}" of="${target}" conv=notrunc bs=1M seek="${offset}"
 }
+
+generate_ova() {
+  local os_vmdk data_vmdk
+  os_vmdk="${1:?}"
+  data_vmdk="${2:?}"
+
+  local os_image_publish_size_gib data_image_publish_size_gib
+  os_image_publish_size_gib="${3:?}"
+  data_image_publish_size_gib="${4:?}"
+
+  local ovf_template uefi_secure_boot sbkeys output_dir
+  ovf_template="${5:?}"
+  uefi_secure_boot="${6:?}"
+  sbkeys="${7:?}"
+  output_dir="${8:?}"
+
+  local ova_dir
+  ova_dir="$(mktemp -d)"
+
+  local file_prefix="${os_vmdk%.vmdk}"
+  local ovf="${file_prefix}.ovf"
+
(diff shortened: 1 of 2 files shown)


twoliter/embedded/imghelper
```diff
@@ -107,3 +107,89 @@ mkfs_data_xfs() {
...
+  data_vmdk="${2:?}"
+
+  local os_image_publish_size_gib data_image_publish_size_gib
+  os_image_publish_size_gib="${3:?}"
+  data_image_publish_size_gib="${4:?}"
+
+  local ovf_template uefi_secure_boot sbkeys output_dir
+  ovf_template="${5:?}"
+  uefi_secure_boot="${6:?}"
+  sbkeys="${7:?}"
```
jpculp, line 123: **sbkeys** will eventually be a constant in `imghelper`, but I'll switch this when the secure boot code moves over.

twoliter/embedded/imghelper
```diff
@@ -107,3 +107,89 @@ mkfs_data_xfs() {
...
+    -e "s/{{DATA_DISK_BYTES}}/${data_disk_bytes}/g" \
+    >"${ova_dir}/${ovf}"
+
+  # The manifest templates for Secure Boot expect the cert data for
+  # PK, KEK, db, and dbx.
+  if [[ "${uefi_secure_boot}" == "yes" ]]; then
+    local data_disk_bytes kek_cert_der_hex db_cert_der_hex dbx_empty_hash_hex
+    pk_cert_der_hex="$(hexdump -ve '1/1 "%02x"' "${sbkeys}/PK.cer")"
+    kek_cert_der_hex="$(hexdump -ve '1/1 "%02x"' "${sbkeys}/KEK.cer")"
+    db_cert_der_hex="$(hexdump -ve '1/1 "%02x"' "${sbkeys}/db.cer")"
```
rpkelly, line 150: Should we check that these exist first?

jpculp: These files will always be created by the prerequisite `build-sbkeys` target, but even if they weren't the `set -e` would catch it as the local variable is declared prior to assignment. If we wanted to check regardless, I would rather it be part of the sanity checks towards the beginning of `rpm2img` since these files are referenced in a couple different places.

bcressey (approved):

rpkelly (approved):

webern (approved):

PR #220, opened 2024-05-15
buildsys: remove variant sensitivity

### LINKED ISSUE #193: remove build-package dependency on variant
Some of this will take place in Bottlerocket's packages, the rest will be here in Twoliter.

Currently buildsys build-package needs to read the variant Cargo.toml and package builds proceed differently based on what image features are found. We need to stop this. Each package should be built the same way regardless of what variant is being built.

In concert with https://github.com/bottlerocket-os/bottlerocket/issues/3883

These are the Bottlerocket packages affected:

```
packages/kernel-5.10/Cargo.toml
9:variant-sensitive = "platform"

packages/systemd/Cargo.toml
13:package-features = ["unified-cgroup-hierarchy"]

packages/kernel-5.15/Cargo.toml
9:variant-sensitive = "platform"

packages/os/Cargo.toml
9:variant-sensitive = true
27:package-features = [ "systemd-networkd" ]

packages/kernel-6.1/Cargo.toml
9:variant-sensitive = "platform"
```

webern:
**Issue number:**

Closes #193
Blocks #70

**Description of changes:**

```
buildsys: remove variant sensitivity documentation

buildsys: deprecate build-package variant sensitivity

    Packages can no longer depend on knowing what variant they are being
    built for, so we raise an error if they declare variant-sensitivity or
    package-features in their Cargo.toml.

    Bottlerocket is not quite ready for this, though. So we place this
    change behind a feature flag variable:
    BUILDSYS_DEPRECATED_FEATURE_VARIANT_SENSITIVITY

buildsys: build-package refactor variant sensitivity

    Move the variant sensitivity logic of build-package into a function that
    can soon be deleted.
```

**Testing done:**

- [x] Built and ran an aws-ecs-2 variant
- [x] Without cleaning: Built and ran an aws-k8s-1.27 variant, and verified that apiclient could get the kubernetes.cluster-name value (which does not exist on aws-ecs-2 and proves conditional compilation of the model took place)
- [x] Build all Bottlerocket variants using Bottlerocket CI.
  - [x] For this I used the following Bottlerocket diff. Note that `os` is the only package relying on `variant-sensitive`! https://github.com/bottlerocket-os/bottlerocket/compare/3243428cde2040f75ef871bbd742bc7ee00d6776...webern:be9d22dd6950fee46683ab5d2e0f1cb30e0f38ed
  - [x] Here is the Bottlerocket actions run in which all variants were built https://github.com/bottlerocket-os/bottlerocket/actions/runs/9135774525

**Terms of contribution:**

### DIFF
--- a/tools/buildsys/src/manifest.rs
+++ b/tools/buildsys/src/manifest.rs
@@ -80,43 +80,6 @@ naming.
 package-name = "better.name"
 ```
 
-`variant-sensitive` lets you specify whether the package should be rebuilt when
-building a new variant, and defaults to false; set it to true if a package is
-using the variant to affect its build process.
-
-```ignore
-[package.metadata.build-package]
-variant-sensitive = true
-```
-
-Some packages might only be sensitive to certain components of the variant
-tuple, such as the platform, runtime, or family. The `variant-sensitive` field
-can also take a string to indicate the source of the sensitivity.
-
-```ignore
-[package.metadata.build-package]
-# sensitive to platform, like "metal" or "aws"
-variant-sensitive = "platform"
-
-# sensitive to runtime, like "k8s" or "ecs"
-variant-sensitive = "runtime"
-
-# sensitive to family, like "metal-k8s" or "aws-ecs"
-variant-sensitive = "family"
-```
-
-`package-features` is a list of image features that the package tracks. This is
-useful when the way the package is built changes based on whether a particular
-image feature is enabled for the current variant, rather than when the variant
-tuple changes.
-
-```ignore
-[package.metadata.build-package]
-package-features = [
-    "grub-set-private-var",
-]
-```
-
 `releases-url` is ignored by buildsys, but can be used by packager maintainers
 to indicate a good URL for checking whether the software has had a new release.
 ```ignore
--- a/tools/buildsys/src/main.rs
+++ b/tools/buildsys/src/main.rs
@@ -17,14 +17,15 @@ mod spec;
 
 use crate::args::{BuildPackageArgs, BuildVariantArgs, Buildsys, Command, RepackVariantArgs};
 use crate::builder::DockerBuild;
-use buildsys::manifest::{BundleModule, Manifest, ManifestInfo, SupportedArch};
+use buildsys::manifest::{BundleModule, ImageFeature, Manifest, ManifestInfo, SupportedArch};
 use cache::LookasideCache;
 use clap::Parser;
 use gomod::GoMod;
 use project::ProjectInfo;
 use snafu::{ensure, ResultExt};
 use spec::SpecInfo;
-use std::path::PathBuf;
+use std::collections::HashSet;
+use std::path::{Path, PathBuf};
 use std::process;
 
 mod error {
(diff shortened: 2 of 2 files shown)


tools/buildsys/src/manifest.rs
```diff
@@ -80,43 +80,6 @@ naming.
 package-name = "better.name"
 ```
 
-`variant-sensitive` lets you specify whether the package should be rebuilt when
-building a new variant, and defaults to false; set it to true if a package is
-using the variant to affect its build process.
-
-```ignore
-[package.metadata.build-package]
-variant-sensitive = true
```
bcressey, line 89: We'll need at least the boolean form of variant sensitivity to remain until none of the programs in `os` are conditionally compiled by way of their build script.

Otherwise what will happen is we'll build all those packages once, with the active variant from the environment, and we won't rebuild them again when the variant changes. So everything that still builds in a settings model will only be built for the first variant.

webern: Check my latest diff. I added a "DEPRECATED" feature flag. As it turns out, `os` is the only remaining package using `variant-sensitive`! So in the `os` `build.rs` I am emitting a `DEPRECATED` feature flag to sneak us by for a while longer while ensuring that nobody else uses `variant-sensitive`.

tools/buildsys/src/builder.rs
```diff
@@ -156,10 +150,6 @@ impl PackageBuildArgs {
         args.build_arg("VARIANT_FLAVOR", &self.variant_flavor);
         args.build_arg("VARIANT_PLATFORM", &self.variant_platform);
         args.build_arg("VARIANT_RUNTIME", &self.variant_runtime);
-        for image_feature in &self.image_features {
-            args.build_arg(format!("{}", image_feature), "1");
-        }
```
bcressey: We need to finish the conditional compilation removal in `os.spec` before it's safe to do this.

The image features serve two roles:
- Outward-facing, to tell cargo when to rebuild.
- Inward-facing, where they control the bconds we generate so the spec file can include or exclude certain programs.

If we stop passing in these arguments then the bconds generated in the "rpm-macros-and-bconds" stage of the Dockerfile will be incorrect for package builds.

We really need to clean up https://github.com/bottlerocket-os/bottlerocket/pull/3828 and get it merged before we can land this change.

Note that the first commit in that PR removes the bconds, but we can't take that in isolation because pluto still can't be built for a non-k8s model until the following patch.

webern: @bcressey check the latest push. All functionality is retained but gated behind the `BUILDSYS_DEPRECATED_FEATURE` flag.

webern:
@bcressey this is tested and safe. All functionality is retained behind a feature flag.

bcressey (approved):

jmt-lab (approved):
LGTM

webern:
> https://github.com/bottlerocket-os/twoliter/compare/9da417bf68654303b2189f2466209b091d735ab0..f11b56c2a7b186a982745fc0fb2bf7fc658ed997

Rebase

PR #221, opened 2024-05-16
move secure boot setup to imghelper

jpculp:
**Description of changes:**

Moves code-blocks related to secure boot to `imghelper`, with some minor adjustments for portability.

**Testing done:**

- Built and smoke tested `aws-k8s-1.28`

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/imghelper
+++ b/twoliter/embedded/imghelper
@@ -479,6 +479,31 @@ provide_certs() {
   mcopy -i "${efi_image}" "${SBKEYS}"/db.{crt,cer} ::/EFI/BOOT
 }
 
+sign_vmlinuz() {
+  local vmlinuz
+  vmlinuz="${1:?}"
+
+  undo_sign "${vmlinuz}"
+  do_sign "${vmlinuz}" "${SBKEYS}/vendor.cer" CODE_SIGN_KEY
+}
+
+generate_hmac() {
+  local vmlinuz
+  vmlinuz="${1:?}"
+  openssl sha512 -hmac FIPS-FTW-RHT2009 -hex "${vmlinuz}" |
+    awk -v vmlinuz="${vmlinuz}" '{ print $2 "  " vmlinuz }' \
+      >"${vmlinuz%/*}.${vmlinuz##*/}.hmac"
+}
+
+sign_grubcfg() {
+  local grub_cfg grub_cfg_sig
+  grub_cfg="${1:?}"
+  grub_cfg_sig="${grub_cfg}.sig"
+  [[ -f "${grub_cfg_sig}" ]] && rm "${grub_cfg_sig}"
+  gpg --batch --no-tty --detach-sign "${grub_cfg}"
+  gpg --batch --no-tty --verify "${grub_cfg_sig}" "${grub_cfg}"
+}
+
 generate_ova() {
   local os_vmdk data_vmdk
   os_vmdk="${1:?}"
--- a/twoliter/embedded/imghelper
+++ b/twoliter/embedded/imghelper
@@ -23,6 +23,11 @@ VERITY_HASH_ALGORITHM=sha256
 VERITY_DATA_BLOCK_SIZE=4096
 VERITY_HASH_BLOCK_SIZE=4096
 
+SBKEYS="${HOME}/sbkeys"
+
+# Pre-emptively declare global arrays to be populated later.
+declare -a SHIM_SIGN_KEY CODE_SIGN_KEY
+
 sanity_checks() {
   local output_fmt partition_plan ovf_template uefi_secure_boot
   output_fmt="${1:?}"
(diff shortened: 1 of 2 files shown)


twoliter/embedded/imghelper
```diff
@@ -108,6 +110,214 @@ mkfs_data_xfs() {
...
+      >"${vmlinuz%/*}.${vmlinuz##*/}.hmac"
+}
+
+sign_grub() {
+  local grub_cfg grub_cfg_sig
+  grub_cfg="${1:?}"
+  grub_cfg_sig="${grub_cfg}.sig"
+  [[ -f "${grub_cfg_sig}" ]] && rm "${grub_cfg_sig}"
+  gpg --batch --no-tty --detach-sign "${grub_cfg}"
+  gpg --batch --no-tty --verify "${grub_cfg_sig}" "${grub_cfg}"
```
jpculp, line 504: The reason for the change in syntax here is when `--batch` is not used `gpg` can make assumptions about the matching data files from a single argument. This is discouraged and instead we should always specify both explicitly. Since we're not interfacing with `gpg` interactively anyway, I just went ahead and set all the `gpg` commands to `--batch --no-tty`, which results in more useful errors when something goes wrong.

twoliter/embedded/imghelper
```diff
@@ -108,6 +110,214 @@ mkfs_data_xfs() {
...
+  "${arch}-objdumpcopy" "${grub}" \
+    --file-alignment 4096 \
+    --remove-section ".pubkey"
+}
+
+uefi_copycert() {
+  local efi_image
+  efi_image="${1:?}"
+  mcopy -i "${efi_image}" "${SBKEYS}"/db.{crt,cer} ::/EFI/BOOT
+}
```
jpculp: This feels a tiny bit silly, but by moving it out of `rpm2img` we no longer need **SBKEYS** outside of `imghelper`.

twoliter/embedded/rpm2img
```diff
@@ -251,126 +249,19 @@ mv "${ROOT_MOUNT}/boot/efi"/* "${EFI_MOUNT}"
...
-  gpg --import "${SBKEYS}/config-sign.key"
-  if [ "${SB_KEY_SOURCE}" == "aws" ] ; then
-    gpg --card-status
+    sbsetup_local_profile shim_sign_key code_sign_key
   fi
-  gpg --export > "${SBKEYS}/config-sign.pubkey"
-  gpg --list-keys
 fi
 
 pushd "${EFI_MOUNT}/EFI/BOOT" >/dev/null
```
jpculp, line 192:
```suggestion

```
I'll move this into the functions.

jpculp:
- Removed assumption that the caller will change directory to `${EFI_MOUNT}/EFI/BOOT` before calling UEFI helpers.
- Removed circular references.

twoliter/embedded/imghelper
```diff
@@ -108,6 +110,228 @@ mkfs_data_xfs() {
...
+  local sb_key_source
+  sb_key_source="${1:?}"
+
+  # Convert certificates from PEM format (ASCII) to DER (binary). This could be
+  # done when the certificates are created, but the resulting binary files are
+  # not as nice to store in source control.
+  for cert in PK KEK db vendor; do
+    openssl x509 \
+      -inform PEM -in "${SBKEYS}/${cert}.crt" \
+      -outform DER -out "${SBKEYS}/${cert}.cer"
```
rpkelly, line 314: Should we be validating that this output is equal to the input somehow, or is it fine to just trust this conversion happened correctly?

twoliter/embedded/imghelper
```diff
@@ -108,6 +110,228 @@ mkfs_data_xfs() {
...
+}
+
+sbsetup_aws_profile() {
+  local -n shim_sign_key code_sign_key
+  shim_sign_key="${1:?}"
+  code_sign_key="${2:?}"
+
+  # Set AWS environment variables from build secrets, if present.
+  local var val
+  for var in AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN; do
```
rpkelly, line 330: Does `AWS_DEFAULT_REGION` also need to be in this list?

twoliter/embedded/imghelper
```diff
@@ -108,6 +110,228 @@ mkfs_data_xfs() {
...
+
+  certutil -d "${pedb}" -A -n vendor -i "${SBKEYS}/vendor.crt" -t ",,C"
+  certutil -d "${pedb}" -A -n code-sign-key -i "${SBKEYS}/code-sign.crt" -t ",,P"
+  pk12util -d "${pedb}" -i "${SBKEYS}/code-sign.p12" -W ""
+
+  certutil -d "${pedb}" -L
+  shim_sign_key=(-c shim-sign-key)
+  code_sign_key=(-c code-sign-key)
+
+  sbsetup_wrapup local
```
rpkelly: I spent too long trying to figure out how `local` was working here. Maybe just a me thing, but, a comment would be nice :)

twoliter/embedded/imghelper
```diff
@@ -108,6 +110,228 @@ mkfs_data_xfs() {
...
+  else
+    pesign -i "${shim}" -o "${shim}.unsigned" -u 0 -r
+  fi
+  "${arch}-objdumpcopy" "${shim}.unsigned" \
+    --update-section ".vendor_cert=${SBKEYS}/vendor.obj"
+  pesign -i "${shim}.unsigned" -o "${shim}.signed" -s "${shim_sign_key[@]}"
+  mv "${shim}.signed" "${shim}"
+  rm "${shim}.unsigned"
+  pesign -i "${shim}" -l
+  pesigcheck -i "${shim}" -n 0 -c "${SBKEYS}/db.cer"
```
rpkelly: This block of code is repeated at least 3 times (possibly a 4th in the `vmlinuz_sbsetup` function). Can it be turned into its own function?

jpculp: They are close, but different enough where it wouldn't be as clean as you'd think. It could be split into a `remove_pesig` and `add_pesign` but might end up being about the same number of lines the end. I don't mind adding it though.

twoliter/embedded/imghelper
jpculp: I'll post a revision and bring back the named variable for readability.

twoliter/embedded/imghelper
```diff
@@ -108,6 +110,228 @@ mkfs_data_xfs() {
...
+  popd >/dev/null || exit 1
+}
+
+uefi_copycert() {
+  local efi_image
+  efi_image="${1:?}"
+  mcopy -i "${efi_image}" "${SBKEYS}"/db.{crt,cer} ::/EFI/BOOT
+}
+
+vmlinuz_sbsetup() {
```
bcressey: Please name these functions more consistently - `vmlinuz` is also an EFI binary and also signed with `pesign`, so it's strange to have `uefi_pesign` and `vmlinuz_sbsetup`.

You should also pick either `noun_verb` or `verb_noun` and stick with it. `verb_noun` is more readable and consistent with other functions already in this file.

twoliter/embedded/imghelper
```diff
@@ -108,6 +110,228 @@ mkfs_data_xfs() {
...
+
+vmlinuz_hmac() {
+  local vmlinuz
+  vmlinuz="${1:?}"
+  openssl sha512 -hmac FIPS-FTW-RHT2009 -hex "${vmlinuz}" |
+    awk -v vmlinuz="${vmlinuz}" '{ print $2 "  " vmlinuz }' \
+      >"${vmlinuz%/*}.${vmlinuz##*/}.hmac"
+}
+
+sign_grub() {
```
bcressey: The function is named `sign_grub` but you're not actually signing grub but rather `grub.cfg` ...

twoliter/embedded/rpm2img
```diff
@@ -251,128 +249,20 @@ mv "${ROOT_MOUNT}/boot/efi"/* "${EFI_MOUNT}"
...
-  pesigcheck -i "${mokm}" -n 0 -c "${SBKEYS}/vendor.cer"
-
-  # Replace the embedded gpg public key, then sign grub with the vendor key.
-  "${ARCH}-objdumpcopy" "${grub}" \
-    --file-alignment 4096 \
-    --update-section ".pubkey=${SBKEYS}/config-sign.pubkey"
-  pesign -i "${grub}" -o "${grub}.signed" -s "${CODE_SIGN_KEY[@]}"
-  mv "${grub}.signed" "${grub}"
-  pesigcheck -i "${grub}" -n 0 -c "${SBKEYS}/vendor.cer"
+  uefi_pesign "${EFI_MOUNT}" "${ARCH}" SHIM_SIGN_KEY CODE_SIGN_KEY
```
bcressey: You should refactor this along the lines of `<sign_function> <ARTIFACT>` for consistency with the other signing functions.

You are also not very consistent about when you pass in keys (here) versus when you have the function decide what key to use (`sign_grub`). I would recommend just specializing all of the functions to sign a particular artifact and have the function pick the key to use. There is only one correct choice for a given artifact - using the wrong key will either result in an unbootable image or weaken security.

twoliter/embedded/imghelper
```diff
@@ -108,6 +110,228 @@ mkfs_data_xfs() {
...
+    pesign -i "${grub}" -o "${grub}.unsigned" -u 0 -r
+  fi
+  "${arch}-objdumpcopy" "${grub}.unsigned" \
+    --file-alignment 4096 \
+    --update-section ".pubkey=${SBKEYS}/config-sign.pubkey"
+  pesign -i "${grub}.unsigned" -o "${grub}.signed" -s "${code_sign_key[@]}"
+  mv "${grub}.signed" "${grub}"
+  rm "${grub}.unsigned"
+  pesign -i "${grub}" -l
+  pesigcheck -i "${grub}" -n 0 -c "${SBKEYS}/vendor.cer"
```
bcressey: As @rpkelly mentioned, the duplication here is a sign that this function is doing too much, which is part of why its signature is inconsistent.

twoliter/embedded/imghelper
jpculp: The region isn't currently a build secret passed to the container. I suspect it's not necessary to add, but if so we'd want to address it in a separate PR.

twoliter/embedded/imghelper
jpculp: I think we can trust this conversion (at least as much as we can trust any other conversion in `rpm2img`/`img2img`). We can do a string comparison of some of the contents using `openssl x509` to print, but I'm not sure if we'd want that potentially logged.

jpculp:
- Split `uefi_pesign` into several more targeted functions.
- Renamed some things for consistency and clarity.

rpkelly (approved):

twoliter/embedded/rpm2img
```diff
@@ -251,81 +249,13 @@ mv "${ROOT_MOUNT}/boot/efi"/* "${EFI_MOUNT}"
...
-    aws sts get-caller-identity
-    # Log all PKCS11 helper activity, to simplify debugging.
-    export AWS_KMS_PKCS11_DEBUG=1
-    SB_KEY_SOURCE="aws"
-    SHIM_SIGN_KEY=(-c shim-sign-key -t shim-sign-key)
-    CODE_SIGN_KEY=(-c code-sign-key -t code-sign-key)
+  # shellcheck disable=SC2034  # variables are used externally by reference
+  declare -a SHIM_SIGN_KEY CODE_SIGN_KEY
+  if [[ -s "${HOME}/.config/aws-kms-pkcs11/config.json" ]]; then
+    sbsetup_aws_profile SHIM_SIGN_KEY CODE_SIGN_KEY
```
bcressey: Do we still need `SHIM_SIGN_KEY` and `CODE_SIGN_KEY` to be passed in by reference, or can they be set and saved in `imghelper`?

As a minor nit: we could also move this branch into a single `sbsetup_signing_profile` function that calls the right function (aws or local).

bcressey (approved):

jpculp:
Rebased.

jpculp:
- Moved **SHIM_SIGN_KEY** and **CODE_SIGN_KEY** declaration to `imghelper`.
- Wrapped profile setup under a single `sbsetup_signing_profile` function.
- Moved setup a few lines down to share the **UEFI_SECURE_BOOT** conditional.

bcressey (approved):

PR #222, opened 2024-05-17
move verity generation to imghelper

jpculp:
**Description of changes:**

Moves verity code-block to `imghelper`.

**Testing done:**

- Built and smoke tested `aws-k8s-1.28`.

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/rpm2img
+++ b/twoliter/embedded/rpm2img
@@ -65,11 +65,6 @@ SELINUX_ROOT="/etc/selinux"
 SELINUX_POLICY="fortified"
 SELINUX_FILE_CONTEXTS="${ROOT_MOUNT}/${SELINUX_ROOT}/${SELINUX_POLICY}/contexts/files/file_contexts"
 
-VERITY_VERSION=1
-VERITY_HASH_ALGORITHM=sha256
-VERITY_DATA_BLOCK_SIZE=4096
-VERITY_HASH_BLOCK_SIZE=4096
-
 BOOTCONFIG_DIR="$(mktemp -d)"
 BOOTCONFIG_INPUT="${BOOTCONFIG_DIR}/bootconfig.in"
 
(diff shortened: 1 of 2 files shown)


twoliter/embedded/rpm2img
```diff
@@ -442,26 +442,10 @@ resize2fs -M "${ROOT_IMAGE}"
...
-VERITY_ROOT_HASH="$(grep '^Root hash:' <<<"${veritysetup_output}" | awk '{ print $NF }')"
-VERITY_SALT="$(grep '^Salt:' <<<"${veritysetup_output}" | awk '{ print $NF }')"
-veritysetup verify "${ROOT_IMAGE}" "${VERITY_IMAGE}" "${VERITY_ROOT_HASH}"
+generate_verity "${ROOT_IMAGE}" "${VERITY_IMAGE}" "${partsize[HASH-A]}" \
+  VERITY_VERSION VERITY_HASH_ALGORITHM VERITY_DATA_BLOCK_SIZE \
+  VERITY_HASH_BLOCK_SIZE VERITY_DATA_4K_BLOCKS VERITY_DATA_512B_BLOCKS \
+  VERITY_ROOT_HASH VERITY_SALT
 dd if="${VERITY_IMAGE}" of="${OS_IMAGE}" conv=notrunc bs=1M seek="${partoff[HASH-A]}"
 
 declare -a DM_VERITY_ROOT
```
jpculp, line 389: The reason `generate_verity` isn't `generate_verity_root` is because `img2img` needs the individual values to avoid issues related to order adjustments.

rpkelly (approved):

twoliter/embedded/rpm2img
```diff
@@ -442,26 +442,10 @@ resize2fs -M "${ROOT_IMAGE}"
...
-fi
-VERITY_DATA_4K_BLOCKS="$(grep '^Data blocks:' <<<"${veritysetup_output}" | awk '{ print $NF }')"
-VERITY_DATA_512B_BLOCKS="$((VERITY_DATA_4K_BLOCKS * 8))"
-VERITY_ROOT_HASH="$(grep '^Root hash:' <<<"${veritysetup_output}" | awk '{ print $NF }')"
-VERITY_SALT="$(grep '^Salt:' <<<"${veritysetup_output}" | awk '{ print $NF }')"
-veritysetup verify "${ROOT_IMAGE}" "${VERITY_IMAGE}" "${VERITY_ROOT_HASH}"
+generate_verity "${ROOT_IMAGE}" "${VERITY_IMAGE}" "${partsize[HASH-A]}" \
+  VERITY_VERSION VERITY_HASH_ALGORITHM VERITY_DATA_BLOCK_SIZE \
+  VERITY_HASH_BLOCK_SIZE VERITY_DATA_4K_BLOCKS VERITY_DATA_512B_BLOCKS \
+  VERITY_ROOT_HASH VERITY_SALT
```
bcressey: If the only reason you need to pass in these variables is to fill out the `DM_VERITY_ROOT` array - you can declare the array here and pass in a reference to `generate_verity`.

I see your comment below (`img2img` needs the individual values) - but this change needs to make sense in isolation. And it seems like the `img2img` interface should also be the same: populate an array, and then use a command like this to update `grub.cfg`:
```
sed -i -e "s,^set dm_verity_root=.*,set_dm_verity_root=\"${DM_VERITY_ROOT[@]}\",g" grub.cfg
```

jpculp: I was told there was a concern about the value order potentially getting changed up between versions of Bottlerocket, but if that wasn't the case my original plan was to just have the function set the root array and `sed` the line. Are we still worried about the order?

jpculp: After discussing it further, this was an issue prior to `dm_verity_root` settings being captured in an array.

jpculp:
Switched from generating verity values to the `dm_verity_root` directly.

jpculp:
- Redeclared some vars as local.
- Adjusted formatting.
- Noticed a missing redirect to stderr.

rpkelly (approved):

bcressey (approved):

jpculp:
Rebased

bcressey (approved):

PR #230, opened 2024-05-21
external kit support: publish-kit functionality

### LINKED ISSUE #229: pubsys: add ability to publish a kit
Add a new command to pubsys and twoliter to publish a local kit to a container registry.

The command should do the following:

* Check and validate that the kit archives are there and associated metadata
* Map a vendor to a container registry using a new shape added to Infra.toml
* Build the archites into docker base images
* Push the images to the oci registry

jmt-lab:
**Issue number:** 229

Closes #229 

**Description of changes:**
Added a new command that will publish a kit to pubsys.
Added a new command to twoliter to invoke pubsys resolving paths needed for pubsys
Expanded Infra.toml to have a kit_vendors section to define publishing targets
Add a sample external-kit to the tests 

**Testing done:**

Used sample external-kit structure to publish a kit first by using pubsys directly
Used twoliter to do the same above

**Terms of contribution:**


jmt-lab:
Rebase

tests/projects/external-kit/Infra.toml
```diff
@@ -0,0 +1,2 @@
+[kit_vendors]
+bottlerocket = "localhost:5000"
```
bcressey: For consistency with other `Infra.toml` shapes, I'd say this should be more like:
```
[vendor.bottlerocket]
registry = "localhost:5000"
```

(We are using `vendor` as a concept for the SDK also.)

tests/projects/external-kit/Twoliter.toml
```diff
@@ -0,0 +1,7 @@
+schema-version = 1
+release-version = "1.0.0"
+
+[sdk]
+registry = "twoliter.alpha"
+repo = "bottlerocket-sdk"
+tag = "latest"
```
bcressey: thought we were going to align this to:
```suggestion
[sdk]
name = "bottlerocket-sdk"
version = "1.4.2"
vendor = "bottlerocket"

[vendor.bottlerocket]
registry = "..."
```

We should clean that up as a preliminary commit / PR change so as not to crowd this one. I'm also fine if we commit the unit tests first.

tests/projects/external-kit/build/rpms/core-kit/kit-metadata.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "bottlerocket-test-kit",
+  "version": "1.1.0",
+  "bottlerocket_sdk": {
```
bcressey:
```suggestion
  "sdk": {
```

tests/projects/external-kit/kit-amd64.tar
bcressey, line 1: I am really not excited about committing binary test files to the repo in light of recent industry events.

twoliter/src/cmd/publish_kit.rs
```diff
@@ -0,0 +1,66 @@
...
+#[derive(Debug, Parser)]
+pub(crate) struct PublishKit {
+    /// Path to Twoliter.toml. Will search for Twoliter.toml when absent
+    #[clap(long = "project-path")]
+    project_path: Option<PathBuf>,
+    /// Kit name to build
+    kit_name: String,
+    /// Vendor to publish to
+    vendor: String,
+}
```
bcressey: This doesn't seem right - why do we need to read Twoliter.toml?

@webern can you suggest a way to factor out the `load_or_find_project` and `install_tools` logic so we don't need to repeat for every command?

twoliter/src/cmd/publish_kit.rs
```diff
@@ -0,0 +1,66 @@
...
+    pub(super) async fn run(&self) -> Result<()> {
+        let project = project::load_or_find_project(self.project_path.clone()).await?;
+        let toolsdir = project.project_dir().join("build/tools");
+        install_tools(&toolsdir).await?;
+        let pubsys_path = toolsdir.join("pubsys");
+        let infra_path = project.project_dir().join("Infra.toml");
+        let kit_path = project
+            .project_dir()
+            .join("build")
+            .join("rpms")
```
bcressey: Kits are placed in `build/kits`.

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    let result = Command::new("docker")
+        .args([
+            "build",
+            "--builder=container",
+            "--push",
+            platform.as_str(),
+            tag.as_str(),
+            ".",
+        ])
```
bcressey: I get this for the `builder=container` argument:
```
❯ docker build --builder=container --platform=linux/amd64 --platform=linux/arm64 .
ERROR: no builder "container" found
```

Guessing you followed [these steps](https://docs.docker.com/build/building/multi-platform/)? seems like it has some potentially non-trivial setup requirements for downstream builds.

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+ADD kit-$TARGETARCH.tar /
+"#,
+        KIT_METADATA_KEY, metadata_encoded
+    );
+    let docker_file_path = kit_path.join("Dockerfile");
+    write(&docker_file_path, docker_file)
+        .await
+        .context(error::IoSnafu)?;
+    let tag = format!("--tag={}", target_uri);
+    let platform = format!("--platform={}", platforms.join(","));
```
bcressey: I don't think pubsys should assume that both architectures exist; it should handle the case where one or both are present, and error out if none are.

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    let result = Command::new("docker")
```
bcressey: I'd prefer to generate as close as possible to the final output in `buildsys` when building the kit, ideally in a form that can be pushed directly.

For example:
* buildsys emits a layer blob and OCI metadata for x86_64
* buildsys emits a layer blob and OCI metadata for aarch64
* pubsys pushes the layer blobs to the target registry so they're available for reference in a manifest
* pubsys generates the manifest and pushes that to the registry to complete the operation

It gets more tricky if we want to support building aarch64 and x86_64 on separate hosts, then bring them together in the manifest. You should consult with @jpculp so this plays nicely with existing automation around container builds for multiple architectures.

E.g. it may be sufficient if all pubsys does push the kit for the target arch and the target version and commit:
```
public.ecr.aws/c2b5m1s5/bottlerocket-core-kit:v1.2.0-abc123-arm64
public.ecr.aws/c2b5m1s5/bottlerocket-core-kit:v1.2.0-abc123-amd64
```

And then we come along in a later phase and generate the combined manifest and push that to the final location:
```
public.ecr.aws/bottlerocket/bottlerocket-core-kit:v1.2.0
```

We can have pubsys do that combining logic too but I think it's best as a separate concern, because we haven't had to do the "bring two architectures together" thing yet in pubsys and it warrants some additional thought.

tests/projects/external-kit/Twoliter.toml
jmt-lab: The changes to Twoliter.toml is coming in another PR I will put that one up today

tests/projects/external-kit/build/rpms/core-kit/kit-metadata.json
jmt-lab: I thought we discussed that we should not use generic sdk to prevent any confusion in the design review?

tools/pubsys/src/kit/publish_kit/mod.rs
jmt-lab: This was not intended, I forgot to delete the --builder. I needed it to test the multiarch stuff on my mac

tools/pubsys/src/kit/publish_kit/mod.rs
jmt-lab: It does not assume that at all if you look up above platforms is generated only by what artifacts exist

tools/pubsys/src/kit/publish_kit/mod.rs
jmt-lab: So I am effectively already assuming this but building the multi platform manifest when needed on the pubsys side. There is no need to generate separate metadata here since we are using the docker trick. Its important to note I am not aware of any way to do what you are suggesting in the example using only the docker cli. 

RIght now the assumption flows in this approach are:

* buildsys builds the x86_64 kit and archives it to a archive called kit-amd64.tar
* buildsys builds the aarch64 kit and archives it to a archive called kit-arm64.tar
* I initially though you could have the registry solve the multi-platform part by just pushing each individaul platform separately but i guess you need to do it together to build a multi image.
* pubsys then will detect which platform archives exist and build the image to push

jmt-lab: Discussed offline, now I understand better how this works and sounds like the right move

tools/pubsys/src/kit/publish_kit/mod.rs
jmt-lab: I understand this more now

twoliter/src/cmd/publish_kit.rs
webern: Will do: adding to backlog.

tests/projects/external-kit/kit-amd64.tar
webern: > I am really not excited about committing binary test files to the repo in light of recent industry events.

Can the tar file be created on-the-fly as part of the test set-up? See https://docs.rs/tar/latest/tar/ 

It also feels like this is data that should not be a part of the test project. The test project, in my opinion, should look like what a user's project would look like. In other words, it shouldn't have json and tar files checked in to it that are only there because we are testing partial functionality of the build system. Possible, if that's the case, the name of the test "project" could indicate that it's not a real project structure?

tests/projects/external-kit/build/rpms/core-kit/kit-metadata.json
```diff
@@ -0,0 +1,16 @@
...
+    "vendor": "bottlerocket"
+  },
+  "dependencies": [
+    {
+      "name": "bottlerocket-core-kit",
+      "version": "1.0.0",
+      "vendor": "bottlerocket"
+    }
+  ]
+}
```
webern: I know I'm not the only one around here triggered by missing newlines at the ends of files, so I'll take one for the team and be the one to point it out this time.

tests/projects/external-kit/build/rpms/core-kit/kit-metadata.json
```diff
@@ -0,0 +1,16 @@
+{
```
webern: Why are we checking something in to the build directory of the project? This is problematic. The entire build directory should be gitignored and during a clean these directories are deleted and re-created.

tools/pubsys-config/src/lib.rs
```diff
@@ -26,6 +26,9 @@ pub struct InfraConfig {
 
     // Config for VMware specific subcommands
     pub vmware: Option<VmwareConfig>,
+
+    // Kit Vendors for publishing kits
+    pub kit_vendors: Option<HashMap<String, String>>,
```
webern: Should we call this `vendors` in case we need to push things other than just kits someday?

tools/pubsys/Infra.toml.example
```diff
@@ -82,3 +82,7 @@ datastore = "WorkloadDatastore" # GOVC_DATASTORE
 network = "sddc-cgw-network-1" # GOVC_NETWORK
 folder = "my_folder" # GOVC_FOLDER
 resource_pool = "/SDDC-Datacenter/host/Cluster/Resources/Compute-ResourcePool" # GOVC_RESOURCE_POOL
+
+# Kit vendors informing pubsys where to publish any local kit chosen to be published
+[kit_vendors]
+bottlerocket = "public.ecr.aws/bottlerocket"
```
webern:
```suggestion
bottlerocket = "public.ecr.aws/bottlerocket"

```

sam-berning:
^ rebased, but haven't changed approach yet

tools/pubsys/Infra.toml.example
```diff
@@ -86,3 +86,4 @@ resource_pool = "/SDDC-Datacenter/host/Cluster/Resources/Compute-ResourcePool" #
 # Container vendor specific configuration
 [vendor.bottlerocket]
 registry = "my.vendor/path"
+insecure = false
```
webern: Document this here as well.

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+    let vendor = infra_config
+        .vendor
+        .as_ref()
+        .unwrap()
+        .get(&publish_kit_args.vendor)
+        .context(error::VendorNotFoundSnafu {
+            name: publish_kit_args.vendor.clone(),
+        })?;
+    let vendor_registry_uri = vendor.registry.clone();
+    info!(
```
webern: Feels a little overeager to announce finding something in a file.

```suggestion
    debug!(
```

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+            image,
+        ]);
+    }
+
+    info!("Pushing kit to {}", &target_uri);
+    docker!(["manifest", "push", &target_uri,]);
+
+    info!("Successfully published kit to {}", target_uri);
+
+    // cleans up the transient images. these don't serve a purpose outside of publishing the kit
```
webern: This won't happen if we error somewhere before this. I'm trying to think of a way to make sure it happens.

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+    for (_, image) in platform_images.iter() {
+        docker!(["rmi", image]);
+    }
+
+    Ok(())
+}
+
+fn find_oci_archives(kit_path: &Path, kit_name: &str) -> Result<Vec<OciArchive>> {
+    let pattern = format!(
+        "{}-(?<version>v[0-9.]+-[0-9a-f]+(-dirty)?)-(?<arch>aarch64|x86_64).tar",
```
webern: Is there an indeterminate number of these files? Or is there something more specific that we can look for? i.e. I notice that kit-name is not in the regex.

sam-berning: there can be 1 or 2 files in a valid scenario, one for each arch. I guess the concern with this approach would be that if there were somehow multiple archives with different version/build-ids in the kit directory this would behave weirdly. maybe I can just use this regex to discover the version, then try to load the file for each arch from its specific filename?

> I notice that kit-name is not in the regex.

it's in there, it's just an argument to the format string

tools/pubsys-config/src/lib.rs
```diff
@@ -111,6 +111,10 @@ impl InfraConfig {
 #[derive(Debug, Default, Deserialize, Serialize, PartialEq, Eq, Clone)]
 pub struct Vendor {
     pub registry: String,
+    // needed in cases where the vendor does not use HTTPS. this is to accomodate a weird quirk in
```
jmt-lab, line 114: Not a blocker, but i'm not sure we really want to have this be an option for production release, I could see AppSec maybe throwing up a red flag on this?

tools/pubsys/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2021"
 publish = false
 
 [dependencies]
+anyhow = "1"
```
cbgbt, line 10: It's a bit unfortunate that we're bubbling `anyhow::Error` out of what has become a more public API. It might be worth opening an issue to tighten this up:
* Use an error type in `twoliter` that is owned by the crate for public types (or move them to a tertiary crate with its own error type)
* Remove `anyhow` as a dependency in pubsys.

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+use log::{debug, info, trace};
+use pubsys_config::InfraConfig;
+use regex::Regex;
+use snafu::{ensure, OptionExt, ResultExt};
+use std::fs::read_dir;
+use std::path::{Path, PathBuf};
+use tokio::process::Command;
+
+/// Takes a local kit built using buildsys and publishes it
+/// to a vendor specified in Infra.toml
```
cbgbt: nit: I believe we tend to use 100 as a line length guide for Bottlerocket. I'm not much of a stickler for line length adherence, I just bring it up because these obviously vary from the rest of the file.

twoliter/src/cmd/publish_kit.rs
```diff
@@ -0,0 +1,66 @@
...
+/// Publish a local kit to a container registry
+#[derive(Debug, Parser)]
+pub(crate) struct PublishKit {
+    /// Path to Twoliter.toml. Will search for Twoliter.toml when absent
+    #[clap(long = "project-path")]
+    project_path: Option<PathBuf>,
+    /// Kit name to build
+    kit_name: String,
+    /// Vendor to publish to
+    vendor: String,
```
cbgbt: nit: 
```suggestion
    /// Path to Twoliter.toml. Will search for Twoliter.toml when absent
    #[clap(long = "project-path")]
    project_path: Option<PathBuf>,

    /// Kit name to build
    kit_name: String,

    /// Vendor to publish to
    vendor: String,
```

twoliter/src/cmd/publish_kit.rs
```diff
@@ -0,0 +1,66 @@
...
+    pub(super) async fn run(&self) -> Result<()> {
+        let project = project::load_or_find_project(self.project_path.clone()).await?;
+        let toolsdir = project.project_dir().join("build/tools");
+        install_tools(&toolsdir).await?;
+        let pubsys_path = toolsdir.join("pubsys");
+        let infra_path = project.project_dir().join("Infra.toml");
+        let kit_path = project
+            .project_dir()
+            .join("build")
+            .join("kits")
```
cbgbt, line 49: I think this is simplified after a recent rebase:

```suggestion
        let kit_path = project
            .external_kits_dir()
```

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+#[derive(Debug, Parser)]
+pub(crate) struct PublishKitArgs {
+    /// Location of the local kit
+    #[arg(long)]
+    kit_path: PathBuf,
```
cbgbt, line 16: nit: For argument parsing-related structs, I believe we tend to newline separate arguments.

tools/pubsys/src/kit/publish_kit/mod.rs
sam-berning: alright, updated

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+    let infra_config = InfraConfig::from_path_or_lock(&args.infra_config_path, false)
+        .context(error::ConfigSnafu)?;
+    trace!("Parsed infra config: {:?}", infra_config);
+    ensure!(infra_config.vendor.is_some(), error::NoVendorsSnafu);
+
+    // Fetch the vendor container registry uri
+    let vendor = infra_config
+        .vendor
+        .as_ref()
+        .unwrap()
```
cbgbt: Should probably return an error rather than panic here.

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+        })?;
+    let vendor_registry_uri = vendor.registry.clone();
+    debug!(
+        "Found vendor container registry at uri: {}",
+        vendor_registry_uri
+    );
+
+    // Auto resolve the expected paths for the kit contents archive
+    let kit_path = publish_kit_args.kit_path.as_path();
+    let kit_name = kit_path.file_name().unwrap().to_string_lossy();
```
cbgbt: Should we similarly handle `file_name()` returning `None` with a returned error?

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+    let kit_version = &archives.first().unwrap().version;
+    let mut platform_images = Vec::new();
+    for archive in archives.iter() {
+        let path = kit_path.join(&archive.filename);
+
+        let out = docker!(["load", format!("--input={}", path.display()).as_str(),]);
+        let out = String::from_utf8_lossy(&out);
+        let digest_expression =
+            Regex::new("(?<digest>sha256:[0-9a-f]{64})").context(error::RegexSnafu)?;
+        let caps = digest_expression.captures(&out).unwrap();
```
cbgbt: Probably best to capture an error here if Docker does the wrong thing, rather than panic.

tools/pubsys-config/src/lib.rs
```diff
@@ -111,6 +111,10 @@ impl InfraConfig {
 #[derive(Debug, Default, Deserialize, Serialize, PartialEq, Eq, Clone)]
 pub struct Vendor {
     pub registry: String,
+    // Needed in cases where the vendor does not use HTTPS. This is to accomodate a weird quirk in
+    // the implementation of `docker manifest create`.
+    #[serde(default)]
+    pub insecure: bool,
```
bcressey: Let's get rid of this - for now we just won't support registries that don't use HTTPS.

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    Ok(())
+}
+
+fn find_version(kit_path: &Path, kit_name: &str) -> Result<String> {
+    let pattern = format!(
+        "{}-(?<version>v[0-9.]+-[0-9a-f]+(-dirty)?)-(aarch64|x86_64).tar",
+        kit_name
+    );
```
cbgbt: Is it possible for there to be multiple versions present in the path?

cbgbt (commented):
Almost every comment I have is stylistic in nature or refers to error handling. Given that we're trying to land everything for https://github.com/bottlerocket-os/bottlerocket/issues/4048, I'd be okay with opening an issue and deferring on these.

I do have one potentially blocking question about finding the OCI archives, though.

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+        docker!(["tag", digest, &arch_specific_target_uri,]);
+
+        info!(
+            "Pushing kit image for platform {} to {}",
+            arch, &arch_specific_target_uri
+        );
+        docker!(["push", &arch_specific_target_uri,]);
+
+        let docker_arch =
+            DockerArchitecture::try_from(arch).context(error::InvalidArchitectureSnafu)?;
```
bcressey: nit: can do this at the outset so we don't do any `docker` commands if this is just going to fail

(we know it's infallible though)

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    info!("Successfully published kit to {}", target_uri);
+
+    // Cleans up the transient images. These don't serve a purpose outside of publishing the kit
+    // and take up space.
+    info!("Cleaning up local images");
+    docker!(["manifest", "rm", &target_uri]);
+    for (_, image) in platform_images.iter() {
+        docker!(["rmi", image]);
+    }
```
bcressey: Does this mean we won't keep a local copy of an image after we push it to a kit? Seems like cleaning up the manifest is fine but we might need the image again right away to test the kit when building variants.

twoliter/src/cmd/publish_kit.rs
sam-berning: looks like it's slightly different, `external_kits_dir()` resolves to `build/external-kits`

tools/pubsys/src/kit/publish_kit/mod.rs
sam-berning: my understanding from a discussion with Matt and Ben yesterday is that it shouldn't be possible. when we build a kit `buildsys` should clean up any old archives under `build/kits/<kit-name>`

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+                .path()
+                .file_name()
+                .unwrap()
+                .to_string_lossy()
+                .to_string()
+        })
+        .collect::<Vec<_>>()
+        .into_iter()
+        .find(|file_name| expression.is_match(file_name))
+        .context(error::NoArchiveSnafu { path: kit_path })?;
```
bcressey: The way we handle this elsewhere in `pubsys` is to take a command line argument that exactly specifies the version - that way we're 100% sure we're pushing the artifact that matches the current repo state.

So we should arrange for `twoliter` to pass in the full path or enough data to reconstruct the file name:
```
${KIT:?}-v${VERSION_ID:?}-${BUILD_ID:?}-${ARCH:?}"
```

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub(super)))]
+    pub(crate) enum Error {
+        #[snafu(display("Error reading config: {}", source))]
+        Config { source: pubsys_config::Error },
+        #[snafu(display("Invalid architecture: {}", source))]
+        InvalidArchitecture { source: anyhow::Error },
+        #[snafu(display("IO error occurred: {}", source))]
+        Io { source: std::io::Error },
```
bcressey: For I/O errors we always want to give more context - what program we were trying to run, what directory we were trying to read. Otherwise the errors are maddeningly obscure.

tools/pubsys/src/kit/publish_kit/mod.rs
sam-berning: alright I'll keep the images around

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    let mut platform_images = Vec::new();
+    for arch in ["aarch64", "x86_64"] {
+        let docker_arch =
+            DockerArchitecture::try_from(arch).context(error::InvalidArchitectureSnafu)?;
+
+        let kit_filename = format!("{}-{}-{}.tar", &kit_name, &kit_version, arch);
+        let path = kit_path.join(&kit_filename);
+
+        let out = docker!(["load", format!("--input={}", path.display()).as_str(),]);
```
bcressey, line 81: For ease of local development, I think we could skip over architectures where the kit is not available. But that creates its own headaches so this is good for now.

tools/pubsys/src/kit/publish_kit/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    info!("Pushing kit to {}", &target_uri);
+    docker!(["manifest", "push", &target_uri,]);
+
+    info!("Successfully published kit to {}", target_uri);
+
+    // Cleans up the manifest list. This doesn't serve a purpose outside of publishing the kit and
+    // takes up space.
+    info!("Cleaning up local image");
```
bcressey:
```suggestion
    // Cleans up the manifest list, so subsequent twoliter operations have to fetch the current one
    info!("Cleaning up local manifest");
```

bcressey (approved):

cbgbt (approved):

webern (approved):

bcressey (approved):

PR #231, opened 2024-05-22
add support for repacking a variant

jpculp:
**Description of changes:**

This adds a new tool to take an existing Bottlerocket image and repackage it with updated signatures and artifacts.

**Testing done:**

- [x] Repack `aws-k8s-1.27` _(no secure boot)_.
- [x] Repack `aws-k8s-1.28` _(secure boot)_.
- [x] Repack `vmware-k8s-1.28`.
- [x] Repack `aws-k8s-1.28` with custom CA bundle using #232.
- [x] Verify nothing prints on successful repack.
- [x] Verify everything prints on unsuccessful repack.
- [x] Verify logs are wiped between repacks.
- [x] Verify log symlink is correct.
- [x] Verify `clean-logs` cleans the logs.
- [x] Verify `exit 1` if something goes wrong in `debugfs`.
 
**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/Makefile.toml
+++ b/twoliter/embedded/Makefile.toml
@@ -14,6 +14,7 @@ BUILDSYS_PACKAGES_DIR = "${BUILDSYS_BUILD_DIR}/rpms"
 BUILDSYS_KITS_DIR = "${BUILDSYS_BUILD_DIR}/kits"
 BUILDSYS_STATE_DIR = "${BUILDSYS_BUILD_DIR}/state"
 BUILDSYS_IMAGES_DIR = "${BUILDSYS_BUILD_DIR}/images"
+BUILDSYS_LOGS_DIR = "${BUILDSYS_BUILD_DIR}/logs"
 BUILDSYS_TOOLS_DIR = "${BUILDSYS_ROOT_DIR}/tools"
 BUILDSYS_SOURCES_DIR = "${BUILDSYS_ROOT_DIR}/sources"
 BUILDSYS_SBKEYS_DIR = "${BUILDSYS_ROOT_DIR}/sbkeys"
@@ -850,8 +851,19 @@ export PATH="${TWOLITER_TOOLS_DIR}:${PATH}"
 # Parse the variant into its components and set additional variables.
 eval "$(bottlerocket-variant)"
 
-buildsys repack-variant \
-  --cargo-manifest-dir variants/${BUILDSYS_VARIANT}
+OUTPUT_LOGS_DIR="${BUILDSYS_LOGS_DIR}/${BUILDSYS_ARCH}-${BUILDSYS_VARIANT}"
+REPACK_OUTPUT_LOG="${OUTPUT_LOGS_DIR}/${BUILDSYS_VERSION_FULL}/repack.out"
+
+rm -rf "${OUTPUT_LOGS_DIR:?}"
+mkdir -p "${OUTPUT_LOGS_DIR}/${BUILDSYS_VERSION_FULL}"
+if ! buildsys repack-variant \
+  --cargo-manifest-dir "variants/${BUILDSYS_VARIANT}" \
+  >"${REPACK_OUTPUT_LOG}"; then
+  printf '\n'
+  cat "${REPACK_OUTPUT_LOG}"
+  exit 1
+fi
+ln -snf "${BUILDSYS_VERSION_FULL}" "${OUTPUT_LOGS_DIR}/latest"
 '''
 ]
 
@@ -1535,6 +1547,7 @@ dependencies = [
   "clean-packages",
   "clean-kits",
   "clean-images",
+  "clean-logs",
   "clean-repos",
   "clean-state",
   "clean-tools",
@@ -1584,6 +1597,14 @@ script_runner = "bash"
 script = [
 '''
 rm -rf ${BUILDSYS_IMAGES_DIR}
+'''
+]
+
+[tasks.clean-logs]
+script_runner = "bash"
+script = [
+'''
+rm -rf ${BUILDSYS_LOGS_DIR}
 '''
 ]
 
--- a/twoliter/embedded/img2img
+++ b/twoliter/embedded/img2img
@@ -1,3 +1,320 @@
 #!/usr/bin/env bash
 
-echo "img2img not yet implemented" >&2 ; exit 1
+set -eux -o pipefail
+shopt -qs failglob
+
+OUTPUT_FMT="raw"
+OVF_TEMPLATE=""
+
+UEFI_SECURE_BOOT="no"
+
+for opt in "$@"; do
+  optarg="$(expr "${opt}" : '[^=]*=\(.*\)')"
+  case "${opt}" in
+  --input-dir=*) INPUT_DIR="${optarg}" ;;
+  --output-dir=*) OUTPUT_DIR="${optarg}" ;;
+  --output-fmt=*) OUTPUT_FMT="${optarg}" ;;
+  --os-image-size-gib=*) OS_IMAGE_SIZE_GIB="${optarg}" ;;
+  --data-image-size-gib=*) DATA_IMAGE_SIZE_GIB="${optarg}" ;;
+  --os-image-publish-size-gib=*) OS_IMAGE_PUBLISH_SIZE_GIB="${optarg}" ;;
+  --data-image-publish-size-gib=*) DATA_IMAGE_PUBLISH_SIZE_GIB="${optarg}" ;;
+  --partition-plan=*) PARTITION_PLAN="${optarg}" ;;
+  --ovf-template=*) OVF_TEMPLATE="${optarg}" ;;
+  --with-uefi-secure-boot=*) UEFI_SECURE_BOOT="${optarg}" ;;
+  *)
+    echo "unexpected arg: ${opt}" >&2
+    exit 1
+    ;;
+  esac
+done
+
+WORKDIR="$(mktemp -d)"
+pushd "${WORKDIR}" >/dev/null
+
+# Clean up working directory to reduce size of layer.
+cleanup() {
+  [[ -d "${WORKDIR}" ]] && rm -rf "${WORKDIR}"
+}
+trap 'cleanup' EXIT
+
+# Since debugfs doesn't pass return codes, extrapolate failures from STDERR.
+check_debugfs_errors() {
+  local stderr
+  stderr="$1"
+
+  # debugfs prints it's version info to STDERR, so we need to filter it out.
+  # shellcheck disable=SC2312  # grep returns an error code if no match.
+  if [[ "$(grep -vc '^debugfs ' "${stderr}")" -ne 0 ]]; then
+    grep -v '^debugfs ' "${stderr}" >&2
+    exit 1
+  fi
+}
+
+# Import the partition helper functions.
+# shellcheck source=partyplanner
+. "${0%/*}/partyplanner"
+
+# Import the image helper functions.
+# shellcheck source=imghelper
+. "${0%/*}/imghelper"
+
+# Validate that the values for the args are sane.
+sanity_checks \
+  "${OUTPUT_FMT}" "${PARTITION_PLAN}" "${OVF_TEMPLATE}" "${UEFI_SECURE_BOOT}"
+
+###############################################################################
+# Section 1: prepare working environment
+
+# Store output artifacts in a versioned directory.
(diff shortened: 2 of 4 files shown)


twoliter/embedded/img2img
```diff
@@ -1,3 +1,318 @@
...
+    echo "start mismatch between partition plan and disk: '${part}'" >&2
+    exit 1
+  fi
+
+  if [[ "${partsize["${part}"]}" -ne "${sgdisk_size%.0}" ]]; then
+    echo "size mismatch between partition plan and disk: '${part}'" >&2
+    exit 1
+  fi
+done
+
```
jpculp: I'm probably going to replace this block with a new `get_partition_sizes` function in `partyplanner` and compare one array to the other.

twoliter/embedded/img2img
```diff
@@ -1,3 +1,318 @@
...
+  mkdir -p "${EFI_MOUNT}/EFI/BOOT"
+  mcopy -i "${EFI_IMAGE}" -sv ::/EFI/BOOT "${EFI_MOUNT}/EFI/"
+  # Remove the original signing certificates to avoid confusion.
+  rm -f "${EFI_MOUNT}/EFI/BOOT/db.cer" "${EFI_MOUNT}/EFI/BOOT/db.crt"
+fi
+
+###############################################################################
+# Section 3: replace needed artifacts
+
+# TODO - CA bundle code goes here
```
jpculp, line 137: This might be a separate PR.

twoliter/embedded/imghelper
```diff
@@ -244,6 +282,18 @@ mkfs_data_xfs() {
   dd if="${bottlerocket_data}" of="${target}" conv=notrunc bs=1M seek="${offset}"
 }
 
+check_image_size() {
```
jpculp, line 285: I considered putting this in `partyplanner`, but if it lives here then we can use it in `generate_verity_root` without sourcing `partyplanner`.

twoliter/embedded/img2img
```diff
@@ -1,3 +1,318 @@
...
+  "${OUTPUT_FMT}" "${PARTITION_PLAN}" "${OVF_TEMPLATE}" "${UEFI_SECURE_BOOT}"
+
+###############################################################################
+# Section 1: prepare working environment
+
+# Store output artifacts in a versioned directory.
+OUTPUT_DIR="${OUTPUT_DIR}/${VERSION_ID}-${BUILD_ID}"
+mkdir -p "${OUTPUT_DIR}"
+
+declare OS_IMAGE DATA_IMAGE
```
bcressey: nit: declare immediately before use by `stage_images`

twoliter/embedded/img2img
```diff
@@ -1,3 +1,318 @@
...
+
+for part in EFI-A BOOT-A ROOT-A HASH-A; do
+  # Each partition is aligned to a 1 MiB boundary, and extends to the sector
+  # before the next partition starts. Specify the end point in sectors so we
+  # can subtract a sector to fix the off-by-one error that comes from adding
+  # start and size together. (1 MiB contains 2048 512-byte sectors.)
+  part_start="${partoff["${part}"]}"
+  part_end="$((part_start + partsize["${part}"]))"
+  part_end="$((part_end * 2048 - 1))"
+done
```
bcressey: Can we remove this? `part_end` doesn't seem to be used.

twoliter/embedded/img2img
```diff
@@ -1,3 +1,318 @@
...
+rm ${artifact#"${ROOT_MOUNT}"}
+write ${artifact} ${artifact#"${ROOT_MOUNT}"}
+ea_set ${artifact#"${ROOT_MOUNT}"} security.selinux system_u:object_r:os_t:s0
+EOF
+  done
+fi
+
+# Validate and write root image back to the OS image.
+check_image_size "${ROOT_IMAGE}" "${partsize["ROOT-A"]}"
+dd if="${ROOT_IMAGE}" of="${OS_IMAGE}" \
```
bcressey, line 163: We should add `iflag=fullblock` here

twoliter/embedded/img2img
```diff
@@ -1,3 +1,318 @@
...
+
+# Generate a new root verity.
+declare -a DM_VERITY_ROOT
+generate_verity_root "${ROOT_IMAGE}" "${VERITY_IMAGE}" \
+  "${partsize["HASH-A"]}" DM_VERITY_ROOT
+
+# Validate and write root verity back to the OS image. The image size check
+# isn't needed here as it's already done in `generate_verity_root`.
+dd if="${VERITY_IMAGE}" of="${OS_IMAGE}" \
+  conv=notrunc bs=1M seek="${partoff["HASH-A"]}"
```
bcressey: The comment refers to validation but I just see the write?

twoliter/embedded/img2img
```diff
@@ -1,3 +1,318 @@
...
+  sbsetup_signing_profile
+
+  # Resign the EFI artifacts.
+  sign_shim "${shim}"
+  sign_mokm "${mokm}"
+  sign_grub "${grub}"
+
+  # Write the resigned artifacts back to the EFI image.
+  mcopy -i "${EFI_IMAGE}" -ov "${EFI_MOUNT}/EFI/BOOT"/*.efi ::/EFI/BOOT
+  mcopy -i "${EFI_IMAGE}" -ov "${SBKEYS}"/db.{crt,cer} ::/EFI/BOOT
```
bcressey: Can we call `provide_certs`?

twoliter/embedded/img2img
```diff
@@ -1,3 +1,318 @@
...
+ea_set .vmlinuz.hmac security.selinux system_u:object_r:os_t:s0
+EOF
+
+fi
+
+###############################################################################
+# Section 6: update boot partition
+
+GRUB_CONFIG="${BOOT_MOUNT}/grub/grub.cfg"
+[[ -s "${GRUB_CONFIG}" ]]
```
bcressey: Do we need this check? It seems a bit lonely - we aren't doing similar checks for other files we expect.

twoliter/embedded/img2img
```diff
@@ -1,3 +1,318 @@
...
+compress_image "ext4.lz4" "root_image" "${OUTPUT_DIR}"
+
+symlink_image "ext4.lz4" "boot_image" "${OUTPUT_DIR}"
+symlink_image "verity.lz4" "verity_image" "${OUTPUT_DIR}"
+symlink_image "ext4.lz4" "root_image" "${OUTPUT_DIR}"
+
+popd >/dev/null
+
+# Ensure proper ownership of the final artifacts.
+find "${OUTPUT_DIR}" -type f -print -exec chown 1000:1000 {} \;
```
bcressey, line 320: Might also be worth cleaning up the temp files, as `rpm2img` does.

bcressey (commented):
Looks good!

twoliter/embedded/img2img
jpculp: Towards the top of the script there is a trap to clean-up **WORKDIR**, which is where all the temp files are consolidated. Would you prefer it just be at the end like in `rpm2img`?

twoliter/embedded/img2img
jpculp: We can drop.

twoliter/embedded/img2img
jpculp: It's on [L316](https://github.com/bottlerocket-os/twoliter/pull/231/files#diff-0b8c4b6920fcc7cd541e1b6cac7893cf86c9b2e80859c9444ebdf3339b5f3196R316) of `imghelper` since it's checked both in `img2img` and `rpm2img`.

jpculp:
- Removed unnecessary blocks of code.
- Added `iflag=fullblock` when `dd`ing to the OS image.
- Added clobbering and verbosity to `mcopy` in `provide_certs`.

jpculp:
Added a more exhaustive comparison of the partition plan and the input image's partition table.

twoliter/embedded/partyplanner
```diff
@@ -118,6 +118,94 @@ PRIVATE_SCALE_FACTOR="24"
...
+      <<<"${os_image_table}" | head -1 || true)"
+      sectors="$(awk /BOTTLEROCKET-RESERV/'{print $4}' \
+      <<<"${os_image_table}" | head -1 || true)"
+      ;;
+    RESERVED-B)
+      offset="$(awk /BOTTLEROCKET-RESERV/'{print $2}' \
+      <<<"${os_image_table}" | tail -1 || true)"
+      sectors="$(awk /BOTTLEROCKET-RESERV/'{print $4}' \
+      <<<"${os_image_table}" | tail -1 || true)"
+      ;;
```
jpculp: This is a little gross because the labels for **RESERVED-A/B** are being truncated. The alternative is iterating through each partition in the GPT table one-by-one, collecting the detailed information, and populating the arrays that way.

twoliter/embedded/img2img
bcressey: No, the trap is better, I just missed the implicit cleanup.

bcressey (approved):

twoliter/embedded/img2img
```diff
@@ -1,3 +1,302 @@
...
+    ;;
+  esac
+done
+
+WORKDIR="$(mktemp -d)"
+pushd "${WORKDIR}" >/dev/null
+
+# Clean up working directory to reduce size of layer.
+cleanup() {
+  [[ -n "${WORKDIR}" ]] && rm -rf "${WORKDIR}"
```
rpkelly: Why not `-d` here?

jpculp:
Added logging mechanism to keep successful builds quiet, but still provide context on error.

jpculp:
- Renamed some variables to reflect **p**_(artition)_**t**_(able)_.
- Made `case` indention consistent with the rest of `partyplanner`.
- Switched cleanup check to `-d`.

twoliter/embedded/img2img
jpculp: If this merges first I'll add a commit to #232.

twoliter/embedded/img2img
jpculp: Replaced.

twoliter/embedded/Makefile.toml
```diff
@@ -850,8 +851,15 @@ export PATH="${TWOLITER_TOOLS_DIR}:${PATH}"
...
 
+OUTPUT_LOGS_DIR="${BUILDSYS_LOGS_DIR}/${BUILDSYS_ARCH}-${BUILDSYS_VARIANT}"
+REPACK_OUTPUT_LOG="${OUTPUT_LOGS_DIR}/${BUILDSYS_VERSION_FULL}/repack.out"
+
+rm -rf "${OUTPUT_LOGS_DIR:?}"
+mkdir -p "${OUTPUT_LOGS_DIR}/${BUILDSYS_VERSION_FULL}"
 buildsys repack-variant \
-  --cargo-manifest-dir variants/${BUILDSYS_VARIANT}
+  --cargo-manifest-dir "variants/${BUILDSYS_VARIANT}" \
+  > "${REPACK_OUTPUT_LOG}" || cat "${REPACK_OUTPUT_LOG}"
```
bcressey: this shows the log on error but clobbers the non-zero exit code, which we need to pass up to avoid reporting success

bcressey (approved):

jpculp:
`exit 1` on repack failure en lieu of the return code captured by the conditional.

rpkelly (approved):

jpculp:
Added a check to validate `debugfs` commands ran successfully as it does not support return codes _(even when running a single command at a time with `-R`)_.

twoliter/embedded/img2img
```diff
@@ -1,3 +1,320 @@
...
+pushd "${WORKDIR}" >/dev/null
+
+# Clean up working directory to reduce size of layer.
+cleanup() {
+  [[ -d "${WORKDIR}" ]] && rm -rf "${WORKDIR}"
+}
+trap 'cleanup' EXIT
+
+# Since debugfs doesn't pass return codes, extrapolate failures from STDERR.
+check_debugfs_errors() {
```
bcressey, line 41: There are spots in `rpm2img` that can use this check also (can be a follow-up PR).

bcressey (approved):

twoliter/embedded/img2img
jpculp: I wasn't aware, but I agree we should check both. I'll try to follow up with a separate PR for that.

rpkelly (approved):

PR #234, opened 2024-05-23
twoliter: adjust project specifications and infra specification

### LINKED ISSUE #233: Adjust Twoliter.toml, Infra.toml and define specification for Twoliter.lock
We need to add new fields to Twoliter.toml and Infra.toml to implement external publish and pull of kits. In addition we need to define the shape for Twoliter.lock

jmt-lab:
**Issue number:** 233

Closes #233

**Description of changes:**

* Adds vendor section to Twoliter.toml
* Adds kit section to Twoliter.toml
* Adds validation for vendors
* Adjusts sdk definition to use same dependency shape as kits
* Defines shape of Twoliter.lock

**Testing done:**
Expanded tests to the new shape. Added unit test to verify vendor validation

**Terms of contribution:**

### DIFF
--- a/twoliter/Cargo.toml
+++ b/twoliter/Cargo.toml
@@ -21,6 +21,7 @@ futures= "0.3"
 hex = "0.4"
 log = "0.4"
 non-empty-string = { version = "0.2", features = [ "serde" ] }
+semver = { version = "1", features = ["serde"] }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 sha2 = "0.10"
--- a/twoliter/src/main.rs
+++ b/twoliter/src/main.rs
@@ -6,13 +6,13 @@ mod cargo_make;
 mod cmd;
 mod common;
 mod docker;
+mod lock;
 mod project;
 mod schema_version;
-mod tools;
-
 /// Test code that should only be compiled when running tests.
 #[cfg(test)]
 mod test;
+mod tools;
 
 /// `anyhow` prints a nicely formatted error message with `Debug`, so we can return a result from
 /// the `main` function.
--- a/twoliter/src/project.rs
+++ b/twoliter/src/project.rs
@@ -6,8 +6,10 @@ use async_recursion::async_recursion;
 use async_walkdir::WalkDir;
 use futures::stream::StreamExt;
 use log::{debug, info, trace, warn};
+use semver::Version;
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha512};
+use std::collections::BTreeMap;
 use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
 use toml::Table;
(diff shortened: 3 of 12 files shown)


bcressey (approved):
Nice.

twoliter/Cargo.toml
```diff
@@ -17,25 +17,26 @@ clap = { version = "4", features = ["derive", "env", "std"] }
 env_logger = "0.11"
 filetime = "0.2"
 flate2 = "1"
-futures= "0.3"
+futures = "0.3"
 hex = "0.4"
 log = "0.4"
-non-empty-string = { version = "0.2", features = [ "serde" ] }
+non-empty-string = { version = "0.2", features = ["serde"] }
+semver = { version = "1", features = ["serde"] }
```
webern: Please turn off auto-formatting for TOML files. I don't expect all of our IDEs to agree on the right formatting and we don't want this churn in our git history.

twoliter/src/main.rs
```diff
@@ -10,6 +10,7 @@ mod project;
 mod schema_version;
 mod tools;
 
+mod lock;
 /// Test code that should only be compiled when running tests.
```
webern:
```suggestion
mod tools;
mod lock;

/// Test code that should only be compiled when running tests.
```

twoliter/src/project.rs
```diff
@@ -156,6 +193,23 @@ impl Project {
...
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, Ord, PartialOrd)]
+#[serde(rename_all = "kebab-case")]
+pub(crate) struct Vendor {
+    pub registry: String,
+}
+
+/// This represents a dependency on a container, primarily used for kits
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, Ord, PartialOrd)]
+#[serde(rename_all = "kebab-case")]
+pub(crate) struct Dependency {
```
webern: I don't like the name `Dependency`. It is over-used already.

```suggestion
pub(crate) struct Image {
```

webern (commented):
I kind of hate the whole vendor thing. Especially when I see it forcing a change to the erstwhile easy-to-understand SDK section. @bcressey can we make the usage of "vendor" optional? It feels like a lot of overhead just to avoid possibly pasting a url a couple of times (down the road, when there could possibly be more than one kit url).

bcressey:
> @bcressey can we make the usage of "vendor" optional? It feels like a lot of overhead just to avoid possibly pasting a url a couple of times (down the road, when there could possibly be more than one kit url).

I don't really want to special case "bottlerocket" as a vendor or default / implied vendor. That seems like the bigger problem than copy/pasting a few lines for a vendor registry definition. We'll end up writing brittle code that just assumes Bottlerocket is the only vendor if we don't force ourselves to deal with that case.

twoliter/src/project.rs
bcressey: Fair call on `Dependency` but `Image` is perhaps the most overloaded word we have across Bottlerocket 😆

jmt-lab:
Addresses comments

webern (approved):

webern:
> I don't really want to special case "bottlerocket" as a vendor or default / implied vendor.

That's not what I meant. The concept of listing repositories, even if the same repository is listed more than once, just seems more straight forward to me. Not blocking though 👍.

PR #235, opened 2024-05-24
fix lints and formatting in core scripts

jpculp:
**Description of changes:**

Tightens up formatting and linting to keep consistency between `rpm2img`, `img2img`, `imghelper`, and `partyplanner`.

**Testing done:**

Built and smoke tested Bottlerocket

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/imghelper
+++ b/twoliter/embedded/imghelper
@@ -1,9 +1,8 @@
 #!/usr/bin/env bash
-# shellcheck disable=SC2034 # some variables are used through parameter substitution
+# shellcheck disable=SC2034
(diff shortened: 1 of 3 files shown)


jpculp:
I do not have strong opinions on these. If we want to keep the original indention for `case` for example, I can switch it in `img2img`/`imghelper` instead.

rpkelly (approved):

jpculp:
- Rebased.
- Fixed a capitalization error in `rpm2img` on L94 and L464.

bcressey (approved):

PR #236, opened 2024-05-24
pubsys: add fetch variant subcommand

### LINKED ISSUE #176: Tooling to import images from the TUF repo

gthao313:
**Issue number:**

Closes #176

**Description of changes:**

add fetch variant subcommand to pubsys to pull a pre-built variant image. Supports both 'unified' and 'split' partitioned variants.

**Testing done:**

- [x] fetch a variant
```shell
[cargo-make][1] INFO - Running Task: fetch-variant
23:01:38 [INFO] Downloading targets to "/home/jpculp/workspace/bottlerocket/build/images/aarch64-aws-k8s-1.28/.tmp34dXdt"
23:01:38 [INFO] 	-> bottlerocket-aws-k8s-1.28-aarch64-1.20.0-fcf71a47.img.lz4
23:01:39 [INFO] 	-> bottlerocket-aws-k8s-1.28-aarch64-1.20.0-fcf71a47-data.img.lz4
23:01:40 [INFO] Moving targets to "/home/jpculp/workspace/bottlerocket/build/images/aarch64-aws-k8s-1.28/1.20.0-fcf71a47"
```

- [x] fetch a friendly variant

```shell
[cargo-make][1] INFO - Running Task: fetch-variant
23:01:47 [INFO] Downloading targets to "/home/jpculp/workspace/bottlerocket/build/images/aarch64-aws-k8s-1.28/.tmptlIreW"
23:01:47 [INFO] 	-> bottlerocket-aws-k8s-1.28-aarch64-v1.20.0.img.lz4
23:01:48 [INFO] 	-> bottlerocket-aws-k8s-1.28-aarch64-v1.20.0-data.img.lz4
23:01:49 [INFO] Moving targets to "/home/jpculp/workspace/bottlerocket/build/images/aarch64-aws-k8s-1.28/1.20.0-fcf71a47"
```

- [x] fetch an ova

```shell
[cargo-make][1] INFO - Running Task: fetch-variant
23:01:53 [INFO] Downloading targets to "/home/jpculp/workspace/bottlerocket/build/images/x86_64-vmware-k8s-1.28/.tmp4cfmX2"
23:01:53 [INFO] 	-> bottlerocket-vmware-k8s-1.28-x86_64-v1.20.0.ova
23:01:54 [INFO] Moving targets to "/home/jpculp/workspace/bottlerocket/build/images/x86_64-vmware-k8s-1.28/1.20.0-fcf71a47"
```

**Terms of contribution:**

### DIFF
--- a/tools/pubsys/src/repo/fetch_variant/mod.rs
+++ b/tools/pubsys/src/repo/fetch_variant/mod.rs
@@ -1,0 +1,287 @@
+//! The fetch_variant module owns the 'fetch-variant' subcommand and provides methods for fetching
+//! a given variant and download its image targets.
+
+use crate::repo::{error as repo_error, repo_urls};
+use crate::{repo, Args};
+use clap::Parser;
+use log::{debug, info, trace};
+use pubsys_config::InfraConfig;
+use snafu::{OptionExt, ResultExt};
+use std::path::PathBuf;
+use tempfile::tempdir_in;
+use tough::{Prefix, Repository, RepositoryLoader, TargetName};
+use url::Url;
+
+use buildsys::manifest::{ImageFormat, ManifestInfo, PartitionPlan};
+
+/// fetching and downdloaing the image targets of a given variant
+#[derive(Debug, Parser)]
+pub(crate) struct FetchVariantArgs {
+    #[arg(long)]
+    /// Use this named repo infrastructure from Infra.toml
+    repo: String,
+
+    #[arg(long)]
+    /// The architecture of the repo being validated
+    arch: String,
+
+    #[arg(long)]
+    /// The variant of the repo being validated
+    variant: String,
+
+    #[arg(long)]
+    /// The version of the repo being validated
+    version: String,
+
+    #[arg(long)]
+    /// The build of the repo being validated
+    build: String,
+
+    #[arg(long)]
+    /// Path to root.json for this repo
+    root_role_path: PathBuf,
+
+    #[arg(long)]
+    /// Where to store the downloaded img files
+    outdir: PathBuf,
+
+    #[arg(long)]
+    /// The variant name without extension
+    filename_prefix: String,
+
+    #[arg(long)]
+    /// The manifest of the variant
+    variant_manifest: PathBuf,
+}
+
+/// Download targets
+async fn handle_download(
+    repository: &Repository,
+    outdir: &PathBuf,
+    raw_names: &[String],
+) -> Result<(), Error> {
+    let target_names: Result<Vec<TargetName>, Error> = raw_names
+        .iter()
+        .map(|s| TargetName::new(s).context(error::InvalidTargetNameSnafu))
+        .collect();
+    let target_names = target_names?;
+
+    if target_names.is_empty() {
+        return error::MissingTargetNamesSnafu.fail();
+    };
+
+    // Attempt to make a temporary directory in the parent directory of the outdir
+    let tempdir = match outdir.parent() {
+        Some(in_dir) => tempdir_in(in_dir).context(error::CreateTempDirSnafu)?,
+        None => return error::InvalidOutdirSnafu { path: outdir }.fail(),
+    };
+    let tempdir_path = tempdir.path();
+
+    let download_target = |name: TargetName| async move {
+        info!("\t-> {}", name.raw());
+        repository
+            .save_target(&name, tempdir_path, Prefix::None)
+            .await
+            .context(error::SaveTargetSnafu)
+    };
+
+    info!("Downloading targets to {tempdir_path:?}");
+    for target in target_names.clone() {
+        download_target(target).await?;
+    }
+
+    debug!("Cleaning up {outdir:?}");
+    tokio::fs::remove_dir_all(outdir)
+        .await
+        .context(error::CleanDirSnafu { path: outdir })?;
+
+    info!("Moving targets to {outdir:?}");
+    tokio::fs::create_dir_all(outdir)
+        .await
+        .context(error::CreateDirSnafu { path: outdir })?;
+    for target in target_names {
+        let mut tmpdir_target_path = PathBuf::from(tempdir.path());
+        tmpdir_target_path.push(target.raw());
+        let mut outdir_target_path = outdir.clone();
+        outdir_target_path.push(target.raw());
+        tokio::fs::rename(tmpdir_target_path, outdir_target_path)
+            .await
+            .context(error::MoveTargetSnafu)?;
+    }
+
+    tempdir.close().context(error::CloseTempDirSnafu)?;
+    Ok(())
+}
+
+/// Fetch the variant and download its image targets
+async fn fetch_variant(
+    root_role_path: &PathBuf,
+    metadata_url: Url,
+    targets_url: &Url,
+    outdir: &PathBuf,
+    filename_prefix: &str,
+    variant_manifest: &PathBuf,
+    variant: &str,
+) -> Result<(), Error> {
+    // Load the repository
+    let repo = RepositoryLoader::new(
+        &repo::root_bytes(root_role_path).await?,
+        metadata_url.clone(),
+        targets_url.clone(),
+    )
+    .load()
+    .await
+    .context(repo_error::RepoLoadSnafu {
+        metadata_base_url: metadata_url.clone(),
+    })?;
+
+    let manifest_info = ManifestInfo::new(variant_manifest).context(error::ManifestParseSnafu)?;
+
+    let image_layout = manifest_info
+        .image_layout()
+        .context(error::MissingImageLayoutSnafu { variant })?;
+    let image_format = manifest_info.image_format();
+    let image_ext = match image_format {
+        Some(ImageFormat::Raw) | None => "img.lz4",
+        Some(ImageFormat::Qcow2) => "qcow2",
+        Some(ImageFormat::Vmdk) => "ova",
+    };
+
+    let targets = match image_format {
+        // Since the OVA will contain all of the necessary VMDKs, the partition plan is irrelevant.
+        Some(ImageFormat::Vmdk) => {
+            vec![format!("{filename_prefix}.{image_ext}")]
+        }
+        _ => match image_layout.partition_plan {
+            PartitionPlan::Split => {
+                vec![
+                    format!("{filename_prefix}.{image_ext}"),
+                    format!("{filename_prefix}-data.{image_ext}"),
+                ]
+            }
+            PartitionPlan::Unified => vec![format!("{filename_prefix}.{image_ext}")],
+        },
+    };
+    handle_download(&repo, outdir, &targets).await
+}
+
+/// Common entrypoint from main()
+pub(crate) async fn run(args: &Args, fetch_variant_args: &FetchVariantArgs) -> Result<(), Error> {
+    let infra_config =
+        InfraConfig::from_path(&args.infra_config_path).context(repo_error::ConfigSnafu)?;
+    trace!("Parsed infra config: {:?}", infra_config);
+    let repo_config = infra_config
+        .repo
+        .as_ref()
+        .context(repo_error::MissingConfigSnafu {
+            missing: "repo section",
+        })?
+        .get(&fetch_variant_args.repo)
+        .context(repo_error::MissingConfigSnafu {
+            missing: format!("definition for repo {}", &fetch_variant_args.repo),
+        })?;
+
+    let repo_urls = repo_urls(
+        repo_config,
+        &fetch_variant_args.variant,
+        &fetch_variant_args.arch,
+    )?
+    .context(repo_error::MissingRepoUrlsSnafu {
+        repo: &fetch_variant_args.repo,
+    })?;
+
+    let version_full = format!(
+        "{}-{}",
+        &fetch_variant_args.version, &fetch_variant_args.build
+    );
+    let mut versioned_outdir = fetch_variant_args.outdir.clone();
+    versioned_outdir.push(version_full);
+
+    fetch_variant(
+        &fetch_variant_args.root_role_path,
+        repo_urls.0,
+        repo_urls.1,
+        &versioned_outdir,
+        &fetch_variant_args.filename_prefix,
+        &fetch_variant_args.variant_manifest,
+        &fetch_variant_args.variant,
+    )
+    .await
+}
+
+mod error {
+    use snafu::Snafu;
+    use std::io;
+    use std::path::PathBuf;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub(crate)))]
+    pub(crate) enum Error {
+        #[snafu(display("Failed to clean directory '{}': {}", path.display(), source))]
+        CleanDir {
+            path: PathBuf,
+            source: io::Error,
+        },
+
+        #[snafu(display("Failed to delete temporary directory: {}", source))]
+        CloseTempDir {
+            source: io::Error,
+        },
+
+        #[snafu(display("Failed to create directory '{}': {}", path.display(), source))]
+        CreateDir {
+            path: PathBuf,
+            source: io::Error,
+        },
+
+        #[snafu(display("Failed to create temporary directory: {}", source))]
+        CreateTempDir {
+            source: io::Error,
+        },
+
+        #[snafu(display("Invalid target name: {}", source))]
+        InvalidTargetName {
+            source: tough::error::Error,
+        },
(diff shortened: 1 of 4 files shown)


tools/pubsys/src/repo/fetch_variant/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    let image_ext = match image_format {
+        Some(ImageFormat::Raw) | None => "img.lz4",
+        Some(ImageFormat::Qcow2) => "qcow2",
+        Some(ImageFormat::Vmdk) => "ova",
+    };
+
+    // TODO - remove this after testing
+    let clean_filename_prefix = filename_prefix
+        .strip_suffix("-dirty")
+        .unwrap_or(filename_prefix);
```
jpculp: We should leave this in here a little longer until after we get some feedback.

twoliter/embedded/Makefile.toml
```diff
@@ -1073,6 +1073,60 @@ pubsys \
...
+   \
+   --variant-manifest "${BUILDSYS_ROOT_DIR}/variants/${BUILDSYS_VARIANT}/Cargo.toml" \
+   --filename-prefix "${BUILDSYS_NAME_FULL}" \
+   --root-role-path "${PUBLISH_REPO_ROOT_JSON}" \
+   \
+   --outdir "${BUILDSYS_VARIANT_DIR}"\
+'''
+]
+
+[tasks.fetch-ova]
```
jpculp: This could also be `fetch-friendly`, but I imagine `fetch-ova` is more intuitive.

twoliter/embedded/Makefile.toml
```diff
@@ -1073,6 +1073,60 @@ pubsys \
...
+   --infra-config-path "${PUBLISH_INFRA_CONFIG_PATH}" \
+   \
+   fetch-variant \
+   \
+   --repo "${PUBLISH_REPO}" \
+   --arch "${BUILDSYS_ARCH}" \
+   --variant "${BUILDSYS_VARIANT}" \
+   \
+   --variant-manifest "${BUILDSYS_ROOT_DIR}/variants/${BUILDSYS_VARIANT}/Cargo.toml" \
+   --filename-prefix "${BUILDSYS_NAME_FRIENDLY}" \
```
bcressey: Do we not publish OVAs with `BUILDSYS_NAME_FULL`? 😞 

Please implement this the way that `tasks.publish-setup-without-key` and `tasks.publish-setup` work, where you set an environment variable and then run the parent task. That will make it clear what the difference between these two nearly identical tasks is.

tools/pubsys/src/repo/fetch_variant/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    let image_format = manifest_info.image_format();
+    let image_ext = match image_format {
+        Some(ImageFormat::Raw) | None => "img.lz4",
+        Some(ImageFormat::Qcow2) => "qcow2",
+        Some(ImageFormat::Vmdk) => "ova",
+    };
+
+    // TODO - remove this after testing
+    let clean_filename_prefix = filename_prefix
+        .strip_suffix("-dirty")
```
bcressey: If we publish images from a dirty checkout into a repo, do they not retain the suffix? It doesn't seem like we should strip it.

tools/pubsys/src/repo/fetch_variant/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+        .iter()
+        .map(|s| TargetName::new(s).context(error::InvalidTargetNameSnafu))
+        .collect();
+    let target_names = target_names?;
+    let download_target = |name: TargetName| async move {
+        println!("\t-> {}", name.raw());
+        repository
+            .save_target(&name, outdir, Prefix::None)
+            .await
+            .context(error::MetadataSnafu)
```
bcressey: `MetadataSnafu` seems very generic - maybe `SaveTargetSnafu` with a message like "failed to save target {name}: {error}".

tools/pubsys/src/repo/fetch_variant/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+        },
+        #[snafu(display("Could not find image layout for {}", variant))]
+        MissingImageLayout {
+            variant: String,
+        },
+        #[snafu(context(false), display("{}", source))]
+        Repo {
+            #[snafu(source(from(crate::repo::Error, Box::new)))]
+            source: Box<crate::repo::Error>,
+        },
```
bcressey, line 278: Is this used?

tools/pubsys/src/repo/fetch_variant/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    outdir: &Path,
+    raw_names: &[String],
+) -> Result<(), Error> {
+    let target_names: Result<Vec<TargetName>, Error> = raw_names
+        .iter()
+        .map(|s| TargetName::new(s).context(error::InvalidTargetNameSnafu))
+        .collect();
+    let target_names = target_names?;
+    let download_target = |name: TargetName| async move {
+        println!("\t-> {}", name.raw());
```
bcressey: Can we use `info!` or `debug!` logging here instead of `println!`?

tools/pubsys/src/repo/fetch_variant/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+        repository
+            .targets()
+            .signed
+            .targets
+            .keys()
+            .cloned()
+            .collect()
+    } else {
+        target_names
+    };
```
bcressey: All available targets could be many hundreds of GiBs for older repos, which is a severe penalty for not passing specific names.

Why do you need that as a fallback case?

tools/pubsys/src/repo/fetch_variant/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    println!("Downloading targets to {outdir:?}");
+    tokio::fs::create_dir_all(outdir)
+        .await
+        .context(error::DirCreateSnafu { path: outdir })?;
```
bcressey: Use logging statements. Also, I think the caller should create the output directory - it's a bit strange to call the function with a non-existent path.

tools/pubsys/src/repo/fetch_variant/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    // TODO - remove this after testing
+    let clean_filename_prefix = filename_prefix
+        .strip_suffix("-dirty")
+        .unwrap_or(filename_prefix);
+
+    match image_format {
+        Some(ImageFormat::Vmdk) => {
+            let target = [format!("{filename_prefix}.{image_ext}")];
+            handle_download(&repo, outdir, &target).await
+        }
```
bcressey: factor out the common code path:
```suggestion
    let targets = match image_format {
        Some(ImageFormat::Vmdk) => {
            [format!("{filename_prefix}.{image_ext}")]
        }
        ...
    };
    handle_download(&repo, outdir, &targets).await
```

tools/pubsys/src/repo/fetch_variant/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+        Some(ImageFormat::Vmdk) => "ova",
+    };
+
+    // TODO - remove this after testing
+    let clean_filename_prefix = filename_prefix
+        .strip_suffix("-dirty")
+        .unwrap_or(filename_prefix);
+
+    match image_format {
+        Some(ImageFormat::Vmdk) => {
```
bcressey, line 152: Add a comment about why vmdk is special (because the OVA will contain one or both images depending on the partition plan).

tools/pubsys/src/repo/fetch_variant/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    .await
+    .context(repo_error::RepoLoadSnafu {
+        metadata_base_url: metadata_url.clone(),
+    })?;
+
+    // To help ensure that downloads are safe, we require that the outdir does not exist.
+    ensure!(
+        !outdir.exists(),
+        error::DownloadOutdirExistsSnafu { path: outdir }
+    );
```
bcressey: I would recommend downloading to a temporary directory that you create, then copying the artifacts that match the expected prefix to the final output directory.

That will make it possible to re-run this step repeatedly and get the same result, rather than having to manually clean up directories between invocations.

bcressey (commented):
Overall this seems reasonable, but the presence of what I hope are just debug code paths - stripping "dirty", falling back to downloading all targets - makes it hard to understand if this is really working as expected.

tools/pubsys/src/repo/fetch_variant/mod.rs
jpculp: Let me check if `info!` works. This function was copied verbatim from `tuftool` and the print is a nice status bar.

tools/pubsys/src/repo/fetch_variant/mod.rs
jpculp: This is another piece taken directly from `tuftool`. I don't mind improving it, but we might want to make the same change on the `tuftool` side for consistency.

tools/pubsys/src/repo/fetch_variant/mod.rs
jpculp: Since this is related to `repack-variant`, I've been testing by downloading the our clean 1.20 images. The `-dirty` stuff will be removed after I test the changes related to feedback.

jpculp:
Rebased before addressing feedback.

tools/pubsys/src/repo/fetch_variant/mod.rs
jpculp: I don't mind writing to a temporary directory first, but we might want to keep this check to avoid the situation where someone fetches a variant that they already built locally (unless we want to clean up the directory as part of the fetch).

tools/pubsys/src/repo/fetch_variant/mod.rs
gthao313: Yes. We are using couple `repo_error` here https://github.com/bottlerocket-os/twoliter/pull/236/files#diff-c83b8ab8d06c1d4c94e10542931d1b6e3f3e99f9ee3bfeea50577502481844deR160-R170

jpculp:
Addressed feedback.

tools/pubsys/src/repo/fetch_variant/mod.rs
```diff
@@ -0,0 +1,265 @@
...
+    info!("Moving targets to {outdir:?}");
+    tokio::fs::create_dir_all(outdir)
+        .await
+        .context(error::CreateDirSnafu { path: outdir })?;
+    for tempdir_direntry in WalkDir::new(outdir) {
+        let tempdir_target = tempdir_direntry.context(error::DirectoryWalkSnafu)?;
+        tokio::fs::copy(tempdir_target.path(), outdir)
+            .await
+            .context(error::CopyTargetSnafu)?;
+    }
```
jpculp: Something isn't quite right here:
```shell
Failed to fetch variant: Failed to copy target: the source path is neither a regular file nor a symlink to a regular file
```

rpkelly: This is actually just a call to `asyncify(|| std::fs::copy(from, to)).await` and `std::fs::copy` throws that error

rpkelly: Since we're already walking through the tempdir, I'd assume this means we have more directories inside it that we need to recursively copy.

jpculp: I was silly. We don't need to walkdir because we already know what these filenames are.

jpculp:
- Fixed copies from the tempdir to the outdir.
- Removed code for testing on `-dirty` trees.

tools/pubsys/src/repo/fetch_variant/mod.rs
```diff
@@ -0,0 +1,263 @@
...
+    /// Where to store the downloaded img files
+    outdir: PathBuf,
+
+    #[arg(long)]
+    /// The variant name without extension
+    filename_prefix: String,
+
+    #[arg(long)]
+    /// The root directory of the buildsys
+    variant_manifest: PathBuf,
```
webern: Is this documentation correct?

tools/pubsys/src/repo/fetch_variant/mod.rs
```diff
@@ -0,0 +1,263 @@
...
+        info!("\t-> {}", name.raw());
+        repository
+            .save_target(&name, tempdir_path, Prefix::None)
+            .await
+            .context(error::SaveTargetSnafu)
+    };
+
+    info!("Downloading targets to {tempdir_path:?}");
+    for target in target_names.clone() {
+        download_target(target).await?;
```
webern, line 90: Is this happening in parallel? If not, maybe we can follow up with an optimization.

tools/pubsys/src/repo/fetch_variant/mod.rs
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    info!("Cleaning up {outdir:?}");
```
webern: nit: maybe it's a little verbose?

```suggestion
    debug!("Cleaning up {outdir:?}");
```

tools/pubsys/src/repo/fetch_variant/mod.rs
```diff
@@ -0,0 +1,263 @@
...
+            PartitionPlan::Unified => vec![format!("{filename_prefix}.{image_ext}")],
+        },
+    };
+    handle_download(&repo, outdir, &targets).await
+}
+
+/// Common entrypoint from main()
+pub(crate) async fn run(args: &Args, fetch_variant_args: &FetchVariantArgs) -> Result<(), Error> {
+    // If a lock file exists, use that, otherwise use Infra.toml
+    let infra_config = InfraConfig::from_path_or_lock(&args.infra_config_path, false)
```
webern: Hmm, Infra.lock isn't a thing anymore. But following the pattern is fine to be cleaned up later.

webern (approved):

tools/pubsys/src/repo/fetch_variant/mod.rs
jpculp: This is how it is in `tuftool`, so if we want to optimize we might want to do it in both places.

jpculp:
Addressed @webern's feedback.

webern (approved):

tools/pubsys/src/repo/fetch_variant/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+        .map(|s| TargetName::new(s).context(error::InvalidTargetNameSnafu))
+        .collect();
+    let target_names = target_names?;
+
+    if target_names.is_empty() {
+        return error::MissingTargetNamesSnafu.fail();
+    };
+
+    let tempdir = tempdir().context(error::CreateTempDirSnafu)?;
+    let tempdir_path = tempdir.path();
```
bcressey: We should create this in `build/.tmp` or something in case the system's tmpdir is tmpfs-backed or otherwise not prepared for multi-gigabyte files

jpculp:
To ensure the temp directory was in the same filesystem as the output directory, the outdir will be the variant level and version and build are passed to create a versioned_outdir, similar to `rpm2img` and `img2img`. This also ensures that we can safely delete the contents of the outdir if `pubsys` is called outside of the makefile target.

rpkelly (approved):

bcressey (approved):

PR #243, opened 2024-05-28
imghelper: check full output of `pesign -l`

jpculp:
**Description of changes:**

Rather than assuming listing signatures with `pesign` will only return `"No signatures found."`, check if the output of the command contains that string.

**Testing done:**

Built bottlerocket using an AWS profile for secure boot signing. (thanks, @bcressey!)

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/imghelper
+++ b/twoliter/embedded/imghelper
@@ -478,7 +478,7 @@ sbsetup_signing_profile() {
 undo_sign() {
   local what
   what="${1:?}"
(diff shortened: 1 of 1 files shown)


bcressey (approved):

bcressey:
Thanks. Seems like this is required because we set `export AWS_KMS_PKCS11_DEBUG=1` - which IMO is still worth keeping for improved debugging.

sumukhballal (approved):

Issue #207, opened 2024-05-07
release v0.2.0


webern:
Release Twoliter version v0.2.0.

- [ ] #205
- [ ] #198 
- [ ] Changelog
- [ ] Test in Bottlerocket (with necessary changes and PR)
- [ ] Test with another OOTB project (with necessary changes and PR)

PR #244, opened 2024-05-29
improve backwards compatibility for existing projects

bcressey:
**Issue number:**

Related: #207

**Description of changes:**
The changes queued up amount to flag day where existing projects have to make structural changes and clean existing build artifacts. This is an annoyance for local development, but could be a bigger headache if past point release branches need to move to a newer version of `twoliter` for any reason.

Try to ease some of the pain in the transition by making a few concessions to backwards compatibility:

1. Use existing RPMs during builds if they exist, but clean them up if the corresponding package is rebuilt so the project converges on the new structure over time.
2. Keep the old target directory that cargo uses to track what's been built, but move it to the new location.
3. Support the workspace manifest in either the old location, under `variants`, or the new location at the project root.
4. Treat manifests with no buildsys-related metadata as "packages" (rather than "variants" or "kits").

**Testing done:**
Built the main project repo with the previous release of `twoliter`, updated to a local build of `twoliter` with these changes, and confirmed that builds and cleanup worked as expected. 

Output showing warnings:
```
[cargo-make] INFO - cargo make 0.36.11
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: default
[cargo-make] INFO - Running Task: install-twoliter
Found Twoliter v0.1.1 installed.
Skipping installation.
[cargo-make] INFO - Execute Command: "/home/fedora/brdev/tools/twoliter/twoliter" "--log-level=info" "make" "default" "--project-path=/home/fedora/brdev/Twoliter.toml" "--cargo-home=/home/fedora/brdev/.cargo" "--"
[2024-05-29T17:22:16Z WARN  twoliter::project] A Release.toml file was found. Release.toml is deprecated. Please remove it from your project.
[cargo-make][1] INFO - Build File: /home/fedora/brdev/build/tools/Makefile.toml
[cargo-make][1] INFO - Task: default
[cargo-make][1] INFO - Profile: development
[cargo-make][1] INFO - Running Task: setup
[cargo-make][1] INFO - Running Task: setup-build
[cargo-make][1] INFO - Running Task: fetch-sdk
[cargo-make][1] INFO - Running Task: fetch-sources
[cargo-make][1] INFO - Running Task: fetch-vendored
[cargo-make][1] INFO - Running Task: check-licenses
bans ok, licenses ok, sources ok
[cargo-make][1] INFO - Running Task: fetch-licenses
Skipping fetching licenses
[cargo-make][1] INFO - Running Task: build-sbkeys
[cargo-make][1] INFO - Running Task: publish-setup
17:22:21 [INFO] Found infra config at path: /home/fedora/brdev/Infra.toml
[cargo-make][1] INFO - Running Task: cargo-metadata
Found deprecated project workspace /home/fedora/brdev/variants/Cargo.toml.
Please relocate it to /home/fedora/brdev/Cargo.toml.
[cargo-make][1] INFO - Running Task: build-variant
   Compiling libtirpc v0.1.0 (/home/fedora/brdev/packages/libtirpc)
   Compiling aws-signing-helper v0.1.0 (/home/fedora/brdev/packages/aws-signing-helper)
   Compiling static-pods v0.1.0 (/home/fedora/brdev/packages/static-pods)
   Compiling ecr-credential-provider-1_27 v0.1.0 (/home/fedora/brdev/packages/ecr-credential-provider-1.27)
   Compiling os v0.1.0 (/home/fedora/brdev/packages/os)
warning: netdog@0.1.0: Expected to find one of 'build-package', 'build-kit', or 'build-variant' in package.metadata. Assuming 'build-package'.
warning: netdog@0.1.0: Expected to find one of 'build-package', 'build-kit', or 'build-variant' in package.metadata. Assuming 'build-package'.
   Compiling kmod-6_1-nvidia v0.1.0 (/home/fedora/brdev/packages/kmod-6.1-nvidia)
...
```

**Terms of contribution:**

### DIFF
--- a/tools/buildsys/src/builder.rs
+++ b/tools/buildsys/src/builder.rs
@@ -303,7 +303,7 @@ pub(crate) struct DockerBuild {
     target: String,
     tag: String,
     root_dir: PathBuf,
-    artifacts_dir: PathBuf,
+    artifacts_dirs: Vec<PathBuf>,
     state_dir: PathBuf,
     artifact_name: String,
     common_build_args: CommonBuildArgs,
@@ -320,6 +320,7 @@ impl DockerBuild {
     ) -> Result<Self> {
         let package = manifest.info().package_name();
         let per_package_dir = format!("{}/{}", args.packages_dir.display(), package).into();
+        let old_package_dir = format!("{}", args.packages_dir.display()).into();
 
         Ok(Self {
             dockerfile: args.common.tools_dir.join("build.Dockerfile"),
@@ -334,7 +335,7 @@ impl DockerBuild {
                 &args.common.root_dir,
             ),
             root_dir: args.common.root_dir.clone(),
-            artifacts_dir: per_package_dir,
+            artifacts_dirs: vec![per_package_dir, old_package_dir],
             state_dir: args.common.state_dir,
             artifact_name: package.to_string(),
             common_build_args: CommonBuildArgs::new(
@@ -375,7 +376,7 @@ impl DockerBuild {
                 &args.common.root_dir,
             ),
             root_dir: args.common.root_dir.clone(),
-            artifacts_dir: per_kit_dir,
+            artifacts_dirs: vec![per_kit_dir],
             state_dir: args.common.state_dir,
             artifact_name: kit.to_string(),
             common_build_args: CommonBuildArgs::new(
@@ -418,7 +419,7 @@ impl DockerBuild {
                 &args.common.root_dir,
             ),
             root_dir: args.common.root_dir.clone(),
-            artifacts_dir: args.common.image_arch_variant_dir,
+            artifacts_dirs: vec![args.common.image_arch_variant_dir],
             state_dir: args.common.state_dir,
             artifact_name: args.variant.clone(),
             common_build_args: CommonBuildArgs::new(
@@ -498,7 +499,7 @@ impl DockerBuild {
                 &args.common.root_dir,
             ),
             root_dir: args.common.root_dir.clone(),
-            artifacts_dir: args.common.image_arch_variant_dir,
+            artifacts_dirs: vec![args.common.image_arch_variant_dir],
             state_dir: args.common.state_dir,
             artifact_name: args.variant.clone(),
             common_build_args: CommonBuildArgs::new(
@@ -549,7 +550,7 @@ impl DockerBuild {
         // Clean up any previous outputs we have tracked.
         match self.common_build_args.cleanup {
             OutputCleanup::BeforeBuild => {
-                clean_build_files(&marker_dir, &self.artifacts_dir)?;
+                clean_build_files(&marker_dir, &self.artifacts_dirs)?;
             }
             OutputCleanup::None => (),
         }
@@ -608,7 +609,7 @@ impl DockerBuild {
         docker(&rmi, Retry::No)?;
 
         // Copy artifacts to the expected directory and write markers to track them.
-        copy_build_files(&marker_dir, &self.artifacts_dir)?;
+        copy_build_files(&marker_dir, &self.artifacts_dirs[0])?;
 
         Ok(())
     }
@@ -807,17 +808,16 @@ where
     Ok(())
 }
 
-/// Remove build artifacts from the output directory.
+/// Remove build artifacts from any of the known output directories.
 /// Any marker file we find could have a corresponding file that should be cleaned up.
 /// We also clean up the marker files so they do not accumulate across builds.
 /// For the same reason, if a directory is empty after build artifacts, marker files, and other
 /// empty directories have been removed, then that directory will also be removed.
-fn clean_build_files<P>(build_dir: P, output_dir: P) -> Result<()>
+fn clean_build_files<P>(build_dir: P, output_dirs: &[PathBuf]) -> Result<()>
 where
     P: AsRef<Path>,
 {
     let build_dir = build_dir.as_ref();
-    let output_dir = output_dir.as_ref();
 
     fn has_markers(entry: &DirEntry) -> bool {
         let is_dir = entry.path().is_dir();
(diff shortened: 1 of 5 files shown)


tools/buildsys/src/builder.rs
```diff
@@ -859,16 +859,18 @@ where
...
+            let mut output_file: PathBuf = output_dir.into();
+            output_file.push(marker_file.strip_prefix(build_dir).context(
+                error::StripPathPrefixSnafu {
+                    path: &marker_file,
+                    prefix: build_dir,
+                },
+            )?);
+            output_file.set_extension("");
+            cleanup(&output_file, output_dir, &mut clean_dirs)?;
+            cleanup(&marker_file, build_dir, &mut clean_dirs)?;
```
webern: Is it a problem that this is inside the loop and being called more than once for each marker file since there is now more than one `output_dir/output_file`?

twoliter/embedded/Makefile.toml
```diff
@@ -329,11 +329,10 @@ dependencies = ["setup"]
 script_runner = "bash"
 script = [
 '''
-if [ -f "${BUILDSYS_SOURCES_DIR}/Cargo.toml" ]; then
-  cargo fetch --locked --manifest-path "${BUILDSYS_SOURCES_DIR}/Cargo.toml"
-fi
-
-cargo fetch --locked --manifest-path "${BUILDSYS_ROOT_DIR}/Cargo.toml"
```
webern, line 336: Edit: (I missed the `.` associated with this commend)

~We still need this or else we are breaking the ability to go to one workspace. Are you saying we will no longer have one workspace as a possibility? It would need to be something like:~

```bash
if [ -f "${BUILDSYS_ROOT_DIR}/Cargo.toml" ]; then
  cargo fetch --locked --manifest-path "${BUILDSYS_ROOT_DIR}/Cargo.toml"
fi
```

twoliter/embedded/Makefile.toml
```diff
@@ -329,11 +329,10 @@ dependencies = ["setup"]
...
+for ws in sources variants .; do
+  [ -s "${BUILDSYS_ROOT_DIR}/${ws}/Cargo.toml" ] || continue
```
webern, line 333: Why is this `-s` and not `-f`?

webern (commented):
Don't you need to remove `[tasks.check-rpm-structure]` (line 782 in the current version of your file)?

twoliter/embedded/Makefile.toml
```diff
@@ -829,6 +856,17 @@ export BUILDSYS_VARIANT_RUNTIME="${BUILDSYS_VARIANT_RUNTIME:?}"
 export BUILDSYS_VARIANT_FAMILY="${BUILDSYS_VARIANT_FAMILY:?}"
 export BUILDSYS_VARIANT_FLAVOR="${BUILDSYS_VARIANT_FLAVOR}"
 
+# Relocate an existing target directory to the new location, to avoid breaking
+# compatibility on upgrade. On downgrade, the missing directory will trigger a
+# full rebuild, which is good because the new layout of RPM artifacts is not
+# compatible with previous versions.
+if [ -d "${BUILDSYS_ROOT_DIR}/variants/target/${BUILDSYS_ARCH}" ]; then
+  mkdir -p "${BUILDSYS_ROOT_DIR}/target"
+  mv \
```
webern, line 837: Don't you also need to check if `"${BUILDSYS_ROOT_DIR}/Cargo.toml` exists?

twoliter/embedded/Makefile.toml
```diff
@@ -1573,7 +1611,19 @@ rm -f ${BUILDSYS_TOOLS_DIR}/bin/*
 script_runner = "bash"
 script = [
 '''
-cargo clean --manifest-path "${BUILDSYS_ROOT_DIR}/Cargo.toml"
+# The workspace manifest could be the old one under variants, or the new one at
+# the project root.
+for ws in variants .; do
```
webern, line 1588: Oh, I see, I think I missed the `.` in a previous comment.

bcressey:
> Don't you need to remove [tasks.check-rpm-structure] (line 782 in the current version of your file)?

I can, but the `check-rpm-structure` task is never called, so we could also leave it alone for when we do want to make this a breaking change.

twoliter/embedded/Makefile.toml
bcressey: I usually prefer `-s` over `-f` for cases where a non-empty file is expected. `cargo` doesn't like empty manifests.

tools/buildsys/src/builder.rs
bcressey: > Is it a problem that this is inside the loop and being called more than once for each marker file since there is now more than one output_dir/output_file?

I'll fix that. It's not directly a problem since `cleanup` returns immediately if the file doesn't exist, but it would mean that this sequence breaks:
1. output file doesn't exist in first directory
2. marker file is cleaned up in first iteration
3. output file exists in next directory, but removal fails with an error

The second time the program is run, the marker file would be gone and so the attempt to clean up the output would not be repeated.

webern (approved):

webern:
> > Don't you need to remove [tasks.check-rpm-structure] (line 782 in the current version of your file)?
> 
> I can, but the `check-rpm-structure` task is never called, so we could also leave it alone for when we do want to make this a breaking change.

Maybe we've headed that off by introducing a backwards compatible version and the `BUILDSYS_OUTPUT_GENERATION_ID` mechanism for a future break.

twoliter/embedded/Makefile.toml
bcressey: > Don't you also need to check if "${BUILDSYS_ROOT_DIR}/Cargo.toml exists?

This works regardless of where the manifest workspace is located - it just relocates the targets output directory. That helps with downgrade compatibility since then `cargo` won't find it with an older `twoliter` release and will rebuild everything.

bcressey:
⬆️ switches from `ln *.rpm` to `find` to fix the case when one of the directories might be empty. `find` also lets us print out the matching RPMs, which is handy for troubleshooting.

webern (approved):

jpculp (approved):

PR #250, opened 2024-05-30
twoliter: lock resolution algorithm

### LINKED ISSUE #256: twoliter: lock creation and resolution
We need to write an algorithm that given a Twoliter.toml will resolve all needed kit dependencies and generate a Twoliter.lock

We should also update the twoliter commands that need sdk or kits to use the lock file instead of Project file alone. 

We should also create a `twoliter update` command that the user can use to update their lockfile

jmt-lab:
**Issue number:** 256

Closes #256 

**Description of changes:**

This implements the ability for the lock file to be resolved from a Project Twoliter.toml file. It uses essentially a bfs algorithm to resolve dependencies and verify sdk compatibility.

**Testing done:**
This algorithm was tested in the PoC but I need to get some unit tests written. Putting this PR up as a draft so it can be looked at early

**Terms of contribution:**

### DIFF
--- a/twoliter/Cargo.toml
+++ b/twoliter/Cargo.toml
@@ -13,6 +13,7 @@ exclude = ["/design", "/target", "/dockerfiles", "/scripts"]
 anyhow = "1"
 async-recursion = "1"
 async-walkdir = "1"
+base64 = "0.22"
 clap = { version = "4", features = ["derive", "env", "std"] }
 env_logger = "0.11"
 filetime = "0.2"
--- a/twoliter/src/cmd/mod.rs
+++ b/twoliter/src/cmd/mod.rs
@@ -2,10 +2,12 @@ mod build;
 mod build_clean;
 mod debug;
 mod make;
+mod update;
 
 use self::build::BuildCommand;
 use crate::cmd::debug::DebugAction;
 use crate::cmd::make::Make;
+use crate::cmd::update::Update;
 use anyhow::Result;
 use clap::Parser;
 use env_logger::Builder;
@@ -35,6 +37,9 @@ pub(crate) enum Subcommand {
 
     Make(Make),
 
+    /// Update Twoliter.lock
+    Update(Update),
+
     /// Commands that are used for checking and troubleshooting Twoliter's internals.
     #[clap(subcommand)]
     Debug(DebugAction),
@@ -45,6 +50,7 @@ pub(super) async fn run(args: Args) -> Result<()> {
     match args.subcommand {
         Subcommand::Build(build_command) => build_command.run().await,
         Subcommand::Make(make_args) => make_args.run().await,
+        Subcommand::Update(update_args) => update_args.run().await,
         Subcommand::Debug(debug_action) => debug_action.run().await,
     }
 }
--- a/twoliter/src/lock.rs
+++ b/twoliter/src/lock.rs
@@ -1,6 +1,18 @@
+use crate::common::fs::{remove_file, write};
+use crate::project::{Image, Project, Vendor};
 use crate::schema_version::SchemaVersion;
+use anyhow::{ensure, Context, Result};
+use base64::Engine;
 use semver::Version;
 use serde::{Deserialize, Serialize};
+use std::collections::{HashMap, HashSet};
+use std::fmt::Display;
+use std::hash::{Hash, Hasher};
+use std::mem::take;
+use tokio::fs::read_to_string;
+use tokio::process::Command;
+
+const TWOLITER_LOCK: &str = "Twoliter.lock";
 
 /// Represents the structure of a `Twoliter.lock` lock file.
 #[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize)]
@@ -14,6 +26,8 @@ pub(crate) struct Lock {
     pub sdk: LockedImage,
     /// Resolved kit dependencies
     pub kit: Vec<LockedImage>,
+    /// sha256 digest of the Project this was generated from
+    pub digest: String,
 }
 
 /// Represents a locked dependency on an image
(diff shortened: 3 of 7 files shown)


twoliter/src/lock.rs
```diff
@@ -28,3 +34,177 @@ pub(crate) struct LockedImage {
...
+            sdk: sdk.context("could not resolve a sdk version from dependencies")?,
+        })
+    }
+
+    async fn find_kit(vendor: &Vendor, image: &LockedImage) -> Result<ImageMetadata> {
+        // First construct the initial image-uri to check, do not rely on image.source as it is currently in a shape
+        // for hash comparisons
+        let image_uri = format!("{}/{}:v{}", vendor.registry, image.name, image.version);
+        // We want to inspect the image
+        let inspect_output = docker_command([
```
jmt-lab: This will need to be tweaked to run first a manifest inspect followed by the image inspect, since the image config for the manifest list is not always populated in the cache

twoliter/src/docker/commands.rs
```diff
@@ -1 +1,19 @@
+use anyhow::{Context, Result};
+use std::ffi::OsStr;
+use std::process::Output;
+use tokio::process::Command;
 
+/// Runs a docker command with full error checking on both launch of command and
+/// result of the command.
+#[allow(dead_code)]
```
bcressey: This doesn't appear to check the result of the command - all of the callers are inspecting the status code.

twoliter/src/docker/commands.rs
```diff
@@ -1 +1,19 @@
...
+where
+    I: IntoIterator<Item = T>,
+    T: AsRef<OsStr>,
+{
+    Command::new("docker")
+        .args(args)
+        .output()
+        .await
+        .context("failed to use docker cli")
+}
```
bcressey: This might be better as a macro where we combine the docker invocation with the error to return if the command fails. As it stands, it isn't doing very much.

twoliter/src/lock.rs
```diff
@@ -28,3 +38,234 @@ pub(crate) struct LockedImage {
...
+                return Ok(lock);
+            }
+        }
+        let lock = Self::resolve(&project).await?;
+        let lock_str = toml::to_string_pretty(&lock).context("failed to serialize lock file")?;
+        write(&lock_file_path, lock_str)
+            .await
+            .context("failed to write lock file")?;
+        Ok(lock)
+    }
```
bcressey: I'm not sure where you want to fix this, but it's problematic to assume that it's always OK to regenerate the lock file.

Much of the point of the lock file is that it can be checked into source control and bind the build to exactly the locked versions and error out if those exact versions are no longer available. If it automatically unlocks and updates itself then it is not enforcing that property.

It would be safer for this to assert that the digest in the lock file matches the project's digest, and then provide another command to regenerate the lock file in a controlled way.

Alternatively you could support flags like cargo's `--locked` or `--frozen`, but the behavior of respecting the lock is the critical one, and the behavior of automatically updating the lock is mostly for developer ease of use.

twoliter/src/lock.rs
```diff
@@ -28,3 +38,234 @@ pub(crate) struct LockedImage {
...
+                .context("failed to read lockfile")?;
+            let lock: Self =
+                toml::from_str(lock_str.as_str()).context("failed to deserialize lockfile")?;
+            // If the digests match we don't need to re-resolve the lockfile
+            if lock.digest == project.digest()? {
+                return Ok(lock);
+            }
+        }
+        let lock = Self::resolve(&project).await?;
+        let lock_str = toml::to_string_pretty(&lock).context("failed to serialize lock file")?;
```
bcressey: aesthetic nit: not sure what `toml::to_string_pretty` does but I would prefer not to have `Twoliter.lock` use that ghastly format where nested tables are indented under the parent table.

twoliter/src/lock.rs
```diff
@@ -28,3 +38,234 @@ pub(crate) struct LockedImage {
...
+            let sdk_vendor = vendor_table.get(&sdk.vendor).context(format!(
+                "vendor '{}' was not specified in Twoliter.toml, and is required for build",
+                sdk.vendor
+            ))?;
+            Some(LockedImage {
+                name: sdk.name.clone(),
+                version: sdk.version.clone(),
+                vendor: sdk.vendor.clone(),
+                source: format!("{}:{}/v{}", sdk_vendor.registry, sdk.name, sdk.version),
+            })
```
bcressey: The algorithm here could be cleaned up by unifying the treatment of kits and SDKs:
```
create a set of known vendors
create an empty map of known (name, vendor) -> version
create an empty vec of locked images

create a vec of images to resolve from the list of kits
if there's an SDK:
    add the SDK to the vec of images to resolve

while the vec of images to resolve is not empty:
    copy the vec of images to resolve to the working vec
    for image in working vec:
        if an entry for this image already exists in the version map:
            assert that the version is the same
            continue with the next image

        assert that the vendor is in the set of known vendors
        add the (name, vendor) -> version entry to the version map
        add the LockedImage to the vec of locked images

        add the image's sdk to the vec of images to resolve
        for each image dependency:
            add the dependency to the vec of images to resolve

finally:
    # error: must have an SDK
    assert that there's at least one SDK in the vec of locked images

    # error: cannot have more than one SDK
    assert that there's exactly one SDK in the vec of locked images
```

You are mostly doing this already, but you are not checking for cases where the same kit name/vendor is added at different versions, and the vendor lookup can be optimized.

twoliter/src/lock.rs
```diff
@@ -28,3 +38,234 @@ pub(crate) struct LockedImage {
...
+        // First construct the initial image-uri to check, do not rely on image.source as it is currently in a shape
+        // for hash comparisons
+        let image_uri = format!("{}/{}:v{}", vendor.registry, image.name, image.version);
+
+        // Now inspect the manifest list
+        let manifest_output = docker_command(["manifest", "inspect", image_uri.as_str()]).await?;
+        // If this hit an error the kit does not exist
+        ensure!(
+            manifest_output.status.success(),
+            "no such kit foud at uri {}",
```
bcressey: nit:
```suggestion
            "no such kit found at uri {}",
```

but also you should indicate the kit's name / vendor / version so the error message is more clear about why we checked that particular URI

twoliter/src/lock.rs
```diff
@@ -28,3 +38,234 @@ pub(crate) struct LockedImage {
...
+            release_version: project.release_version().to_string(),
+            kit: kit.iter().cloned().collect(),
+            sdk: sdk.context("could not resolve a sdk version from dependencies")?,
+            digest: project.digest()?,
+        })
+    }
+
+    async fn find_kit(vendor: &Vendor, image: &LockedImage) -> Result<ImageMetadata> {
+        // First construct the initial image-uri to check, do not rely on image.source as it is currently in a shape
+        // for hash comparisons
```
bcressey: I don't understand this comment - it seems weird that we can't use the image's source field and are instead recreating it here.

twoliter/src/lock.rs
```diff
@@ -28,3 +38,234 @@ pub(crate) struct LockedImage {
...
+            manifest_output.status.success(),
+            "no such kit foud at uri {}",
+            image_uri
+        );
+        let manifest_list: ManifestListView =
+            serde_json::from_slice(manifest_output.stdout.as_slice())
+                .context("failed to deserialize manifest list")?;
+        let image_digest = manifest_list
+            .manifests
+            .first()
```
bcressey: Are we guaranteed that the element of interest is always the first? Can we check some structural property of the element instead?

twoliter/src/lock.rs
```diff
@@ -28,3 +38,234 @@ pub(crate) struct LockedImage {
...
+
+        // Now we want to fetch the metadata from the OCI image config
+        let inspect_output = docker_command([
+            "image",
+            "inspect",
+            image_uri.as_str(),
+            "--format \"{{ json .Config.Labels }}\"",
+        ])
+        .await?;
+        // If this error'd we know that it failed to find any matching the exact tag, this can occur when
```
bcressey: "this can occur when ...?" when what?

twoliter/src/cmd/mod.rs
```diff
@@ -35,6 +37,9 @@ pub(crate) enum Subcommand {
 
     Make(Make),
 
+    /// Update Twoliter.lock
+    Update(Update),
```
webern, line 41: @bcressey do we want to use `twoliter update` for this purpose? Are there other things we might update in the future? If so should we have this be `twoliter update lock`?

twoliter/src/project.rs
```diff
@@ -208,6 +258,8 @@ pub(crate) struct Image {
     pub name: String,
     pub version: Version,
     pub vendor: String,
+    #[serde(skip, default)]
+    pub(crate) is_sdk: bool,
```
webern: Hmmm, I rather hate this. @jmt-lab says he's going to go back to a version of the resolve function that doesn't require this.

(if we really did need the struct member...)

```suggestion
    pub(crate) image_type: ImageType,
```

```rust
pub(crate) ImageType {
    Sdk,
    Kit,
}
```

twoliter/src/lock.rs
```diff
@@ -27,4 +37,208 @@ pub(crate) struct LockedImage {
...
+            for image in working_set.iter() {
+                if let Some(version) = known.get(&(image.name.clone(), image.vendor.clone())) {
+                    ensure!(
+                        image.version == *version,
+                        "cannot have multiple versions of the same kit ({0}-{1}@{2} != {0}-{3}@{2}",
+                        image.name,
+                        image.version,
+                        image.vendor,
+                        version
+                    );
```
bcressey: Add a `continue` here so you can unindent the rest of the loop body.

twoliter/src/lock.rs
```diff
@@ -27,4 +37,208 @@ pub(crate) struct LockedImage {
...
+            .context("failed to write lock file")?;
+        Ok(lock)
+    }
+
+    async fn resolve(project: &Project) -> Result<Self> {
+        let vendor_table = project.vendor();
+        let mut known: HashMap<(String, String), Version> = HashMap::new();
+        let mut locked: Vec<LockedImage> = Vec::new();
+
+        let mut resolve: Vec<Image> = project.kits();
```
bcressey:
```suggestion
        let mut remaining: Vec<Image> = project.kits();
```

twoliter/src/lock.rs
```diff
@@ -27,4 +37,208 @@ pub(crate) struct LockedImage {
...
+        if let Some(sdk) = project.sdk_image() {
+            let mut sdk = sdk.clone();
+            sdk.is_sdk = true;
+            resolve.push(sdk);
+        }
```
bcressey: Rather than adding an `is_sdk` field to image that's only used for this algorithm, you can just maintain a set of SDKs that you find during traversal:
```suggestion
        let mut sdks = HashSet::new();
        if let Some(sdk) = project.sdk_image() {
            sdks.push(sdk.clone());
            resolve.push(sdk.clone());
        }
```

(need to derive Hash on Image but that should be trivial)

twoliter/Cargo.toml
```diff
@@ -29,14 +30,14 @@ tar = "0.4"
...
-buildsys = { version = "0.1.0", artifact = [ "bin:buildsys", "bin:bottlerocket-variant" ], path = "../tools/buildsys" }
-pubsys = { version = "0.1.0", artifact = [ "bin:pubsys" ], path = "../tools/pubsys" }
-pubsys-setup = { version = "0.1.0", artifact = [ "bin:pubsys-setup" ], path = "../tools/pubsys-setup" }
-testsys = { version = "0.1.0", artifact = [ "bin:testsys" ], path = "../tools/testsys" }
-tuftool = { version = "0.10", artifact = [ "bin:tuftool" ] }
+buildsys = { version = "0.1.0", artifact = ["bin:buildsys", "bin:bottlerocket-variant"], path = "../tools/buildsys" }
+pubsys = { version = "0.1.0", artifact = ["bin:pubsys"], path = "../tools/pubsys" }
+pubsys-setup = { version = "0.1.0", artifact = ["bin:pubsys-setup"], path = "../tools/pubsys-setup" }
+testsys = { version = "0.1.0", artifact = ["bin:testsys"], path = "../tools/testsys" }
+tuftool = { version = "0.10", artifact = ["bin:tuftool"] }
```
webern: Ugh! Stop churning on toml formatting!

twoliter/src/lock.rs
```diff
@@ -27,4 +37,208 @@ pub(crate) struct LockedImage {
...
+        // Now inspect the manifest list
+        let manifest_bytes = docker!(
+            ["manifest", "inspect", image.source.as_str()],
+            "failed to find a kit at uri {}",
+            image.source
+        );
+        let manifest_list: ManifestListView = serde_json::from_slice(manifest_bytes.as_slice())
+            .context("failed to deserialize manifest list")?;
+        // The metadata will be identical across architecture images, so we really only need to
+        // pull the first found to get the metadata from the image
```
bcressey: What's the level of effort to add a check that all the entries are identical?

jmt-lab: we would have to do one more image inspect

twoliter/src/lock.rs
```diff
@@ -27,4 +37,208 @@ pub(crate) struct LockedImage {
...
+            kit: locked.iter().filter(|x| !x.is_sdk).cloned().collect(),
+        })
+    }
+
+    async fn find_kit(vendor: &Vendor, image: &LockedImage) -> Result<ImageMetadata> {
+        // Now inspect the manifest list
+        let manifest_bytes = docker!(
+            ["manifest", "inspect", image.source.as_str()],
+            "failed to find a kit at uri {}",
+            image.source
```
bcressey: nit: it seems slightly less magical just to do the format here:

```suggestion
            format!("failed to find a kit at uri {}", image.source)
```

jmt-lab:
Updates based off comments

twoliter/src/lock.rs
```diff
@@ -28,3 +38,230 @@ pub(crate) struct LockedImage {
...
+impl Hash for LockedImage {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.name.hash(state);
+        self.version.hash(state);
+        self.vendor.hash(state);
+    }
+}
+
+#[derive(Deserialize, Debug)]
+#[allow(dead_code)]
```
bcressey: Do we still need these `#[allow(dead_code)]` attributes?

twoliter/src/lock.rs
```diff
@@ -28,3 +38,230 @@ pub(crate) struct LockedImage {
...
+    pub(crate) async fn load(project: &Project) -> Result<Self> {
+        let lock_file_path = project.project_dir().join("Twoliter.lock");
+        if lock_file_path.exists() {
+            let lock_str = read_to_string(&lock_file_path)
+                .await
+                .context("failed to read lockfile")?;
+            let lock: Self =
+                toml::from_str(lock_str.as_str()).context("failed to deserialize lockfile")?;
+            // The digests must match, if changes are needed twoliter
+            ensure!(lock.digest == project.digest()?, "changes have occured to Twoliter.toml that require an update to Twoliter.lock, if intentional please run twoliter update");
```
bcressey:
```suggestion
            ensure!(lock.digest == project.digest()?, "changes have occurred to Twoliter.toml that require an update to Twoliter.lock, if intentional please run twoliter update");
```

twoliter/src/lock.rs
```diff
@@ -28,3 +38,230 @@ pub(crate) struct LockedImage {
...
+            return Ok(lock);
+        }
+        Self::create(project).await
+    }
+
+    pub(crate) async fn create(project: &Project) -> Result<Self> {
+        let lock_file_path = project.project_dir().join("Twoliter.lock");
```
bcressey: nit: can make this a constant:
```suggestion
        let lock_file_path = project.project_dir().join(TWOLITER_LOCK);
```

twoliter/src/lock.rs
```diff
@@ -28,3 +38,230 @@ pub(crate) struct LockedImage {
...
+
+        let mut remaining: Vec<Image> = project.kits();
+        let mut sdk_set: HashSet<Image> = HashSet::new();
+        if let Some(sdk) = project.sdk_image() {
+            remaining.push(sdk.clone());
+            sdk_set.insert(sdk.clone());
+        }
+        while !remaining.is_empty() {
+            let working_set = remaining.clone();
+            remaining = Vec::new();
```
bcressey:
```suggestion
            let working_set = remaining.drain(..).collect();
```

twoliter/src/lock.rs
```diff
@@ -28,3 +38,230 @@ pub(crate) struct LockedImage {
...
+#[allow(dead_code)]
+struct ImageMetadata {
+    /// The name of the kit
+    pub name: String,
+    /// The version of the kit
+    pub version: Version,
+    /// The required sdk of the kit,
+    pub sdk: Image,
+    /// Any dependent kits
+    pub kit: Vec<Image>,
```
bcressey: nit: should we call this field `kits` and serialize as `kit`?

twoliter/src/lock.rs
```diff
@@ -28,3 +38,230 @@ pub(crate) struct LockedImage {
...
+            kit: locked,
+        })
+    }
+
+    async fn resolve_kit(
+        vendors: &HashMap<String, Vendor>,
+        image: &Image,
+    ) -> Result<ImageMetadata> {
+        let vendor = vendors.get(&image.vendor).context(format!(
+            "no vendor '{}' specified in Twoltier.toml",
```
bcressey:
```suggestion
            "no vendor '{}' specified in Twoliter.toml",
```

bcressey (approved):
some minor issues, nothing blocking

twoliter/src/lock.rs
```diff
@@ -28,3 +42,239 @@ pub(crate) struct LockedImage {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_fmt(format_args!(
+            "{}-{}@{} ({})",
+            self.name, self.version, self.vendor, self.source,
+        ))
+    }
+}
+
+/// The hash should not contain the source to allow for collision detection
+impl Hash for LockedImage {
```
webern, line 67: I'm surprised clippy isn't bugging you about Eq and Ord being different from Hash. Seems fine? Ish?

twoliter/src/project.rs
```diff
@@ -191,6 +210,38 @@ impl Project {
         modules.sort();
         Ok(modules)
     }
+
+    /// Returns a base64 encoded sha256 hash of the contents of the Project structure.
+    /// Used for verifying if a change would occure in Twoliter.lock
+    pub(crate) fn digest(&self) -> Result<String> {
```
webern, line 216: Does this function need to be updated anytime we change or add something to `Project` ? Or is it just going to be things that we care about in `.lock`?

twoliter/src/project.rs
```diff
@@ -102,10 +105,26 @@ impl Project {
...
+    pub(crate) fn schema_version(&self) -> SchemaVersion<1> {
+        self.schema_version
+    }
+
     pub(crate) fn release_version(&self) -> &str {
         self.release_version.as_str()
     }
 
+    pub(crate) fn vendor(&self) -> BTreeMap<String, Vendor> {
+        self.vendor.clone()
```
webern: Seems a bit unidiomatic to do the clone here. Usually you return a `&` and let the caller do what they want.

twoliter/src/main.rs
```diff
@@ -1,3 +1,5 @@
+extern crate core;
+
 use crate::cmd::{init_logger, Args};
```
webern: I think `extern crate` is deprecated? Did your IDE add this? 

```suggestion
use core;
use crate::cmd::{init_logger, Args};
```

webern (approved):

twoliter/src/project.rs
jmt-lab: Primarily should be only the things we care about in .lock that would effect dependency changes, so if we make a change that we want to signal a need to regen Twoliter.lock we should change it

twoliter/src/main.rs
jmt-lab: Thats weird

jmt-lab:
Address final comments and rebased

PR #263, opened 2024-06-03
application-inventory: use RPM's Version and Release, set Epoch

### LINKED ISSUE #140: change the way application inventory is created
The build produces a JSON where packages are represented like this:

We need to consider what this looks like when some packages are coming from the OOTB project and some from kits (or the alpha SDK). Basically we need to track down how this produced and decide how it needs to look based on downstream requirements.

```json
    {
      "Name": "systemd",
      "Publisher": "Bottlerocket",
      "Version": "1.18.0",
      "Release": "ad294c88-dirty",
      "InstalledTime": "2024-01-24T22:00:21Z",
      "ApplicationType": "Unspecified",
      "Architecture": "aarch64",
      "Url": "https://www.freedesktop.org/wiki/Software/systemd",
      "Summary": "System and Service Manager"
    }
```

ginglis13:
**Issue number:**

Closes #140 

**Description of changes:**

Before the onset of kits, Bottlerocket would build RPMs with Release values set to the commit of Bottlerocket being released (or, more accurately, the commit of Bottlerocket being built). This worked fine under the monorepo model, but with the advent of kits, this is not enough for indicating incremental changes to a project. There must be some monotonically increasing value present in the application-inventory for a Bottlerocket variant for systems to key off of for versioning and Release comparisons.

This change computes the unix timestamp in ms of the project commit from which a package is built by buildsys and sets the Release of the package as \<timestamp\>+\<commit short sha\>. It also sets the `Version` of a package to the actual package's version, and not the Bottlerocket version.

**Testing done:**

Built a Bottlerocket variant with these changes and inspected packages:

```
cargo make \
    -e=TWOLITER_ALLOW_SOURCE_INSTALL=false \
    -e=TWOLITER_ALLOW_BINARY_INSTALL=false \
    -e=TWOLITER_SKIP_VERSION_CHECK=true \
    build-variant
```

```
rpm -qip build/rpms/shim/bottlerocket-shim-15.8-11717197220_a87879e6.aarch64.rpm
Name        : bottlerocket-shim
Version     : 15.8
Release     : 11717197220_a87879e6
...
```

Launched a bottlerocket instance:

```
bash-5.1# cat aarch64-bottlerocket-linux-gnu/sys-root/usr/share/bottlerocket/application-inventory.json  | grep Epoch 
....

     "Epoch": "1",
     "Epoch": "1",
     "Epoch": "1",

bash-5.1# head aarch64-bottlerocket-linux-gnu/sys-root/usr/share/bottlerocket/application-inventory.json
{
  "Content": [
    {
      "Name": "acpid",
      "Publisher": "Bottlerocket",
      "Version": "2.0.34",
      "Release": "11717197220_a87879e6",
      "Epoch": "1",
      "InstalledTime": "2024-06-04T17:36:24Z",
      "ApplicationType": "Unspecified",
```

**Terms of contribution:**


twoliter/embedded/Makefile.toml
```diff
@@ -24,6 +24,9 @@ BUILDSYS_CARGO_METADATA_PATH = "${BUILDSYS_METADATA_DIR}/cargo_metadata.json"
 BUILDSYS_SBKEYS_PROFILE = { script = ['echo "${BUILDSYS_SBKEYS_PROFILE:-local}"'] }
 BUILDSYS_TIMESTAMP = { script = ["date +%s"] }
 BUILDSYS_VERSION_BUILD = { script = ["git describe --always --dirty --exclude '*' || echo 00000000"] }
+# The unix timestamp in ms of the latest commit of the project.
+# This is an input for setting the Release value of a package.
+BUILDSYS_VERSION_BUILD_TIMESTAMP = { script = ["git show -s --format=%ct HEAD || echo 00000000"] }
```
bcressey: so the number of digits stays consistent:
```suggestion
BUILDSYS_VERSION_BUILD_TIMESTAMP = { script = ["git show -s --format=%ct HEAD || echo 0000000000"] }
```

twoliter/embedded/build.Dockerfile
```diff
@@ -166,9 +169,12 @@ RUN --mount=source=.cargo,target=/home/builder/.cargo \
     --mount=type=cache,target=/home/builder/.cache,from=cache,source=/cache \
     --mount=type=cache,target=/home/builder/rpmbuild/BUILD/sources/models/src/variant,from=variantcache,source=/variantcache \
     --mount=source=sources,target=/home/builder/rpmbuild/BUILD/sources \
+    # '-' is an illegal char in RPM Release field; -dirty commits become _dirty
+    BUILD_ID=$(echo ${BUILD_ID} | tr '-' '_') ; \
     rpmbuild -ba --clean \
       --undefine _auto_set_build_flags \
       --define "_target_cpu ${ARCH}" \
+      --define "dist ${BUILD_ID_TIMESTAMP}_${BUILD_ID}" \
```
bcressey: `-` instead of `_` is more conventional. Also may as well add the `br1` distro tag if we are going to use this field.
```suggestion
      --define "dist ${BUILD_ID_TIMESTAMP}-${BUILD_ID}.br1" \
```

twoliter/embedded/rpm2img
```diff
@@ -159,8 +159,8 @@ rpm -iv --ignorearch --root "${ROOT_MOUNT}" "${PACKAGE_DIR}"/*.rpm
 # Inventory installed packages.
 INVENTORY_QUERY="\{\"Name\":\"%{NAME}\"\
 ,\"Publisher\":\"Bottlerocket\"\
-,\"Version\":\"${VERSION_ID}\"\
-,\"Release\":\"${BUILD_ID}\"\
+,\"Version\":\"%{Version}\"\
```
bcressey, line 167: Let's split out the change to `rpm2img` as a separate commit, since this is where the actual externally visible change happens.
```suggestion
,\"Epoch\":\"1\"\
,\"Version\":\"%{Version}\"\
```

bcressey (commented):
Looks good!

twoliter/embedded/build.Dockerfile
ginglis13: I also preferred `-` but I'm seeing it's an illegal character in the `Release` field:

```
2.699 error: line 10: Illegal char '-' (0x2d) in: Release: 11717197220-a87879e6
```

I'll add `br1` as a tag

twoliter/embedded/rpm2img
webern: > Let's split out the change to `rpm2img` as a separate commit, since this is where the actual externally visible change happens.

What is the "Epoch" in this suggestion? When would it change?

Also, as usual, shell is a mystery to me and I don't understand what `%{Version}` is? Where is that defined?

Currently the only documentation of this json output is `# Inventory installed packages.`. Should we update that to say more about the fields and what they mean?

twoliter/embedded/build.Dockerfile
```diff
@@ -100,8 +100,11 @@ ARG ARCH
 ARG NOCACHE
 ARG VARIANT
 ARG SYSTEMD_NETWORKD
+ARG BUILD_ID
+ARG BUILD_ID_TIMESTAMP
 ENV SYSTEMD_NETWORKD=${SYSTEMD_NETWORKD}
 ENV VARIANT=${VARIANT}
+ENV BUILD_ID=${BUILD_ID} BUILD_ID_TIMESTAMP=${BUILD_ID_TIMESTAMP}
```
webern: Did you mean for these to be on the same line? Does that have some meaning?

```suggestion
ENV BUILD_ID=${BUILD_ID}
ENV BUILD_ID_TIMESTAMP=${BUILD_ID_TIMESTAMP}
```

tools/buildsys/src/args.rs
```diff
@@ -114,6 +115,12 @@ pub(crate) struct BuildPackageArgs {
     #[arg(long, env = "BUILDSYS_VARIANT_FLAVOR")]
     pub(crate) variant_flavor: String,
 
+    #[arg(long, env = "BUILDSYS_VERSION_BUILD")]
+    pub(crate) version_build: String,
+
+    #[arg(long, env = "BUILDSYS_VERSION_BUILD_TIMESTAMP")]
+    pub(crate) version_build_timestamp: String,
```
webern: Other args have set a bad precedent by not having doc comments. Can you add them here and say wheat these are being used for?

webern (approved):

twoliter/embedded/build.Dockerfile
ginglis13: No reason in particular, just following some other conventions I'd seen in this Dockerfile. I can separate these onto two lines 👍

tools/buildsys/src/args.rs
ginglis13: sure thing!

twoliter/embedded/rpm2img
ginglis13: "Epoch" is essentially a trump card to version comparisons. For example, a Bottlerocket 1.20.0 variants includes runc 1.1.12, but under the old process of building application inventory, we specify the `Version` as Bottelerocket's release version - 1.20.0.  Going forward, we will specify package `Version` as the package's actual version, not the project (previously Bottlerocket monorepo)'s version - a variant would then specify, say runc 1.1.13, as `Version` 1.1.13. This is unfortunately lower than our `Version` in previous variants and would break inventory / security advisory comparisons. To circumvent this, packages in kits will start from `Epoch: "1"` (configurable to be overriden by `BUILDSYS_VERSION_BUILD_EPOCH`)


> Also, as usual, shell is a mystery to me and I don't understand what %{Version} is? Where is that defined?

This is querying RPM inventory. `Version` is the corresponding field in a package spec.

I can update the comment here to be more descriptive

ginglis13:
^ split rpm2img changes into a separate commit; add `.br1` tag to `Release`, add more comments to changes.

ginglis13:
Rebased locally and seeing the failure now. There are some integ tests that expect package naming of a certain form that changes in this PR

webern:
> Rebased locally and seeing the failure now. There are some integ tests that expect package naming of a certain form that changes in this PR

Hmm yeah the test ensures that the packages were built for the kit by checking for the RPM output file. Can you update the test to accommodate the new naming convention?

ginglis13:
^ update [twoliter/src/cmd/build.rs](https://github.com/bottlerocket-os/twoliter/pull/263/files#diff-71563dd7908a2799009ca90878db01896433bf0609a4cb7c2ec1b54c997bc780) tests to include defaults for `Release` that populate an rpm name.

ginglis13:
^ use `.` instead of `_` as separator in `Release`

webern (approved):

yeazelm (approved):

ginglis13:
^ set `Epoch` in package spec and each subpackage, source `Epoch` for application-inventory via rpm query now instead of via environment variable

The kernel package definitions are a special case in terms of `Requires` for subpackages. Each has

```
# Pull in expected modules and development files.
Requires: %{name}-modules = %{version}-%{release}
Requires: %{name}-devel = %{version}-%{release}
```

Defined in their spec. This change includes some `sed`ing to insert the subpackage `Epoch` value in these statements.

E.g. showing package and its subpackages set `Epoch`:



```
╰─➤  rpm -qip ~/bottlerocket/build/rpms/kernel-5.15/bottlerocket-kernel-5.15-5.15.158-11717197220.a87879e6.br1.aarch64.rpm
Name        : bottlerocket-kernel-5.15
Epoch       : 1
...
```

```
╰─➤  rpm -qip ~/bottlerocket/build/rpms/kernel-5.15/bottlerocket-kernel-5.15-devel-5.15.158-11717197220.a87879e6.br1.aarch64.rpm
Name        : bottlerocket-kernel-5.15-devel
Epoch       : 1
...
```

tools/buildsys/src/args.rs
```diff
@@ -114,6 +116,23 @@ pub(crate) struct BuildPackageArgs {
     #[arg(long, env = "BUILDSYS_VARIANT_FLAVOR")]
     pub(crate) variant_flavor: String,
 
+    /// version_build is used in building the "Release" of a Package. This value is later
+    /// used in the package's application-inventory. The value defaults to the latest
```
cbgbt: Can you say more in the help text about how `Release` is constructed?

twoliter/embedded/Makefile.toml
```diff
@@ -24,6 +24,11 @@ BUILDSYS_CARGO_METADATA_PATH = "${BUILDSYS_METADATA_DIR}/cargo_metadata.json"
 BUILDSYS_SBKEYS_PROFILE = { script = ['echo "${BUILDSYS_SBKEYS_PROFILE:-local}"'] }
 BUILDSYS_TIMESTAMP = { script = ["date +%s"] }
 BUILDSYS_VERSION_BUILD = { script = ["git describe --always --dirty --exclude '*' || echo 00000000"] }
+# The unix timestamp in ms of the latest commit of the project.
+# This is an input for setting the Release value of a package.
+BUILDSYS_VERSION_BUILD_TIMESTAMP = { script = ["git show -s --format=%ct HEAD || echo 0000000000"] }
+# The epoch field is used for application inventory generation in rpm2img.
```
cbgbt: nit: I'd be more specific here.

```
The epoch field sets the RPM epoch for all built packages. This versioning information is ultimately reflected in the application inventory generated by rpm2img
```

Transparency here could make it easier to debug issues that may occur in the future.

twoliter/embedded/build.Dockerfile
```diff
@@ -88,7 +89,8 @@ RUN \
    && echo -e -n "${UEFI_SECURE_BOOT:+%bcond_without uefi_secure_boot\n}" >> "${RPM_BCONDS}" \
    && echo -e -n "${SYSTEMD_NETWORKD:+%bcond_without systemd_networkd\n}" >> "${RPM_BCONDS}" \
    && echo -e -n "${UNIFIED_CGROUP_HIERARCHY:+%bcond_without unified_cgroup_hierarchy\n}" >> "${RPM_BCONDS}" \
-   && echo -e -n "${XFS_DATA_PARTITION:+%bcond_without xfs_data_partition\n}" >> "${RPM_BCONDS}"
+   && echo -e -n "${XFS_DATA_PARTITION:+%bcond_without xfs_data_partition\n}" >> "${RPM_BCONDS}" \
+   && echo -e -n "Epoch: ${BUILD_EPOCH}\n" >> "${RPM_BCONDS}"
```
cbgbt: This isn't really a bcond. I'd suggest changes here and do all `Epoch` processing in the `rpmbuild` layer.

twoliter/embedded/build.Dockerfile
```diff
@@ -114,6 +122,8 @@ COPY --chown=builder --from=rpm-macros-and-bconds /home/builder/generated.* .
 RUN \
    cat "/usr/lib/rpm/platform/${ARCH}-bottlerocket/macros" generated.rpmmacros > .rpmmacros \
    && cat generated.bconds ${PACKAGE}.spec >> rpmbuild/SPECS/${PACKAGE}.spec \
+   && sed -i "/%package.*$/a Epoch: ${BUILD_EPOCH}" rpmbuild/SPECS/${PACKAGE}.spec \
+   && sed -i "s;\([Requires|Conflicts|Obsoletes]:.*=\) \(\%{version}\-\%{release}$\);\1 ${BUILD_EPOCH}:\2;" rpmbuild/SPECS/${PACKAGE}.spec \
    && find . -maxdepth 1 -not -path '*/\.*' -type f -exec mv {} rpmbuild/SOURCES/ \; \
    && echo ${NOCACHE}
```
cbgbt: I'd update the comment above the `RUN` section to reflect what's going on here. The mogrification of `Requires/etc` lines is pretty unusual without context.

twoliter/src/cmd/build.rs
```diff
@@ -234,7 +234,9 @@ mod test {
         );
 
         for package in packages {
-            let rpm = kit_output_dir.join(&format!("bottlerocket-{package}-0.0-0.{arch}.rpm"));
+            let rpm = kit_output_dir.join(&format!(
+                "bottlerocket-{package}-0.0-00000000000.00000000.br1.{arch}.rpm"
+            ));
```
cbgbt, line 239: Will this test still succeed if the commit is `-dirty`?

twoliter/embedded/Makefile.toml
```diff
@@ -857,6 +862,9 @@ export BUILDSYS_VARIANT_PLATFORM="${BUILDSYS_VARIANT_PLATFORM:?}"
 export BUILDSYS_VARIANT_RUNTIME="${BUILDSYS_VARIANT_RUNTIME:?}"
 export BUILDSYS_VARIANT_FAMILY="${BUILDSYS_VARIANT_FAMILY:?}"
 export BUILDSYS_VARIANT_FLAVOR="${BUILDSYS_VARIANT_FLAVOR}"
+export BUILDSYS_VERSION_BUILD="${BUILDSYS_VERSION_BUILD}"
+export BUILDSYS_VERSION_BUILD_TIMESTAMP="${BUILDSYS_VERSION_BUILD_TIMESTAMP}"
+export BUILDSYS_VERSION_BUILD_EPOCH="${BUILDSYS_VERSION_BUILD_EPOCH}"
```
webern: Aren't these also needed in `build-kit`?

twoliter/src/cmd/build.rs
webern: I'm confused as to how these default values are arriving in the `build-package` calls. I think this test is actually demonstrating the problem of having these args absent from `cargo make build-kit`. `build-kit` will result in package builds, and the package builds will then have these values absent.

I think buildsys needs to choke on the absence of these values and the Makefile needs to ensure they exist.

ginglis13: > Will this test still succeed if the commit is `-dirty`?

If the commit were dirty, the rpm would be named `"bottlerocket-{package}-0.0-00000000000.00000000_dirty.br1.{arch}.rpm"` per https://github.com/bottlerocket-os/twoliter/pull/263/files#diff-a08098dd4b8cb6afda85dec0555534eaeb5a75f4e43f2db7ddcbc2e8c63d0e17R170-R171. So no, this test would not succeed if the package is being built from a dirty tree.

> I'm confused as to how these default values are arriving in the build-package calls. I think this test is actually demonstrating the problem of having these args absent from cargo make build-kit. build-kit will result in package builds, and the package builds will then have these values absent.

These values are set via environment variables in twoliter's Makefile.toml: https://github.com/bottlerocket-os/twoliter/pull/263/files#diff-ad5b91511e741abc0f447ce47f0d0c399ec15a4acd7666aa0f30094356d53af5R27-R31. They are read in `BuildPackageArgs` here: https://github.com/bottlerocket-os/twoliter/pull/263/files#diff-d69207eb08549801c5be3079f2fef42b9c132b83ecdb9d73160b91db074395cdR119-R135, which ultimately is used in a `build-package` call. I see between this and other similar comments, we must set these values in kit build as well, otherwise we'll always have default values present for packages built from kits...

> I think buildsys needs to choke on the absence of these values and the Makefile needs to ensure they exist.

can do

ginglis13: I did some validation with kit building given these changes. Because kit building also builds packages, the required environment variables will be set for package building, and resultant kits will have packages with these values properly set:

```
$ ls tests/projects/local-kit/build/kits/extra-3-kit/aarch64/Packages
bottlerocket-pkg-e-0.0-01717804500.7185d90.br1.aarch64.rpm
bottlerocket-pkg-f-0.0-01717804500.7185d90.br1.aarch64.rpm
bottlerocket-pkg-g-0.0-01717804500.7185d90.br1.aarch64.rpm

$ ls tests/projects/local-kit/build/rpms/pkg-b
bottlerocket-pkg-b-0.0-01717804500.7185d90.br1.aarch64.rpm
```

As for why the integration test is using default values, that's because the integ test builds kits in a temp dir that and is not a git project.

twoliter/embedded/Makefile.toml
ginglis13: See above comment, we won't need these in `build-kit` as they're already included in `build-package`

ginglis13:
^ take feedback to add some clarity in comments, group all Epoch changes to an RPM in the `rpmbuild` layer

webern (approved):

twoliter/embedded/build.Dockerfile
```diff
@@ -218,7 +231,9 @@ USER builder
 
 COPY --chown=builder --from=rpm-macros-and-bconds /home/builder/generated.* .
 
-# Build the metadata RPM for the variant.
+# Build the metadata RPM for the variant. Give it an Epoch.
+ARG BUILD_EPOCH
+ENV BUILD_EPOCH=${BUILD_EPOCH}
```
bcressey: Do we need this? The variables don't seem to be used.

twoliter/embedded/rpm2img
```diff
@@ -157,10 +157,16 @@ INSTALL_TIME="$(date -u +%Y-%m-%dT%H:%M:%SZ)"
 rpm -iv --ignorearch --root "${ROOT_MOUNT}" "${PACKAGE_DIR}"/*.rpm
 
 # Inventory installed packages.
+# Values specified as %{field-name} are retrieved from the package's corresponding 
+# "field-name:" in the RPM. Otherwise, the "Epoch" field, defaulting to 1, is used
+# to trump breaking changes that may occur when the versioning scheme for a package
+# changes between releases - for example, if a newer release of a package were to
+# have a lower "Release".
```
bcressey: Comment is out of date now - we aren't defaulting Epoch here.

tools/buildsys/src/args.rs
```diff
@@ -188,6 +210,9 @@ pub(crate) struct RepackVariantArgs {
     #[arg(long, env = "BUILDSYS_VERSION_BUILD")]
     pub(crate) version_build: String,
 
+    #[arg(long, env = "BUILDSYS_VERSION_BUILD_EPOCH")]
+    pub(crate) version_build_epoch: String,
```
bcressey: We don't build RPMs while repacking, so this isn't needed.

tools/buildsys/src/args.rs
```diff
@@ -169,6 +188,9 @@ pub(crate) struct BuildVariantArgs {
...
```
bcressey: We can drop this - we control the only package that's built during variant builds via `metadata.spec`, and we can just set Epoch values there directly if needed.

tools/buildsys/src/args.rs
```diff
@@ -29,6 +29,8 @@ const REBUILD_VARS: [(&str, u8); 15] = [
     ("BUILDSYS_TIMESTAMP", VARIANT),
     ("BUILDSYS_VARIANT", VARIANT),
     ("BUILDSYS_VERSION_BUILD", VARIANT),
+    ("BUILDSYS_VERSION_BUILD_TIMESTAMP", VARIANT),
+    ("BUILDSYS_VERSION_BUILD_EPOCH", VARIANT),
```
bcressey: We should re-drive all the package builds if Epoch changes, because newly built packages may end up with dependencies on previous package builds that assume the Epoch is set as expected:
```suggestion
    ("BUILDSYS_VERSION_BUILD_EPOCH", PACKAGE),
```

tools/buildsys/src/builder.rs
```diff
@@ -198,6 +204,7 @@ struct VariantBuildArgs {
     variant_platform: String,
     variant_runtime: String,
     version_build: String,
+    version_build_epoch: String,
```
bcressey:
```suggestion
```

tools/buildsys/src/builder.rs
```diff
@@ -211,6 +218,7 @@ impl VariantBuildArgs {
             self.data_image_publish_size_gib.to_string(),
         );
         args.build_arg("BUILD_ID", &self.version_build);
+        args.build_arg("BUILD_EPOCH", &self.version_build_epoch);
```
bcressey:
```suggestion
```

tools/buildsys/src/builder.rs
```diff
@@ -248,6 +256,7 @@ struct RepackVariantBuildArgs {
     partition_plan: String,
     variant: String,
     version_build: String,
+    version_build_epoch: String,
```
bcressey:
```suggestion
```

tools/buildsys/src/builder.rs
```diff
@@ -268,6 +277,7 @@ impl RepackVariantBuildArgs {
         args.build_arg("PARTITION_PLAN", &self.partition_plan);
         args.build_arg("VARIANT", &self.variant);
         args.build_arg("BUILD_ID", &self.version_build);
+        args.build_arg("BUILD_EPOCH", &self.version_build_epoch);
```
bcressey:
```suggestion
```

twoliter/embedded/build.Dockerfile
```diff
@@ -67,7 +67,6 @@ ARG SYSTEMD_NETWORKD
 ARG UNIFIED_CGROUP_HIERARCHY
 ARG XFS_DATA_PARTITION
 ARG FIPS
-
```
bcressey: nit: revert this change

twoliter/embedded/build.Dockerfile
```diff
@@ -110,10 +115,15 @@ COPY ./packages/${PACKAGE}/ .
 
 COPY --chown=builder --from=rpm-macros-and-bconds /home/builder/generated.* .
 
-# Merge generated bconds with the package spec, and put sources in the right place.
+# Merge generated bconds with the package spec, and put sources in the right place. Set Epoch values
+# for the package and all of its subpackages. Additionally, for packages declaring explicit package versions
+# via "{Requires,Conflicts,Obsoletes}: = ...", ensure the epoch for the specified package is set.
 RUN \
    cat "/usr/lib/rpm/platform/${ARCH}-bottlerocket/macros" generated.rpmmacros > .rpmmacros \
+   && echo -e -n "Epoch: ${BUILD_EPOCH}\n" >> rpmbuild/SPECS/${PACKAGE}.spec \
```
bcressey:
```suggestion
   && echo "Epoch: ${BUILD_EPOCH}" >> rpmbuild/SPECS/${PACKAGE}.spec \
```

twoliter/embedded/build.Dockerfile
```diff
@@ -110,10 +115,15 @@ COPY ./packages/${PACKAGE}/ .
...
    && cat generated.bconds ${PACKAGE}.spec >> rpmbuild/SPECS/${PACKAGE}.spec \
+   && sed -i "/%package.*$/a Epoch: ${BUILD_EPOCH}" rpmbuild/SPECS/${PACKAGE}.spec \
```
bcressey: consider deleting existing `Epoch` lines, and being more rigorous about matches:
```suggestion
   && sed -i '/^Epoch:/d' rpmbuild/SPECS/${PACKAGE}.spec \
   && sed -i "/^%package\s.*$/a Epoch: ${BUILD_EPOCH}" rpmbuild/SPECS/${PACKAGE}.spec \
```

twoliter/embedded/build.Dockerfile
```diff
@@ -110,10 +115,15 @@ COPY ./packages/${PACKAGE}/ .
...
+   && sed -i "s;\([Requires|Conflicts|Obsoletes]:.*=\) \(\%{version}\-\%{release}$\);\1 ${BUILD_EPOCH}:\2;" rpmbuild/SPECS/${PACKAGE}.spec \
```
bcressey: Why do we need this? It feels like it will be brittle in the face of our more complex boolean dependencies.

twoliter/embedded/build.Dockerfile
```diff
@@ -157,9 +167,12 @@ RUN --mount=source=.cargo,target=/home/builder/.cargo \
     --mount=type=cache,target=/home/builder/.cache,from=cache,source=/cache \
     --mount=type=cache,target=/home/builder/rpmbuild/BUILD/sources/models/src/variant,from=variantcache,source=/variantcache \
     --mount=source=sources,target=/home/builder/rpmbuild/BUILD/sources \
+    # '-' is an illegal char in RPM Release field; -dirty commits become _dirty
+    BUILD_ID=$(echo ${BUILD_ID} | tr '-' '_') ; \
     rpmbuild -ba --clean \
       --undefine _auto_set_build_flags \
       --define "_target_cpu ${ARCH}" \
+      --define "dist ${BUILD_ID_TIMESTAMP}.${BUILD_ID}.br1" \
```
bcressey: I'm usually the one that complains about "sanitizing" `-dirty` but here I would do that instead:
```suggestion
      --define "dist ${BUILD_ID_TIMESTAMP}.${BUILD_ID//-dirty/}.br1" \
```

We don't rebuild RPMs when the build ID changes, so for local development this would make most packages always have the "-dirty" release tag. That's going to irk me because it will show up in the RPM file name even if that package doesn't have any uncommitted changes in my checkout.

twoliter/embedded/build.Dockerfile
ginglis13: thanks, yes these are no longer needed

twoliter/embedded/build.Dockerfile
ginglis13: This is needed because when we set an Epoch for a package, *and* all subpacakges within a package, we must now reference those subpackages with their Epoch, else rpmbuild will complain:

```
nothing provides bottlerocket-kernel-5.15-devel = 5.15.158-11717197220.a87879e6.br1 needed
 by bottlerocket-kernel-5.15-1:5.15.158-11717197220.a87879e6.br1.aarch64 from repo
```

We only have these situations in kernel package in b-os/b:

```
$   git grep '%{version}' | grep 'Requires'
packages/kernel-5.10/kernel-5.10.spec:Requires: %{name}-modules = %{version}-%{release}
packages/kernel-5.10/kernel-5.10.spec:Requires: %{name}-devel = %{version}-%{release}
packages/kernel-5.15/kernel-5.15.spec:Requires: %{name}-modules = %{version}-%{release}
packages/kernel-5.15/kernel-5.15.spec:Requires: %{name}-devel = %{version}-%{release}
packages/kernel-6.1/kernel-6.1.spec:Requires: %{name}-modules = %{version}-%{release}
packages/kernel-6.1/kernel-6.1.spec:Requires: %{name}-devel = %{version}-%{release}
```

and they all use simple `=` comparison. I agree it is definitely more robust to cover all the operators possible here, I can give that a shot

twoliter/embedded/build.Dockerfile
ginglis13: > That's going to irk me because it will show up in the RPM file name even if that package doesn't have any uncommitted changes in my checkout.

Yea that's a fair point. I'll take your suggestion :)

ginglis13:
^ first force push, address feedback:

* remove unnecessary inclusion of the epoch environment variable in buildsys subcommands where it is not needed.
* clean up some `sed` commands to be more thorough, especially, in checking comparators for `Requires|Obsoletes|Conflicts`
* strip `-dirty` completely from `BUILD_ID` when building an RPM's `Release` field

^ second force push, rebase

^ 3rd, fix a missed nit in above two pushes

twoliter/embedded/build.Dockerfile
cbgbt: I think this fits into the case that we could alter the expectations of a packager by modifying their Epoch value. Would it be better to fail the build than to produce something unexpected?

cbgbt (approved):

webern (approved):

twoliter/embedded/build.Dockerfile
ginglis13: In the latest revision of this PR, I opted not to delete existing `Epoch` lines for two reasons:

1. attempting to delete previous Epoch was resulting in invalid RPM, with an error along the lines of `File must begin with "/": Epoch: `. I didn't dive further becaue I concluded that instead of making this change, we can:
2. By always attempting to set our `Epoch`, we will induce RPM failures for the case where a packager attempts to specify their own `Epoch`. This seems more reasonable than silently changing the value they intend to set as `Epoch`

yeazelm (approved):

PR #265, opened 2024-06-03
twoliter: skip missing per-package dirs

### LINKED ISSUE #264: "no such file or directory" build failure after twoliter upgrade
When moving to the newer version of `twoliter`, the build fails with this message if crossing a commit that does not include glibc changes that would trigger a rebuild of that package:
```
find: '/host/build/rpms/glibc/': No such file or directory
```

bcressey:
**Issue number:**

Closes #264

**Description of changes:**
This fixes backwards compatibility when moving between releases when the twoliter version changes but some of the packages (such as glibc) do not. The cargo metadata for the older builds is still valid, and the RPMs will be present in the top-level output directory, but the per-package subdirectories will not yet exist.

**Testing done:**
Tested a local branch where I'd made changes to `os.spec` and added some packages. First reverted the `twoliter` upgrade as my base commit, then built all my changes. Then dropped the `twoliter` upgrade in a rebase, and verified that everything built again. [Here](https://gist.github.com/bcressey/022ad0a9948334ffe71bcd2b32b14853) is the output of `tree build/rpms` after the second build.

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/build.Dockerfile
+++ b/twoliter/embedded/build.Dockerfile
@@ -122,6 +122,7 @@ RUN --mount=target=/host \
     find /host/build/rpms/ -mindepth 1 -maxdepth 1 -name '*.rpm' -size +0c -print -exec \
       ln -snft ./rpmbuild/RPMS {} \+ && \
     for pkg in ${PACKAGE_DEPENDENCIES} ; do \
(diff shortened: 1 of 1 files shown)


yeazelm (approved):
Nice! 🚢

arnaldo2792 (approved):

webern (approved):

PR #270, opened 2024-06-04
twoliter: fetch command for downloading and extracting external kits

### LINKED ISSUE #258: twoliter: download and extract kits to build folder
We need twoliter to fetch all the required external kits and extract them into the build directory like follows:

```
build/external-kits/<vendor>/<name>/<arch>/...
```

jmt-lab:
**Issue number:** 258

Closes #258 

**Description of changes:**

* Adds functionality to Lock to download and extract kits resolved by Twoliter.lock to build/external-kits
* Writes a metadata file that can be used by rpm2kit for generating the kit metadata for publishing a local kit
* Adds `twoliter fetch` command
* Remove sdk from remaining in lock resolution as an sdk image is not a kit image and does not have metadata

**Testing done:**

* Uploaded a test kit image to ECR and updated Twoliter.toml to depend on it
* Ran twoliter update to resolve Twoliter.lock
* Ran twoliter fetch to fetch kits
* Verified metadata and extraction path

**Terms of contribution:**


twoliter/src/cmd/fetch.rs
```diff
@@ -0,0 +1,23 @@
...
+use anyhow::Result;
+use clap::Parser;
+use std::path::PathBuf;
+
+#[derive(Debug, Parser)]
+pub(crate) struct Fetch {
+    /// Path to Twoliter.toml. Will search for Twoliter.toml when absent
+    #[clap(long = "project-path")]
+    project_path: Option<PathBuf>,
+    #[clap(long = "arch", default_value = "x86_64")]
```
webern:
```suggestion
    project_path: Option<PathBuf>,
    
    #[clap(long = "arch", default_value = "x86_64")]
```

twoliter/src/cmd/fetch.rs
```diff
@@ -0,0 +1,23 @@
+use crate::lock::Lock;
+use crate::project;
+use anyhow::Result;
+use clap::Parser;
+use std::path::PathBuf;
+
+#[derive(Debug, Parser)]
+pub(crate) struct Fetch {
```
webern, line 8: I guess I always ask this. Would we anticipate further subcommands? `twoliter fetch sdk` and `twoliter fetch sources` for example? If so, should this be `twoliter fetch kits`?

twoliter/src/main.rs
```diff
@@ -1,3 +1,5 @@
+extern crate core;
+
```
webern:
```suggestion
```

twoliter/src/lock.rs
```diff
@@ -139,6 +170,111 @@ impl Lock {
...
+            kits.push(self.extract_kit(&target_dir, image, arch).await?);
+        }
+        let kit_list = serde_json::to_vec(&ExternalKitMetadata {
+            sdk: Image {
+                name: self.sdk.name.clone(),
+                version: self.sdk.version.clone(),
+                vendor: self.sdk.vendor.clone(),
+            },
+            kits,
+        })
```
webern:
```suggestion
        let kit_list = serde_json::to_vec(&self.external_kit_metadata(kits));
```

twoliter/src/lock.rs
```diff
@@ -147,7 +283,7 @@ impl Lock {
         let mut remaining: Vec<Image> = project.kits();
         let mut sdk_set: HashSet<Image> = HashSet::new();
         if let Some(sdk) = project.sdk_image() {
-            remaining.push(sdk.clone());
+            // We don't scan over the sdk images as they are not kit images and there is no
```
webern: Typo: `there is no`

```
            // We don't scan over the sdk images as they are not kit images and there is no
```

twoliter/src/lock.rs
```diff
@@ -139,6 +170,111 @@ impl Lock {
...
+        let manifest_list: ManifestListView = serde_json::from_slice(manifest_bytes.as_slice())
+            .context("failed to deserialize manifest list")?;
+        let manifest = manifest_list
+            .manifests
+            .iter()
+            .find(|x| x.platform.as_ref().unwrap().architecture == docker_arch)
+            .context(format!(
+                "could not find kit image for architecture '{}' at {}",
+                arch, image.source
+            ))?;
```
webern: It seems like a few smaller functions could be factored out to improve readability:

```suggestion
        let manifest = self.get_manifest(image.source.as_str(), image, docker_arch)?;
```

twoliter/src/lock.rs
```diff
@@ -139,6 +170,111 @@ impl Lock {
...
+        let image_source = image.source.replace(
+            format!(":v{}", image.version).as_str(),
+            format!("@{}", manifest.digest).as_str(),
+        );
```
webern: Seems like something the object could do.

```suggestion
        let image_source = image.image_source();
```

twoliter/src/lock.rs
```diff
@@ -139,6 +170,111 @@ impl Lock {
...
+        let mut oci_archive = Archive::new(oci_archive.as_slice());
+        // Unpack the oci image to tempdirectory
+        oci_archive
+            .unpack(tmp_dir.path())
+            .context("failed to unpack oci image")?;
+        // Now read in the index
+        let index_bytes = read(tmp_dir.path().join("index.json")).await?;
+        let index: IndexView = serde_json::from_slice(index_bytes.as_slice())
+            .context("failed to deserialize oci image index")?;
```
webern: This looks like a function.

```suggestion
        let index = self.get_index(x, y, z)?;
```

twoliter/src/lock.rs
```diff
@@ -139,6 +170,111 @@ impl Lock {
...
+    {
+        let tmp_dir =
+            tempdir().context("failed to create a temporary directory for kit extraction")?;
+        let vendor = image.vendor.clone();
+        let name = image.name.clone();
+        let docker_arch = match arch {
+            "x86_64" => Ok("amd64"),
+            "aarch64" => Ok("arm64"),
+            _ => Err(anyhow!("unsupported architecture '{}'", arch)),
+        }?;
```
cbgbt: nit: probably a good idea to extract this.

twoliter/src/lock.rs
cbgbt: Can we include image digests as part of this metadata?

We need this file to alert Cargo that rebuilds need to be performed. It's possible that a local developer could iterate on an existing version, keeping all existing metadata the same while altering only the digest of one of the images.

twoliter/src/lock.rs
```diff
@@ -139,6 +170,111 @@ impl Lock {
         Ok(lock)
     }
 
+    /// Fetches all external kits defined in a Twoliter.lock to the build directory
+    pub(crate) async fn fetch(&self, project: &Project, arch: &str) -> Result<()> {
+        let target_dir = project.project_dir().join("build/external-kits");
```
cbgbt: Should this be a `CONSTANT` or else have some accessor like `project.external_kits_dir()`?

twoliter/src/lock.rs
```diff
@@ -139,6 +170,111 @@ impl Lock {
...
+        create_dir_all(&target_dir)
+            .await
+            .context("failed to create external-kits directory")?;
+        let mut kits: Vec<Image> = Vec::new();
+        for image in self.kit.iter() {
+            kits.push(self.extract_kit(&target_dir, image, arch).await?);
+        }
+        let kit_list = serde_json::to_vec(&ExternalKitMetadata {
+            sdk: Image {
```
sam-berning: since we do basically the same conversion in `extract_kit`, might just make sense to impl `Into<Image>` for `LockedImage`

twoliter/src/lock.rs
```diff
@@ -243,20 +379,25 @@ impl Lock {
...
                     image.to_string(),
                     manifest.digest
                 )
             );
+            let label_str = String::from_utf8_lossy(label_bytes.as_slice()).to_string();
+            let label_str = label_str
+                .trim()
+                .strip_prefix('"')
+                .unwrap_or(label_str.as_str());
+            let label_str = label_str.strip_suffix('"').unwrap_or(label_str);
```
```suggestion
            let label_str = label_str.trim().trim_matches('"');
```

jmt-lab:
Updated by comments
Added digest to external kit metadata file

jmt-lab:
Fix clippy issues

twoliter/src/cmd/fetch.rs
webern: Let's leave this one unresolved since it's an outstanding question with no corresponding change.

webern (approved):

twoliter/src/lock.rs
```diff
@@ -139,6 +195,130 @@ impl Lock {
...
+    fn external_kit_metadata<E>(&self, kits: E) -> ExternalKitMetadata
+    where
+        E: AsRef<[ExternalImage]>,
+    {
+        ExternalKitMetadata {
+            sdk: Image {
+                name: self.sdk.name.clone(),
+                version: self.sdk.version.clone(),
+                vendor: self.sdk.vendor.clone(),
+            },
```
cbgbt: Should the SDK not also include the image digest? I think we need to ensure we also trigger rebuilds if someone overrides the SDK in some way.

jmt-lab:
Added digest to LockedImage
Calculating digest for image list off the image index
Ensured external-kit-metadata.json is printed out in canonical json
Added external kit metadata to rerun-if-changed

tools/buildsys/src/main.rs
```diff
@@ -121,9 +121,11 @@ fn run(args: Buildsys) -> Result<()> {
 }
 
 fn build_package(args: BuildPackageArgs) -> Result<()> {
+    let external_kit_metadata = "build/external-kits/external-kit-metadata.json";
```
webern: Could be a const that is used both here and for `project.external_kit_metadata()`.

twoliter/src/cmd/fetch.rs
jmt-lab: Yeah hit resolve by accident there, I'm not sure either way, I kind of like it being just fetch cause it operates off Twoliter.lock which is generated just by update

webern (approved):

twoliter/src/lock.rs
```diff
@@ -171,8 +330,8 @@ impl Lock {
                     (image.name.clone(), image.vendor.clone()),
                     image.version.clone(),
                 );
-                let locked_image = LockedImage::new(vendor, image);
-                let kit = Self::find_kit(vendor, &locked_image).await?;
+                let mut locked_image = LockedImage::new(vendor, image, "");
+                let kit = Self::find_kit(vendor, &mut locked_image).await?;
```
cbgbt: nit: mutating the object after creation is a strange contract that could lead to bugs in the future -- the state of a `LockedImage` can't be known without context. I'm guessing we do this so that we can take advantage of `LockedImage` providing us the source URI?

Can we do something here to make the OO contract stronger?

twoliter/src/lock.rs
```diff
@@ -92,9 +108,37 @@ struct ManifestListView {
...
+#[derive(Deserialize, Debug, Clone)]
 struct ManifestView {
     digest: String,
+    platform: Option<Platform>,
+}
+
+#[derive(Deserialize, Debug, Clone)]
+struct Platform {
+    architecture: String,
+}
```
bcressey: Can we use a stronger type here? We know only `amd64` or `arm64` are expected.

twoliter/src/lock.rs
```diff
@@ -92,9 +108,37 @@ struct ManifestListView {
...
+
+#[derive(Deserialize, Debug)]
+struct ManifestLayoutView {
+    layers: Vec<Layer>,
+}
+
+#[derive(Deserialize, Debug)]
+struct Layer {
+    digest: String,
+}
```
bcressey: This could be a type with stronger validation rules applied during deserialization.

twoliter/src/lock.rs
```diff
@@ -139,6 +183,121 @@ impl Lock {
         Ok(lock)
     }
 
+    fn external_kit_metadata(&self) -> ExternalKitMetadata {
+        ExternalKitMetadata {
+            sdk: self.sdk.clone(),
+            kits: self.kit.clone(),
+        }
+    }
```
bcressey, line 339: Is this different from just calling `self.clone()`? Can we just derive `Clone` instead?

twoliter/src/lock.rs
```diff
@@ -1,14 +1,20 @@
-use crate::common::fs::{remove_file, write};
+use crate::common::fs::{create_dir_all, read, remove_file, write};
 use crate::project::{Image, Project, Vendor};
 use crate::schema_version::SchemaVersion;
-use anyhow::{ensure, Context, Result};
+use anyhow::{anyhow, ensure, Context, Result};
 use base64::Engine;
+use olpc_cjson::CanonicalFormatter;
```
bcressey: maybe this:
```suggestion
use olpc_cjson::CanonicalFormatter as CanonicalJsonFormatter;
```

twoliter/src/lock.rs
```diff
@@ -139,6 +195,121 @@ impl Lock {
...
+        );
+        let mut oci_archive = Archive::new(oci_archive.as_slice());
+        // Unpack the oci image to tempdirectory
+        oci_archive
+            .unpack(path.as_ref())
+            .context("failed to unpack oci image")?;
+        // Now read in the index
+        let index_bytes = read(path.as_ref().join("index.json")).await?;
+        serde_json::from_slice(index_bytes.as_slice())
+            .context("failed to deserialize oci image index")
```
bcressey: nit: these comments seem pretty obvious based on variable names and errors
```suggestion
        oci_archive
            .unpack(path.as_ref())
            .context("failed to unpack oci image")?;
        let index_bytes = read(path.as_ref().join("index.json")).await?;
        serde_json::from_slice(index_bytes.as_slice())
            .context("failed to deserialize oci image index")
```

twoliter/src/lock.rs
```diff
@@ -1,19 +1,37 @@
...
 use semver::Version;
 use serde::{Deserialize, Serialize};
+use sha2::Digest;
 use std::collections::{HashMap, HashSet};
 use std::fmt::Display;
+use std::fs::File;
 use std::hash::{Hash, Hasher};
 use std::mem::take;
+use std::path::Path;
+use tar::Archive;
```
bcressey:
```suggestion
use tar::Archive as TarArchive;
```

twoliter/src/lock.rs
```diff
@@ -139,6 +195,121 @@ impl Lock {
...
+    {
+        let tmp_dir =
+            tempdir().context("failed to create a temporary directory for kit extraction")?;
+        let vendor = image.vendor.clone();
+        let name = image.name.clone();
+        let docker_arch = match arch {
+            "x86_64" => Ok("amd64"),
+            "aarch64" => Ok("arm64"),
+            _ => Err(anyhow!("unsupported architecture '{}'", arch)),
+        }?;
```
bcressey: This should be handled as an implementation detail in `get_manifest` if we don't have another need for `docker_arch` in this function.

jmt-lab:
Updated const for metadata

twoliter/src/lock.rs
```diff
@@ -139,6 +195,121 @@ impl Lock {
...
+        let target_path = path.as_ref().join(format!("{vendor}/{name}/{arch}"));
+        create_dir_all(&target_path).await?;
+
+        // First we want to get the correct digest for our architecture
```
bcressey: It would be good to add more comments for this algorithm. It's not really clear where "getting the correct digest" ends, it seems like we're doing other stuff with side effects along the way toward `let digest = ...` (assuming that's where it happens).

tools/buildsys/src/main.rs
```diff
@@ -124,6 +126,7 @@ fn build_package(args: BuildPackageArgs) -> Result<()> {
     let manifest_file = "Cargo.toml";
     let manifest_path = args.common.cargo_manifest_dir.join(manifest_file);
     println!("cargo:rerun-if-changed={}", manifest_file);
+    println!("cargo:rerun-if-changed={}", EXTERNAL_KIT_METADATA);
```
bcressey, line 128: `cargo` will always rerun a build if told to `rerun-if-changed` on a file that doesn't exist. So you must either emit this conditionally (if the file exists), or ensure that `twoliter` always creates at least an empty file.

twoliter/src/lock.rs
```diff
@@ -139,6 +195,121 @@ impl Lock {
...
+            self.extract_kit(&target_dir, image, arch).await?;
+        }
+        let mut kit_list = Vec::new();
+        let mut ser =
+            serde_json::Serializer::with_formatter(&mut kit_list, CanonicalFormatter::new());
+        self.external_kit_metadata()
+            .serialize(&mut ser)
+            .context("failed to serialize external kit metadata")?;
+        write(
+            target_dir.join("external-kit-metadata.json"),
```
bcressey: Maybe the file and the full path can both be constants, to ensure they line up.

twoliter/src/lock.rs
```diff
@@ -139,6 +195,121 @@ impl Lock {
...
+            .context("failed to deserialize oci manifest")?;
+        for layer in manifest_layout.layers {
+            let digest = layer.digest.replace(':', "/");
+            let mut layer_blob = File::open(tmp_dir.path().join(format!("blobs/{digest}")))
+                .context("failed to read layer of oci image")?;
+            let decompressed = flate2::read::GzDecoder::new(&mut layer_blob);
+            let mut layer_archive = Archive::new(decompressed);
+            layer_archive
+                .unpack(&target_path)
+                .context("failed to unpack layer to disk")?;
```
bcressey: It's not clear where / when we might be retrieving large files or loading them into memory.

Where does the primary content fetch happen? Not the metadata but the potentially GiBs of RPMs. It must be `get_manifest` or `get_index` but those functions sound like they should be fast, and do not imply "do a giant download as a side effect".

twoliter/src/lock.rs
```diff
@@ -243,20 +390,21 @@ impl Lock {
...
+            let label_str = label_str.trim().trim_matches('"');
             // Otherwise we should have a list of json blobs we can fetch the metadata from the label
-            let labels: HashMap<String, String> = serde_json::from_slice(label_bytes.as_slice())
-                .context(format!(
-                    "could not deserialize labels on the image for {}",
-                    image
-                ))?;
+            let labels: HashMap<String, String> = serde_json::from_str(label_str).context(
+                format!("could not deserialize labels on the image for {}", image),
+            )?;
```
bcressey, line 505: What does the "otherwise" comment above refer to? Is there some other source of kit metadata?

twoliter/src/lock.rs
jmt-lab: ExternalKitMetadata is a subset of the fields in the lock file, otherwise irrelevant information to cargo rerun would end up in the json

twoliter/src/lock.rs
jmt-lab: The primary content fetch occurs in get_index as that is when we need the oci image to use data with docker save. I'll rename the function

tools/buildsys/src/main.rs
jmt-lab: This occurs when you run twoliter fetch, there should always be that file with at minimum the sdk info. We could also do this somewhere else too if you have a good idea where?

jmt-lab:
Updated per ben's comments

twoliter/src/lock.rs
jmt-lab: Wasn't very helpful comment, removed it

jmt-lab:
Forgot to add license field to new crate

twoliter/src/lock.rs
bcressey: > ExternalKitMetadata is a subset of the fields in the lock file, otherwise irrelevant information to cargo rerun would end up in the json

Oh, I see - it's because we're in `impl Lock` here. It'd be better not to have so many intervening data structures with their own impls between where `struct Lock` is defined and this point.

twoliter/src/lock.rs
```diff
@@ -139,6 +249,124 @@ impl Lock {
...
+            kits: self.kit.clone(),
+        }
+    }
+
+    /// Fetches all external kits defined in a Twoliter.lock to the build directory
+    pub(crate) async fn fetch(&self, project: &Project, arch: &str) -> Result<()> {
+        let target_dir = project.external_kits_dir();
+        create_dir_all(&target_dir)
+            .await
+            .context("failed to create external-kits directory")?;
```
bcressey: Let's use the real directory path here so it's clear what directory we tried to create.

twoliter/src/lock.rs
```diff
@@ -139,6 +249,124 @@ impl Lock {
...
+        }
+        let mut kit_list = Vec::new();
+        let mut ser =
+            serde_json::Serializer::with_formatter(&mut kit_list, CanonicalJsonFormatter::new());
+        self.external_kit_metadata()
+            .serialize(&mut ser)
+            .context("failed to serialize external kit metadata")?;
+        write(project.external_kits_metadata(), kit_list.as_slice())
+            .await
+            .context("failed to write external kit metadata")?;
```
bcressey: Add the file path to this error message.

twoliter/src/lock.rs
```diff
@@ -139,6 +249,124 @@ impl Lock {
...
+    }
+
+    async fn pull_image<P>(&self, path: P, digest_uri: &str) -> Result<IndexView>
+    where
+        P: AsRef<Path>,
+    {
+        let oci_archive = docker!(
+            ["save", digest_uri],
+            format!("failed to fetch kit and save to disk from {}", digest_uri)
+        );
```
bcressey: Is this saving the entire image into memory via console output?

If it's implicitly saving a file, let's make that explicit. If it's not, please make it save a file. We don't want to read a remote file into a stack variable ...

twoliter/src/lock.rs
```diff
@@ -139,6 +249,124 @@ impl Lock {
...
+        let mut oci_archive = Archive::new(oci_archive.as_slice());
+        oci_archive
+            .unpack(path.as_ref())
+            .context("failed to unpack oci image")?;
+        let index_bytes = read(path.as_ref().join("index.json")).await?;
+        serde_json::from_slice(index_bytes.as_slice())
+            .context("failed to deserialize oci image index")
+    }
```
bcressey: How do we ensure that these files are garbage collected? (Proposed answer: we don't and it's a `cargo make clean` or `twoliter clean` concern for now.)

However it seems like you need to do at least a little bit cache management, or performance will be dire: if we already have the archive pulled and unpacked, we should not do that whole operation again.

Maybe maintain a blob or layer store under `/build` and add a `clean` task to prune it.

twoliter/src/lock.rs
```diff
@@ -139,6 +249,124 @@ impl Lock {
...
+            tempdir().context("failed to create a temporary directory for kit extraction")?;
+        let vendor = image.vendor.clone();
+        let name = image.name.clone();
+        let docker_arch = match arch {
+            "x86_64" => Ok("amd64"),
+            "aarch64" => Ok("arm64"),
+            _ => Err(anyhow!("unsupported architecture '{}'", arch)),
+        }?;
+        let target_path = path.as_ref().join(format!("{vendor}/{name}/{arch}"));
+        create_dir_all(&target_path).await?;
```
bcressey: Somewhere we either need to enforce that vendor / name / arch only contain path-safe characters - no dots, slashes, whitespace, etc - or else we need to URL encode them here.

twoliter/src/lock.rs
```diff
@@ -139,6 +249,124 @@ impl Lock {
...
+            .replace(':', "/");
+        let manifest_bytes = read(tmp_dir.path().join(format!("blobs/{digest}")))
+            .await
+            .context("failed to read manifest blob")?;
+        let manifest_layout: ManifestLayoutView = serde_json::from_slice(manifest_bytes.as_slice())
+            .context("failed to deserialize oci manifest")?;
+
+        // Extract each layer into the target directory
+        for layer in manifest_layout.layers {
+            let digest = layer.digest.to_string().replace(':', "/");
```
bcressey: If we're just splitting the string on `algo:digest` - let's do a split instead.

twoliter/src/lock.rs
```diff
@@ -139,6 +249,124 @@ impl Lock {
...
+        serde_json::from_slice(index_bytes.as_slice())
+            .context("failed to deserialize oci image index")
+    }
+
+    async fn extract_kit<P>(&self, path: P, image: &LockedImage, arch: &str) -> Result<()>
+    where
+        P: AsRef<Path>,
+    {
+        let tmp_dir =
+            tempdir().context("failed to create a temporary directory for kit extraction")?;
```
bcressey: Please ensure that the temp directories are created under `build`. `pubsys` dealt with this recently for the fetch image work.

twoliter/src/lock.rs
jmt-lab: I did a replace cause this cleanly creates the path for use, i can do a split join if you would prefer?

twoliter/src/lock.rs
jmt-lab: Reordered some things to make this cleaner

jmt-lab:
Updates to comments including

* Add ValidIdentifier to validate the name and vendor fields
* Implemented caching of the oci archive and external kits
* Added buildsys changes here so clean-kits will clean up external-kit caches as well

jmt-lab:
Fixed tests

twoliter/src/lock.rs
```diff
@@ -139,15 +249,181 @@ impl Lock {
...
+    where
+        P: AsRef<Path>,
+    {
+        let digest_uri = image.digest_uri(digest);
+        let path = path.as_ref();
+        let digest_file = path.join("digest");
+        let cached_path = path.join("cached");
+        let oci_archive_path = cached_path.join(digest.replace(':', "-"));
+        if !oci_archive_path.exists() {
+            let oci_archive_str = oci_archive_path.to_str().context(format!(
```
bcressey: nit: `to_string_lossy` should be fine here

twoliter/src/lock.rs
```diff
@@ -139,15 +249,181 @@ impl Lock {
...
+        // the oci index
+        let digest_file = target_path.join("digest");
+        if digest_file.exists()
+            && read_to_string(&digest_file).await.context(format!(
+                "failed to read cached digest id from {}",
+                digest_file.display()
+            ))? == manifest.digest
+        {
+            return Ok(());
+        }
```
bcressey: You end up with this special "digest" marker file in two places that are loosely coupled. It might be better to refactor it in a more object-oriented spirit:
```
let oci_archive = OciArchive::new(&digest);
# checks for the saved image locally, or else pulls and saves it
oci_archive.pull_image()?;
# checks for the matching digest in path
# otherwise cleans up the path and unpacks the archive
oci_archive.unpack_layers(&target_path)?
```

twoliter/src/lock.rs
```diff
@@ -139,15 +249,181 @@ impl Lock {
...
+            .digest
+            .replace(':', "/");
+        let manifest_bytes = read(tmp_dir.path().join(format!("blobs/{digest}")))
+            .await
+            .context("failed to read manifest blob")?;
+        let manifest_layout: ManifestLayoutView = serde_json::from_slice(manifest_bytes.as_slice())
+            .context("failed to deserialize oci manifest")?;
+
+        // Extract each layer into the target directory
+        for layer in manifest_layout.layers {
```
bcressey: Doesn't seem like you are cleaning up an existing directory, which could lead to weirdly merged content from two different revisions of the kit.

twoliter/src/lock.rs
bcressey: No, this is fine if there's a reason.

jmt-lab:
Updated per latest comments

twoliter/src/lock.rs
```diff
@@ -92,21 +119,186 @@ struct ManifestListView {
...
+            "failed to read digest file at {}",
+            digest_file.display()
+        ))?;
+        if digest == self.digest {
+            return Ok(());
+        }
+
+        remove_dir_all(path).await?;
+        create_dir_all(path).await?;
+        write(&digest_file, self.digest.as_str())
```
bcressey: Now this should be the last operation, since if this write succeeds but a later step fails, then we'll never retry because the digest will match.

twoliter/src/lock.rs
```diff
@@ -139,15 +331,98 @@ impl Lock {
...
+            "aarch64" => Ok("arm64"),
+            _ => Err(anyhow!("unsupported architecture '{}'", arch)),
+        }?;
+        let target_path = path.as_ref().join(format!("{vendor}/{name}/{arch}"));
+        let cache_path = path.as_ref().join("cache");
+        create_dir_all(&target_path).await?;
+        create_dir_all(&cache_path).await?;
+
+        // First get the manifest for the specific requested architecture
+        let manifest = self.get_manifest(image, docker_arch).await?;
```
bcressey: can get rid of `let docker_arch = ...` above and change this:
```suggestion
        // First get the manifest for the specific requested architecture
        let manifest = self.get_manifest(image, arch).await?;
```

webern (approved):

bcressey (approved):

webern (approved):

PR #290, opened 2024-06-13
twoliter: fix use of sdk

jmt-lab:
**Description of changes:**
Changes twoliter to use the sdk source resolved by the lock

**Terms of contribution:**

### DIFF
--- a/twoliter/src/project.rs
+++ b/twoliter/src/project.rs
@@ -136,22 +136,6 @@ impl Project {
         self.sdk.clone()
     }
 
-    pub(crate) fn sdk(&self) -> Result<Option<ImageUri>> {
-        if let Some(sdk) = self.sdk.as_ref() {
-            let vendor = self.vendor.get(&sdk.vendor).context(format!(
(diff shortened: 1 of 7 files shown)


twoliter/src/project.rs
```diff
@@ -136,22 +136,6 @@ impl Project {
         self.sdk.clone()
     }
 
-    pub(crate) fn sdk(&self) -> Result<Option<ImageUri>> {
```
webern, line 139: Is this never being read from the Project by anything? What happens if I specify the SDK in the Project? Nothing? Is it ignored?

Does the resolve function during `twoliter update` throw an error if the SDK listed in the project does not match one of the external kits?

What if there are no external kits at all (e.g. in the bottlerocket-core-kit build)? In that case the SDK must come from the project.

jmt-lab: Yes, the lock resolution verifies both if a sdk exists in Project or auto resolves if not and if there is a mismatch anywhere it will error out

bcressey (approved):

webern (approved):

PR #291, opened 2024-06-14
pubsys: remove build id from tag when publishing kit

sam-berning:
**Issue number:**

Closes #

**Description of changes:**

Removes the build ID from the tag of the kit manifest list.

Also, has `twoliter` automatically discover the build id and version from the project context so that it doesn't need to be manually specified on the command line.

**Testing done:**

Pushed a kit to an ECR repository

**Terms of contribution:**


bcressey (approved):

jmt-lab (changes requested):
Looks like you are missing changing the manifest create and push to only use version only and not the build id

jmt-lab (approved):
LGTM

bcressey (approved):

PR #294, opened 2024-06-14
Ootb/docker build fix

jmt-lab:
**Description of changes:**

* Fixes find command in build.Dockerfile
* Turns off compression in docker build in buildsys

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/build.Dockerfile
+++ b/twoliter/embedded/build.Dockerfile
@@ -376,8 +376,8 @@ WORKDIR /tmp
 RUN --mount=target=/host \
     mkdir -p /local/archives \
     && KERNEL="$(printf "%s\n" ${PACKAGES} | awk '/^kernel-/{print $1}')" \
-    && find /host/build/rpms/ -maxdepth 2 -type f \
-        -name "bottlerocket-${KERNEL}-archive-*.rpm" \
+    && find /host/build/ -type f \
+        -name "bottlerocket-${KERNEL}-archive-*.${ARCH}.rpm" \
         -exec cp '{}' '/local/archives/' ';' \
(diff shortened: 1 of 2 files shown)


twoliter/embedded/build.Dockerfile
```diff
@@ -376,7 +376,7 @@ WORKDIR /tmp
 RUN --mount=target=/host \
     mkdir -p /local/archives \
     && KERNEL="$(printf "%s\n" ${PACKAGES} | awk '/^kernel-/{print $1}')" \
-    && find /host/build/rpms/ -maxdepth 2 -type f \
+    && find /host/build/ -type f \
```
bcressey, line 379: This is not ideal and needs at least an issue in the backlog to fix.

It's also only safe if we are 100% sure that older / different versions of the core kit are cleaned up. Otherwise we might copy the wrong RPM.

At a minimum we should filter the name to `bottlerocket-${KERNEL}-archive-*.${ARCH}.rpm`.

jmt-lab: Should I change it to the above?

bcressey: Yes, please do.

tools/buildsys/src/builder.rs
```diff
@@ -571,6 +571,7 @@ impl DockerBuild {
             "build {context} \
             --target {target} \
             --tag {tag} \
+            --compress=false \
```
bcressey: I'm not confident that this has much or any impact, so we can drop it for now.

jmt-lab:
Addressed comments

bcressey (approved):

webern (approved):

PR #302, opened 2024-06-18
switch to pipesys for build input and output

bcressey:
**Issue number:**
N/A

**Description of changes:**
About a year ago, as a weekend hack, I set out to solve the problem of passing make jobserver file descriptors into a build in order to remove the `BUILDSYS_JOBS = "8"` constraint in favor of more dynamic parallelism that would use all the CPUs available on the build host in an intelligent way. `pipesys` was the solution I hit upon, but the performance benefits weren't clear cut, and there were a lot of annoying edge cases across the different `make` invocations in various packages.

While benchmarking, I observed that the greatest impediment to additional parallelism was not efficient use of all available CPU cores, but rather that the Docker daemon struggled under the increased load of additional parallel jobs, with each one copying a slightly different version of a multi-gigabyte build context as input, and compressing the intermediate stages on the way to delivering the package and image outputs.

I realized that if I could pass jobserver file descriptors from `make`, I could also pass directory file descriptors for the project workspace. That provided far more obvious wins even without adding more simultaneous package builds: it sped up builds and reduced `dockerd` CPU and memory usage. I also realized I could pass a file descriptor for the output directory, and avoid the need to spin up a container just to copy out the build artifacts. Accordingly, I redesigned `pipesys` around the use case of build inputs and outputs.

The commits here are best read in reverse order. My ultimate goal was to change the `.dockerignore` file to exclude almost everything from the build context. The Dockerfile changes, while extensive, mostly amount to changing the source of input files from "`/host`" to "`/bypass`", and to write output files directly to "`/output`" rather than leaving that up to Docker. (The most complicated parts of `pipesys` are in the "link" command, and that's done to support the very convenient `Dockerfile` interface.)

The `buildsys` changes are relatively straightforward: we need to serve two file descriptors - one for input, one for output - via `pipesys`. We can also dispense with caching for intermediate layers and switch back to a single `Dockerfile`.

`unplug` is a simple program that offers just the isolation semantics we need - no network access - without  requiring any additional privileges in the build environment.

`pipesys` is complex new code, but may be easier to follow once the specific uses are understood. 

**Testing done:**
Built https://github.com/bottlerocket-os/bottlerocket-core-kit. Before, the build was so flaky that it was difficult to finish building all packages, and the core kit took several minutes to create. After, the build was reliable across a nightly loop of `make twoliter clean && make`, and the core kit takes around 20 seconds to create.

**Terms of contribution:**


bcressey:
Definitely curious to see if this works as expected in the CI environment!

bcressey:
The irony is not lost on me that `unplug` - what I called the "simple program" - was responsible for the above two force pushes. Just to add the license statement and appease `cargo fmt`, but still not a promising start.

bcressey:
⬆️ force push fixes a couple of bugs with variant builds.

The more significant change is the addition of `rm -rf /output/*` to the various stages that interact with the output directory. What's going on here is that if a build fails, it's possible for the directory to contain root-owned artifacts that were partly written and not yet `chown`'d by one of the parallel scripts (`rpm2img`, `rpm2migration`, `rpm2kmodkit`).

These artifacts are slightly annoying for `buildsys` to clean up because it is (presumably) running as a non-root user, which defeats a simple `std::fs::remove_dir_all`.

Doing it in the Dockerfile means that clean up can happen as root, and then all of the changes as far as adding, removing, and chmod'ing contents happen in the same place.

bcressey:
Oh but now there is a bug in the `imgbuild` stage since the three `rm -rf /output/*` calls are unsynchronized; if one stage runs significantly later than the others, it may remove content that's been newly written.

Fix incoming ...

tools/buildsys/src/builder.rs
```diff
@@ -571,31 +578,71 @@ impl DockerBuild {
...
 
         build.extend(self.build_args());
         build.extend(self.secrets_args.clone());
 
-        let create = format!("create --name {} {} true", self.tag, self.tag).split_string();
-        let cp = format!("cp {}:/output/. {}", self.tag, marker_dir.display()).split_string();
-        let rm = format!("rm --force {}", self.tag).split_string();
-        let rmi = format!("rmi --force {}", self.tag).split_string();
+        // Run a container with the project's root as a read-only volume mount, so that pipesys can
+        // serve a read-only file descriptor that's safe to pass into builds.
```
cbgbt, line 595: Cool idea.

tools/pipesys/src/cmd/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+                )
+                .init();
+        }
+    }
+}
+
+/// Helper function to retrieve file descriptors via an abstract socket.
+fn fetch_fds(socket: &str, wanted: usize) -> Result<Vec<i32>> {
+    let addr = uds::UnixSocketAddr::from_abstract(socket.as_bytes())
+        .with_context(|| format!("failed to create socket {}", socket))?;
```
cbgbt: Seems like `wanted` doesn't actually behave like expected. Should probably alter the function or else just always fetch one FD as a constant.

tools/pipesys/src/cmd/link.rs
```diff
@@ -0,0 +1,194 @@
...
+    #[clap(long = "target")]
+    target: PathBuf,
+}
+
+impl Link {
+    /// Retrieve the file descriptor, then spawn a background process to create the link so that it
+    /// survives the return of the foreground process.
+    pub(crate) async fn execute(&self) -> Result<()> {
+        // Remove the existing symlink, if present.
+        let _ = fs::remove_file(&self.target).await;
```
cbgbt: Should we be more defensive when we remove the target?

I'm not sure if we can assert that it is a pipesys-managed link here, but we could maybe refuse to delete non-symlinks.

tools/pipesys/src/cmd/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+    let client = uds::UnixSeqpacketConn::connect_unix_addr(&addr)
+        .with_context(|| format!("failed to connect to socket {}", socket))?;
+
+    let mut fd_buf = [-1; 8];
+    let (_, _, fds) = client
+        .recv_fds(&mut [0u8; 1], &mut fd_buf)
```
cbgbt, line 85: I didn't know that Linux could do this. That's awesome.

tools/pipesys/src/cmd/link.rs
```diff
@@ -0,0 +1,194 @@
...
+    check: &dyn Fn(&'a Path) -> F,
+) -> Result<()>
+where
+    F: Future<Output = bool> + 'a,
+{
+    // Check whether the provided function succeeds right away, in case the watched file was
+    // created or removed before the inotify watch could be started.
+    if check(path).await {
+        return Ok(());
+    }
```
cbgbt, line 184: Do we need to perform this check after the `event_stream` is created? I'm imagining a very short race window otherwise.

cbgbt (commented):
Nice!

The multi-fd support in pipesys would be more useful if there was some scheme for providing abstract names to the descriptors so that clients could understand what the returned descriptors mapped to.

Since we're only interested in serving a single descriptor per-process, maybe we should reduce the scope of the program to focus on one descriptor?

bcressey:
> Since we're only interested in serving a single descriptor per-process, maybe we should reduce the scope of the program to focus on one descriptor?

Sounds good. The multi-fd stuff is left over from the initial goal of passing `make`'s jobserver fds, where there are two (the read & write ends of a pipe).

tools/pipesys/src/cmd/link.rs
bcressey: > Do we need to perform this check after the event_stream is created? I'm imagining a very short race window otherwise.

My understanding is that the kernel will publish events to the watch descriptor as soon as the `inotify.watches.add(...)` returns. On the Rust side, we won't start reading those events until we poll the event stream, but we shouldn't miss any.

tools/pipesys/src/server.rs
```diff
@@ -0,0 +1,114 @@
+use anyhow::{Context, Result};
+use clap::Parser;
+use log::warn;
+use std::fs::OpenOptions;
+use std::os::fd::{AsRawFd, RawFd};
+use std::path::{Path, PathBuf};
+use uds::{tokio::UnixSeqpacketListener, UnixSocketAddr};
+
+#[derive(Clone, Debug, Parser)]
+pub struct Server {
```
webern:
```suggestion

/// Describe what the program does.
#[derive(Clone, Debug, Parser)]
pub struct Server {
```

tools/pipesys/src/cmd/link.rs
cbgbt: When I reviewed that function I forgot that it was called in the context of already having called `inotify_init`. Makes sense!

tools/buildsys/src/builder.rs
```diff
@@ -25,6 +26,7 @@ use std::collections::HashSet;
 use std::env;
 use std::fs::{self, read_dir, File};
 use std::num::NonZeroU16;
+use std::os::unix::fs::MetadataExt;
```
webern, line 29: Things start becoming interesting (I hadn't thought of this before), if `buildsys` needs to be compiled for Linux (because it runs in a container, e.g. if cargo make and cargo are used from a "bootstrap" container, as I used to call it.) and `twoliter` runs on a host that isn't Linux.

twoliter/embedded/build.Dockerfile.dockerignore
```diff
@@ -1,17 +1,16 @@
...
-!/build/tools/*
-**/target/*
-/sbkeys
-/tests
+# Exclude everything from the build context by default.
+*
+
+# Include the specific files expected by the Dockerfile.
+!/build/tools
+!/packages/*/*.spec
```
webern, line 6: Why are these unignored but, for example, tar.gz files are not?

tools/pipesys/src/cmd/link.rs
```diff
@@ -0,0 +1,194 @@
...
+    let mut event_stream = watcher
+        .into_event_stream(&mut buf)
+        .with_context(|| "failed to get event stream")?;
+
+    while let Some(event) = event_stream.next().await {
+        let event = event.with_context(|| "failed to read event".to_string())?;
+        if let Some(event_file_name) = event.name {
+            if event_file_name == file_name {
+                trace!("found event for {}", path.display());
+                if check(path).await {
```
webern, line 200: Is this `fs`?

```suggestion
                if fs::check(path).await {
```

tools/unplug/src/main.rs
```diff
@@ -0,0 +1,60 @@
...
+    if let Some(program) = args.next() {
+        let ret = Command::new(&program)
+            .args(args)
+            .status()
+            .with_context(|| format!("failed to run {program}"))?;
+        let code = ret.code().unwrap_or(1) as u8;
+        return Ok(code.into());
+    }
+
+    Ok(0.into())
```
webern: nit, since most of what you're doing is way beyond me.

```suggestion
    Ok(ExitCode::Success)
```

twoliter/embedded/build.Dockerfile
```diff
@@ -65,46 +65,53 @@ RUN \
...
       ln -snft ./rpmbuild/RPMS {} \+ && \
     for pkg in ${PACKAGE_DEPENDENCIES} ; do \
-      [ -d "/host/build/rpms/${pkg}" ] || continue ; \
-      find /host/build/rpms/${pkg}/ -mindepth 1 -maxdepth 1 -name '*.rpm' -size +0c -print -exec \
+      [ -d "/bypass/build/rpms/${pkg}" ] || continue ; \
+      find /bypass/build/rpms/${pkg}/ -mindepth 1 -maxdepth 1 -name '*.rpm' -size +0c -print -exec \
         ln -snft ./rpmbuild/RPMS {} \+ ; \
     done && \
     createrepo_c \
-        -o ./rpmbuild/RPMS \
```
webern, line 71: nit: but making the diff bigger than it needs to be is probably not the best in a complicated PR like this. I would suggest re-format the Dockerfile in a subsequent PR.

twoliter/embedded/repack.Dockerfile
```diff
@@ -1,69 +0,0 @@
-# syntax=docker/dockerfile:1.4.3
```
webern, line 1: This commit seems to be missing the delete of repack.Dockerfile.dockerignore `twoliter: combine build and repack Dockerfiles`

twoliter/embedded/repack.Dockerfile.dockerignore
```diff
@@ -1,4 +0,0 @@
-/
```
webern, line 1: Maybe it's just a git thing, but this file delete is in the commit after the one that removed it from Twoliter's embedding.

webern (approved):

twoliter/embedded/build.Dockerfile.dockerignore
bcressey: > Why are these unignored but, for example, tar.gz files are not?

Spec files are tiny, and - like files under sources - are somewhat likely to be edited while a build is in progress.

Including these in the build context has the effect of snapshotting them at the moment when the build starts (more or less); copying them from the bypass directory would mean that edits would go "live" in a way that could lead to confusing build errors.

tools/pipesys/src/cmd/link.rs
bcressey: > Is this `fs`?

`check` is a function parameter - maybe it needs a better name?

tools/buildsys/src/builder.rs
bcressey: > Things start becoming interesting (I hadn't thought of this before), if buildsys needs to be compiled for Linux ...

Good point. It doesn't seem likely that cargo's bindeps will ever support building a binary for a different target than the crate is being built for, so we'd need to approach the build rather differently to get a Mac or Windows binary with Linux tools embedded.

bcressey: For now I've gone with the approach of making everything build on non-Linux, while leaving the actual logic unimplemented for `pipesys` and `unplug`. That way it's still possible to develop Twoliter on a different OS, though building kits or variants will be Linux-only.

cbgbt (approved):
Looks good, though I'm not sure I see how you resolved this:
> Oh but now there is a bug in the imgbuild stage since the three rm -rf /output/* calls are unsynchronized; if one stage runs significantly later than the others, it may remove content that's been newly written.

Does it have something to do with the modified stage dependencies?

bcressey:
> Does it have something to do with the modified stage dependencies?

Now the output cleanup happens just once in the `repobuild` stage, which is used by the other three stages for building a variant (`imgbuild`, `kmodkitbuild`, `migrationbuild`). `migrationbuild` now depends on `repobuild` like the other two.

tools/pipesys/src/cmd/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+mod link;
+
+use self::link::Link;
+use pipesys::server::Server as Serve;
+
+use anyhow::Result;
+#[cfg(target_os = "linux")]
+use anyhow::{ensure, Context};
+use clap::Parser;
+use env_logger::Builder;
```
sam-berning: nit: I think it'd be slightly more readable not to `use` this since it's such a generic name

sam-berning (approved):

webern (approved):

bcressey:
⬆️ force push switches to `env_logger::Builder` rather than importing it.

sam-berning (approved):

PR #305, opened 2024-06-18
twoliter: add support for crane

jmt-lab:
**Description of changes:**

* Adds support for crane if it is available.

**Testing done:**

Testing is still underway, building core kit locally to use for tests on my dev instance

**Terms of contribution:**

### DIFF
--- a/tools/oci-cli-wrapper/src/lib.rs
+++ b/tools/oci-cli-wrapper/src/lib.rs
@@ -1,0 +1,131 @@
+//! ImageTool enablement library implements a standardized way of calling commandline container image
+//! tools for interacting primarily with kit images in a container registry.
+//!
+//! Current two tools are supported:
+//! * crane, gcrane, krane
+//!     Crane provides a more direct interaction with the container registry,
+//!     allowing us to query image information in the registry without having to pull the full image to
+//!     disk. It also does not require a daemon to operate and has optimizations for pulling large images to disk
+//! * docker
+//!     Docker can perform all interactions we need with several caveats that make it less efficient than
+//!     crane. The image needs to be pulled locally in order for docker to inspect the manifest and extract
+//!     metadata. In addition, in order to operate with OCI image format, the containerd-snapshotter
+//!     feature has to be enabled in the docker daemon
+use std::{collections::HashMap, env, path::Path, rc::Rc};
+
+use async_trait::async_trait;
+use cli::CommandLine;
+use crane::CraneCLI;
+use docker::DockerCLI;
+use serde::Deserialize;
+use snafu::ResultExt;
+use which::which;
+
+mod cli;
+mod crane;
+mod docker;
+
+#[async_trait]
+pub trait ImageTool {
+    /// Pull an image archive to disk
+    async fn pull_oci_image(&self, path: &Path, uri: &str) -> Result<()>;
+    /// Fetch the manifest
+    async fn get_manifest(&self, uri: &str) -> Result<Vec<u8>>;
+    /// Fetch the image config
+    async fn get_config(&self, uri: &str) -> Result<ConfigView>;
+}
+
+/// Auto-select the container tool to use by environment variable
+/// and-or auto detection
+pub fn image_tool() -> Result<Rc<dyn ImageTool>> {
+    if let Ok(name) = env::var("TWOLITER_KIT_IMAGE_TOOL") {
+        return match name.as_str() {
+            "docker" => Ok(Rc::new(DockerCLI {
+                cli: CommandLine {
+                    path: which("docker").context(error::NotFoundSnafu { name: "docker" })?,
+                },
+            })),
+            tool @ ("crane" | "gcrane" | "krane") => Ok(Rc::new(CraneCLI {
+                cli: CommandLine {
+                    path: which(tool).context(error::NotFoundSnafu { name: tool })?,
+                },
+            })),
+            _ => error::UnsupportedSnafu { name }.fail(),
+        };
+    }
+    let crane = which("krane").or(which("gcrane")).or(which("crane"));
+    if let Ok(path) = crane {
+        return Ok(Rc::new(CraneCLI {
+            cli: CommandLine { path },
+        }));
+    };
+    Ok(Rc::new(DockerCLI {
+        cli: CommandLine {
+            path: which("docker").context(error::NoneFoundSnafu)?,
+        },
+    }))
+}
+
+#[derive(Deserialize, Debug)]
+#[serde(rename_all = "snake_case")]
+struct ImageView {
+    config: ConfigView,
+}
+
+#[derive(Deserialize, Debug)]
+#[serde(rename_all = "PascalCase")]
+pub struct ConfigView {
+    pub labels: HashMap<String, String>,
+}
+
+pub type Result<T> = std::result::Result<T, error::Error>;
+
+pub mod error {
+    use std::path::PathBuf;
+
+    use snafu::Snafu;
+
+    #[derive(Snafu, Debug)]
+    #[snafu(visibility(pub(super)))]
+    pub enum Error {
+        #[snafu(display("Failed to extract archive: {source}"))]
+        ArchiveExtract { source: std::io::Error },
+
+        #[snafu(display("Failed to read archive: {source}"))]
+        ArchiveRead { source: std::io::Error },
+
+        #[snafu(display("Failed to execute image tool, {message}: {source}"))]
+        CommandFailed {
+            message: String,
+            source: std::io::Error,
+        },
+
+        #[snafu(display("Failed to deserialize image config: {source}"))]
+        ConfigDeserialize { source: serde_json::Error },
+
+        #[snafu(display("Failed to create temporary directory for docker save: {source}"))]
+        DockerTemp { source: std::io::Error },
+
+        #[snafu(display(
+            "Unable to find any supported container image tool, please install docker or crane: {}",
+            source
+        ))]
+        NoneFound { source: which::Error },
+
+        #[snafu(display(
+            "Unable to find a container image tool by name '{}' in current environment",
+            name
+        ))]
+        NotFound { name: String, source: which::Error },
+
+        #[snafu(display("Failed to run operation with image tool: {message}\n command: {} {}", program.display(), args.join(" ")))]
+        OperationFailed {
+            message: String,
+            program: PathBuf,
+            args: Vec<String>,
+        },
+
+        #[snafu(display("Unsupported container image tool '{}'", name))]
+        Unsupported { name: String },
+    }
+}
--- a/twoliter/Cargo.toml
+++ b/twoliter/Cargo.toml
@@ -12,6 +12,7 @@ exclude = ["/design", "/target", "/dockerfiles", "/scripts"]
 [dependencies]
 anyhow = "1"
 async-recursion = "1"
+async-trait = "0.1"
 async-walkdir = "1"
 base64 = "0.22"
 buildsys-config = { version = "0.1", path = "../tools/buildsys-config" }
@@ -23,6 +24,7 @@ futures= "0.3"
 hex = "0.4"
 log = "0.4"
 non-empty-string = { version = "0.2", features = [ "serde" ] }
+oci-cli-wrapper = { version = "0.1", path = "../tools/oci-cli-wrapper" }
 olpc-cjson = "0.1"
 semver = { version = "1", features = ["serde"] }
 serde = { version = "1", features = ["derive"] }
@@ -33,6 +35,7 @@ tempfile = "3"
 tokio = { version = "1", default-features = false, features = ["fs", "macros", "process", "rt-multi-thread"] }
 toml = "0.8"
 uuid = { version = "1", features = [ "v4" ] }
+which = "6"
 
 # Binary dependencies. These are binaries that we want to embed in the Twoliter binary.
 buildsys = { version = "0.1.0", artifact = [ "bin:buildsys", "bin:bottlerocket-variant" ], path = "../tools/buildsys" }
--- a/twoliter/src/lock.rs
+++ b/twoliter/src/lock.rs
@@ -4,6 +4,7 @@ use crate::schema_version::SchemaVersion;
 use anyhow::{ensure, Context, Result};
 use base64::Engine;
 use buildsys_config::DockerArchitecture;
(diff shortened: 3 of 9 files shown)


twoliter/Cargo.toml
```diff
@@ -12,34 +12,50 @@ exclude = ["/design", "/target", "/dockerfiles", "/scripts"]
...
 semver = { version = "1", features = ["serde"] }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 sha2 = "0.10"
 tar = "0.4"
 tempfile = "3"
-tokio = { version = "1", default-features = false, features = ["fs", "macros", "process", "rt-multi-thread"] }
+tokio = { version = "1", default-features = false, features = [
+    "fs",
+    "macros",
```
bcressey: nit: can you turn off the format changes here?

twoliter/src/lock.rs
```diff
@@ -399,11 +367,13 @@ impl Lock {
...
 
-    async fn get_manifest(&self, image: &LockedImage, arch: &str) -> Result<ManifestView> {
-        let manifest_bytes = docker!(
-            ["manifest", "inspect", image.source.as_str()],
-            format!("failed to find a kit {}", image.to_string())
-        );
+    async fn get_manifest(
+        &self,
+        cli: ImageTool,
+        image: &LockedImage,
```
cbgbt, line 362: Why not `&dyn ImageToolImpl` (or just rename the type to `ImageTool` and `&dyn ImageTool`)?

This is fine, though whenever I will read this without context, I will say "strange that this function takes ownership of an `ImageTool` and drops it" and then only discover the truth when I dig around to find that `ImageTool` is an `Rc`.

twoliter/src/lock.rs
```diff
@@ -216,24 +191,16 @@ impl OCIArchive {
         self.cache_dir.join(self.digest.replace(':', "-"))
     }
 
-    async fn pull_image(&self) -> Result<()> {
+    async fn pull_image(&self, cli: ImageTool) -> Result<()> {
```
cbgbt: readability nit
```suggestion
    async fn pull_image(&self, image_tool: ImageTool) -> Result<()> {
```

Although this has a name collision with the function (maybe `discover_image_tool`?)

tools/image-tool/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+mod docker;
+
+#[async_trait]
+pub trait ImageToolImpl {
+    /// Get the path of the cli tool to call
+    fn path(&self) -> PathBuf;
+
+    /// Pull an image archive to disk
+    async fn pull_archive(&self, file: &Path, uri: &str, cache: &Path) -> Result<()>;
+    /// Fetch the manifest
```
cbgbt: nit: inconsistent whitespace

tools/image-tool/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+    async fn get_manifest(&self, uri: &str) -> Result<Vec<u8>>;
+    /// Fetch the image config
+    async fn get_config(&self, uri: &str) -> Result<Vec<u8>>;
+
+    async fn output(&self, args: &[&str], error_msg: String) -> Result<Vec<u8>> {
```
cbgbt: I wonder if we should avoid exposing this in the trait? Seems like we want callers to use one of the abstract methods, and that this `output` function is more of an implementation detail for the impls we have. (same with `spawn`)

tools/image-tool/src/docker.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    async fn get_manifest(&self, uri: &str) -> Result<Vec<u8>> {
+        self.output(
+            &["manifest", "inspect", uri],
+            format!("failed to inspect manifest of resource at {}", uri),
+        )
+        .await
+    }
+
+    async fn get_config(&self, uri: &str) -> Result<Vec<u8>> {
```
cbgbt: Per https://github.com/bottlerocket-os/twoliter/pull/306 I suppose we have to pull first :(

tools/image-tool/src/crane.rs
```diff
@@ -0,0 +1,51 @@
+use std::path::{Path, PathBuf};
```
cbgbt: Module-level docstring comment referencing `crane`'s documentation/home and explaining why we support it here is probably worth it.

tools/image-tool/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+        Ok(())
+    }
+}
+
+pub type ImageTool = Rc<dyn ImageToolImpl>;
+
+/// Auto-select the container tool to use by environment variable
+/// and-or auto detection
+pub fn image_tool() -> Result<ImageTool> {
+    if let Ok(name) = env::var("KIT_IMAGE_TOOL") {
```
bcressey: nit: prefix with our tool name:
```suggestion
    if let Ok(name) = env::var("TWOLITER_KIT_IMAGE_TOOL") {
```

tools/image-tool/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+        return match name.as_str() {
+            "docker" => Ok(Rc::new(DockerCLI {
+                path: which("docker").context(error::NotFoundSnafu { name: "docker" })?,
+            })),
+            "crane" => Ok(Rc::new(CraneCLI {
```
bcressey: We should allow the other tools from the `crane` family:
```suggestion
            "crane" | "krane" | "gcrane" => Ok(Rc::new(CraneCLI {
```

bcressey (commented):
Looks good. Not much beyond what @cbgbt suggested.

twoliter/src/lock.rs
jmt-lab: So I had to wrap it into an Rc to avoid async lifetime issues, aka need for &'static dyn ImageToolImpl

tools/image-tool/src/lib.rs
jmt-lab: DO they all have the same cli command set?

bcressey: > Do they all have the same cli command set?

I believe that `crane` and `krane` do, and `gcrane` has additional commands on top of the shared set.

jmt-lab:
Working on fixing the duplicate collision

tools/image-tool/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+            })),
+            tool @ ("crane" | "gcrane" | "krane") => Ok(Rc::new(CraneCLI {
+                cli: CommandLine {
+                    path: which(tool).context(error::NotFoundSnafu { name: tool })?,
+                },
+            })),
+            _ => error::UnsupportedSnafu { name }.fail(),
+        };
+    }
+    let crane = which("kcrane").or(which("gcrane")).or(which("crane"));
```
sam-berning:
```suggestion
    let crane = which("krane").or(which("gcrane")).or(which("crane"));
```

sam-berning: how did you decide the order of precedence here?

twoliter/src/lock.rs
```diff
@@ -69,12 +42,9 @@ pub(crate) struct LockedImage {
 }
 
 impl LockedImage {
-    pub async fn new(vendor: &Vendor, image: &Image) -> Result<Self> {
+    pub async fn new(cli: Rc<dyn ImageTool>, vendor: &Vendor, image: &Image) -> Result<Self> {
```
```suggestion
    pub async fn new(image_tool: Rc<dyn ImageTool>, vendor: &Vendor, image: &Image) -> Result<Self> {
```

tools/image-tool/src/crane.rs
```diff
@@ -0,0 +1,41 @@
...
+    async fn get_manifest(&self, uri: &str) -> Result<Vec<u8>> {
+        self.cli
+            .output(
+                &["manifest", uri],
+                format!("failed to fetch manifest for resource at {}", uri),
+            )
+            .await
+    }
+
+    async fn get_config(&self, uri: &str) -> Result<Vec<u8>> {
```
sam-berning: as neat as this particular function is I think it may make more sense for this interface to be `get_image_labels` instead of `get_config`. the `docker` and `crane` implementations of this are returning different JSON objects that just happen to both contain a field named either "config"/"Config" that also happen to contain a field called "Labels". IMO the sort-of duplicated code is more robust than having these impls return completely different JSON strings.

tools/image-tool/src/crane.rs
```diff
@@ -0,0 +1,41 @@
...
+
+use crate::{cli::CommandLine, ImageTool, Result};
+
+pub struct CraneCLI {
+    pub(crate) cli: CommandLine,
+}
+
+#[async_trait]
+impl ImageTool for CraneCLI {
+    async fn pull_archive(&self, path: &Path, uri: &str, _cache: &Path) -> Result<()> {
```
sam-berning: nit: this name isn't totally accurate since it pulls the extracted version right? maybe `pull_oci_image` would be better?

tools/image-tool/src/lib.rs
jmt-lab: From what Ben told me about capability krane and gcrane are supersets of crane so i figured I would try and go more capable -> less capable. Honestly for our purposes the precedence doesn't really matter

tools/image-tool/src/crane.rs
jmt-lab: Talked offline, we landed on a compromise of returning the actual Config object that is shared between both data types and is consistent

tools/image-tool/src/cli.rs
```diff
@@ -0,0 +1,47 @@
...
+            .args(args)
+            .output()
+            .await
+            .context(error::CommandFailedSnafu { message: error_msg })?;
+        ensure!(
+            output.status.success(),
+            error::OperationFailedSnafu {
+                message: String::from_utf8_lossy(&output.stderr)
+            }
+        );
```
bcressey: It'd be good to give the full CLI invocation - path, args - in addition to this message, since that's valuable context that may be obscured, and could help troubleshoot outside of `twoliter`.

tools/image-tool/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "image-tool"
```
bcressey: nit: `image-tool` is a pretty generic name - maybe `oci-cli-wrapper` or something like that.

twoliter/src/lock.rs
```diff
@@ -198,42 +172,41 @@ struct OCIArchive {
...
-        Self {
+        Ok(Self {
             image: image.clone(),
             digest: digest.into(),
             cache_dir: cache_dir.as_ref().to_path_buf(),
-        }
+            temp: TempDir::new_in(cache_dir.as_ref()).context(format!(
+                "failed to create temporary directory in {}",
+                cache_dir.as_ref().display()
+            ))?,
```
bcressey: This temp dir bothers me for a couple of reasons. I don't think it's very well-named since it ends up being the primary way that blobs and the manifest are passed back to the caller.

The docker treatment is a bit strange, you end up creating another temp dir inside the cache dir, where the contents are unpacked to the provided dir, which is this temp dir passed from the caller.

The crane CLI doesn't use the cache dir at all and just populates the provided dir, which is this temp dir inside the cache dir.

I would try to reduce this to just one directory that gets passed around, and maybe `cache_dir` can be `work_dir` - a temporary directory that's created by the caller that instantiates the `OCIArchive`, on the same filesystem as the eventual output directory, with the contract "it is OK to write into this directory, including creating any temporary dirs needed, but you're expected to at least create a `index.json` and some files under `blobs`.

twoliter/src/lock.rs
```diff
@@ -430,11 +404,11 @@ impl Lock {
...
 
         // First get the manifest for the specific requested architecture
-        let manifest = self.get_manifest(image, arch).await?;
-        let oci_archive = OCIArchive::new(image, manifest.digest.as_str(), &cache_path);
+        let manifest = self.get_manifest(image_tool.clone(), image, arch).await?;
+        let oci_archive = OCIArchive::new(image, manifest.digest.as_str(), &cache_path)?;
 
         // Checks for the saved image locally, or else pulls and saves it
-        oci_archive.pull_image().await?;
+        oci_archive.pull_image(image_tool.clone()).await?;
```
bcressey, line 402: I am surprised and a bit suspicious that we're not passing `arch` here. Both `docker` and `crane` allow the use of a `--platform` argument when pulling.

jmt-lab: Yes that is true but if we have the image digest (which at this point we do) the platform is not needed as the image digest here refers only to a specific architecture image.

twoliter/src/lock.rs
jmt-lab: Not sure if my change does what you had intended, but I realized that the two directories had no real purpose anymore. So i kept it to one and actually removed it as a tempdir on the OCIArchive side because turns out using the temp dir prevented that digest caching check from even working..woops. Now only the docker approach creates a temp directory for saving the archive out so that the appropriate cache directory is created

tools/oci-cli-wrapper/src/lib.rs
```diff
@@ -0,0 +1,126 @@
+//! ImageTool enablement library implements a standardized way of calling commandline container image
+//! tools for interacting primarily with kit images in a container registry.
+//!
+//! Current two tools are supported:
+//! * crane, gcrane, kcrane
+//!     Crane provides a more direct interaction with the container registry, allowing us to query image information in the registry without
+//!     having to pull the full image to disk. It also does not require a daemon to operate and has optimizations for pulling large images to disk
```
bcressey: nit: these lines seem quite long

tools/oci-cli-wrapper/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+//! * docker
```
bcressey, line 9: nit: no love for Docker? we could mention that the containerd snapshotter must be enabled, or talk about why it's less efficient (save/load cycle, have to pull before inspecting, etc)

bcressey (approved):

sam-berning (approved):

jmt-lab:
Fixed caching of kit images and overuse of temp directories. Reran tests with both docker and crane ensuring caching and the expected directory works

jmt-lab:
Fix comments per bens comments

sam-berning (approved):

sam-berning (approved):

PR #306, opened 2024-06-19
twoliter: add a pull so we can inspect the image config

jmt-lab:
**Description of changes:**

Docker does not allow you to inspect the image config without pulling it locally. For now we need to pull at least one architecture image.

**Terms of contribution:**

### DIFF
--- a/twoliter/src/lock.rs
+++ b/twoliter/src/lock.rs
@@ -513,44 +513,43 @@ impl Lock {
     async fn find_kit(vendor: &Vendor, image: &LockedImage) -> Result<ImageMetadata> {
         let manifest_list: ManifestListView = serde_json::from_slice(image.manifest.as_slice())
             .context("failed to deserialize manifest list")?;
-        let mut encoded_metadata: Option<String> = None;
-        for manifest in manifest_list.manifests.iter() {
-            let image_uri = format!("{}/{}@{}", vendor.registry, image.name, manifest.digest);
-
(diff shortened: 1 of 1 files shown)


sam-berning:
Are we only checking one architecture's labels now to save time/space?

twoliter/src/lock.rs
```diff
@@ -514,40 +514,50 @@ impl Lock {
...
+            "could not deserialize labels on the image for {}",
+            image
+        ))?;
+        let encoded = labels
+            .get("dev.bottlerocket.kit.v1")
+            .context("no metadata stored on image, this image appears to not be a kit")?;
+        if let Some(metadata) = encoded_metadata.as_ref() {
+            ensure!(
+                encoded == metadata,
+                "metadata does match between images in manifest list"
```
bcressey: Seems like this branch won't be taken since we're only pulling one image rather than looping over both? If so then you can simplify this and just keep the original `let encoded = ...` assignment.

bcressey (approved):

cbgbt (approved):

twoliter/src/lock.rs
webern:
```suggestion
                "metadata does not match between images in manifest list"
```

webern (approved):

bcressey (approved):

cbgbt (approved):

PR #313, opened 2024-06-23
Respect GO_MODULES environment variable when preparing modules from `sources` directory

### LINKED ISSUE #312: twoliter ignores `GO_MODULES` environment variable
When building variants, twoliter currently finds all golang modules that live under `sources` and prepares those modules to be used during the build process. In some cases, there may be sources that the author does not intend to use during the build.

It would be useful if the `GO_MODULES` environment variable could be supplied to explicitly set the modules that should be used during the build.

cbgbt:
**Issue number:**

Closes #312

**Description of changes:**
* twoliter: allow overriding GO_MODULES
* docker-go: fix GOPATH to be set after arg parse

**Testing done:**
* Provided unit tests pass

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/docker-go
+++ b/twoliter/embedded/docker-go
@@ -51,12 +51,12 @@ parse_args() {
   required_arg "--command" "${COMMAND}"
 }
 
-# We need to mount the ../.. parent of GO_MOD_CACHE
-GOPATH=$(cd "${GO_MOD_CACHE}/../.." && pwd)
-
 DOCKER_RUN_ARGS="--network=host"
 
 parse_args "${@}"
+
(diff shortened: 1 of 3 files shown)


cbgbt:
I'm pretty sure my tests failed because tests are run concurrently and I'm modifying an environment variable during the test. I'll come up with a fix.

cbgbt:
I've decided that testing the environment variable modification in Rust unit tests is a terrible idea if we would like to have a future without flaky tests.

Instead, the environment searching feature is extracted to an additional function which I have tested manually.

bcressey (approved):

sam-berning (approved):

PR #325, opened 2024-07-12
Update to latest nightly, cargo dep update and remove workflow hacks

jmt-lab:
**Description of changes:**

* updates nightly to 2024-07-11 so new used features in serde exist
* updates cargo dependencies to ensure time is updated to a suported version for the nightly
* removes previous workarounds in workflows

**Terms of contribution:**

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -119,7 +119,7 @@ jobs:
       - name: Install Cross
         # Pin cargo cross to a version that we know is working for us.
         run: |
-          cargo install cross --locked --git https://github.com/cross-rs/cross/ --rev d6511b7b
+          cargo install cross --git https://github.com/cross-rs/cross/ --rev d6511b7b --locked
       - name: Run cargo-dist
         # This logic is a bit janky because it's trying to be a polyglot between
         # powershell and bash since this will run on windows, macos, and linux!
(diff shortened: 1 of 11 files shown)


sam-berning (approved):

.github/workflows/release.yml
```diff
@@ -60,7 +60,7 @@ jobs:
           submodules: recursive
       - name: Install cargo-dist
         run: |
-          cargo install --locked \
+          cargo install \
             --git "https://github.com/webern/cargo-dist" \
```
cbgbt: Locking dependencies in these cases seems like a good idea to me. Should we really be removing this?

.github/workflows/rust.yml
```diff
@@ -22,6 +22,6 @@ jobs:
       labels: bottlerocket_ubuntu-latest_16-core
     steps:
       - uses: actions/checkout@v3
-      - run: cargo install --locked cargo-deny
-      - run: cargo install --locked cargo-make@0.37.12
```
cbgbt, line 26: I think we originally locked to `0.37.12` for `cargo-make` due to a tendency to introduce breaking changes upstream. Might be worth continuing to require a specific version.

cbgbt (approved):
This mostly LGTM, though I think we will be happier if we maintain a versioned dependency on `cargo-make`, even if the version changes.

.github/workflows/rust.yml
jmt-lab: Actually in this case I added this version lock before realizing that we needed to bump nightly. The newer patch of cargo-make would pull in the serde version that was incompatible with the old nightly

cbgbt: Ah, I just misremembered. Forget my feedback then! 🚀

PR #336, opened 2024-07-26
imghelper: remove full path from .vmlinuz.hmac

jpculp:
**Description of changes:**

When hmac generation moved to `imghelper`, the `vmlinuz` variable went from the basename to the full-path. This makes sure that the full-path doesn't make it into the final artifact.

Also adds a missing ShellCheck exception.

**Testing done:**

Built and smoke tested Bottlerocket.

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/imghelper
+++ b/twoliter/embedded/imghelper
@@ -478,6 +478,7 @@ sbsetup_signing_profile() {
 undo_sign() {
   local what
   what="${1:?}"
+  # shellcheck disable=SC2076 # literal match is intended.
   if [[ ! "$(pesign -i "${what}" -l)" =~ 'No signatures found.' ]]; then
     mv "${what}" "${what}.orig"
     pesign -i "${what}.orig" -o "${what}" -u 0 -r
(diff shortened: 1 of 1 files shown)


bcressey (approved):

twoliter/embedded/imghelper
```diff
@@ -574,7 +575,7 @@ generate_hmac() {
   local vmlinuz
   vmlinuz="${1:?}"
   openssl sha512 -hmac FIPS-FTW-RHT2009 -hex "${vmlinuz}" |
-    awk -v vmlinuz="${vmlinuz}" '{ print $2 "  " vmlinuz }' \
+    awk '{ print $2 "  vmlinuz" }' \
```
jpculp, line 578: I wonder if this would be better.
```suggestion
    awk -v vmlinuz="${vmlinuz##*/}" '{ print $2 " " vmlinuz }' \
```

rpkelly (approved):

cbgbt (approved):

PR #337, opened 2024-07-26
lock: update lock change resolution to fix mutating image collision

jmt-lab:
**Description of changes:**

* Changes lock change resolution to no longer only rely on the Twoliter.toml digest. Instead it will rerun the resolution and compare the result in order to verify that the remote images haven't mutated
* Adds --update flag to twoliter fetch to allow for performing an update at fetch time
* Removes release-version and root digest from Twoliter.toml
* Consolidates the constructors

**Testing done:**

Tested that Twoliter.lock collision still correctly occurs including remote digest change.

**Terms of contribution:**

### DIFF
--- a/twoliter/src/lock.rs
+++ b/twoliter/src/lock.rs
@@ -1,4 +1,4 @@
-use crate::common::fs::{create_dir_all, read, remove_dir_all, remove_file, write};
+use crate::common::fs::{create_dir_all, read, remove_dir_all, write};
 use crate::project::{Image, Project, ValidIdentifier, Vendor};
 use crate::schema_version::SchemaVersion;
 use anyhow::{ensure, Context, Result};
@@ -22,7 +22,7 @@ use tokio::fs::read_to_string;
 const TWOLITER_LOCK: &str = "Twoliter.lock";
 
 /// Represents a locked dependency on an image
-#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize)]
+#[derive(Debug, Clone, Eq, Ord, PartialOrd, Serialize, Deserialize)]
 pub(crate) struct LockedImage {
     /// The name of the dependency
     pub name: String,
@@ -36,6 +36,12 @@ pub(crate) struct LockedImage {
     pub digest: String,
     #[serde(skip)]
     pub(crate) manifest: Vec<u8>,
+}
+
+impl PartialEq for LockedImage {
+    fn eq(&self, other: &Self) -> bool {
+        self.source == other.source && self.digest == other.digest
+    }
 }
 
 impl LockedImage {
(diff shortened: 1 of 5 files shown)


twoliter/src/lock.rs
```diff
@@ -38,6 +38,12 @@ pub(crate) struct LockedImage {
     pub(crate) manifest: Vec<u8>,
 }
 
+impl PartialEq for LockedImage {
+    fn eq(&self, other: &Self) -> bool {
+        self.source == other.source && self.digest == other.digest
+    }
+}
```
cbgbt, line 45: I think the image digest should be sufficient for equivalency, no?

twoliter/src/lock.rs
```diff
@@ -252,49 +258,38 @@ impl OCIArchive {
...
 /// Represents the structure of a `Twoliter.lock` lock file.
-#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, Serialize, Deserialize)]
 #[serde(rename_all = "kebab-case")]
 pub(crate) struct Lock {
     /// The version of the Twoliter.toml this was generated from
     pub schema_version: SchemaVersion<1>,
-    /// The workspace release version
-    pub release_version: String,
     /// The resolved bottlerocket sdk
```
cbgbt, line 266: Does this require a `SchemaVersion` change?

twoliter/src/lock.rs
```diff
@@ -252,49 +258,38 @@ impl OCIArchive {
...
     /// Resolved kit dependencies
     pub kit: Vec<LockedImage>,
-    /// sha256 digest of the Project this was generated from
-    pub digest: String,
 }
 
 #[allow(dead_code)]
 impl Lock {
-    pub(crate) async fn load(project: &Project) -> Result<Self> {
+    pub(crate) async fn load(project: &Project, create: bool) -> Result<Self> {
```
cbgbt: I think I'm not a fan of the implicit `twoliter update` option, though there are two reasons that are somewhat disjoint:
* From a customer perspective, I like needing to separately see and accept lockfile changes.
* From an implementation perspective, if we keep this, I'd rather see `load` not take on modal functionality. We should prefer to keep it simple and handle the additional behavior in some wrapper.

twoliter/src/lock.rs
jmt-lab: The source could change with vendor but the digest could theoretically be the same

twoliter/src/lock.rs
```diff
@@ -252,49 +258,38 @@ impl OCIArchive {
...
-    }
-
-    pub(crate) async fn create(project: &Project) -> Result<Self> {
-        let lock_file_path = project.project_dir().join(TWOLITER_LOCK);
-        if lock_file_path.exists() {
-            remove_file(&lock_file_path).await?;
+            if lock_state == lock {
+                return Ok(lock);
+            }
+            ensure!(create, "changes have occured to Twoliter.toml or the remote kit images that require an update to Twoliter.lock.\n If intentional, please run twoliter update or pass the --update to twoliter fetch");
```
```suggestion
            ensure!(create, "changes have occured to Twoliter.toml or the remote kit images that require an update to Twoliter.lock.\n If intentional, please run twoliter update or pass the --update flag to twoliter fetch");
```

twoliter/src/lock.rs
```diff
@@ -252,49 +258,38 @@ impl OCIArchive {
...
-            ensure!(lock.digest == project.digest()?, "changes have occurred to Twoliter.toml that require an update to Twoliter.lock, if intentional please run twoliter update");
-            return Ok(lock);
-        }
-        Self::create(project).await
-    }
-
-    pub(crate) async fn create(project: &Project) -> Result<Self> {
-        let lock_file_path = project.project_dir().join(TWOLITER_LOCK);
-        if lock_file_path.exists() {
-            remove_file(&lock_file_path).await?;
```
sam-berning, line 290: do we need to remove the existing lockfile when `create == true`?

cbgbt (commented):
Ah, I just noticed I never clicked "submit" on this review. Sorry!

twoliter/src/lock.rs
jmt-lab: No not really the schema version here is more to track the schema version of Twoliter.toml

twoliter/src/lock.rs
jmt-lab: Yeah, let me update this

twoliter/src/lock.rs
cbgbt: I'm imagining that it could also be a problem if you update twoliter and the lockfile no longer matches, despite the contents being the same. I believe that shouldn't be the case here, though.

twoliter/src/lock.rs
```diff
@@ -252,48 +258,42 @@ impl OCIArchive {
...
+    }
+
+    pub(crate) async fn load(project: &Project) -> Result<Self> {
+        let lock_file_path = project.project_dir().join(TWOLITER_LOCK);
+        let lock_state = Self::resolve(project).await?;
+        ensure!(
+            lock_file_path.exists(),
+            "Twoliter.lock does not exist please run twoliter update first"
+        );
+        let lock_str = read_to_string(&lock_file_path)
```
cbgbt: nit: You could move the resolution below the `ensure` to save some work in the case that the workspace isn't created yet.

cbgbt (approved):

twoliter/src/lock.rs
```diff
@@ -252,48 +258,42 @@ impl OCIArchive {
...
             .context("failed to write lock file")?;
+        Ok(lock_state)
+    }
+
+    pub(crate) async fn load(project: &Project) -> Result<Self> {
+        let lock_file_path = project.project_dir().join(TWOLITER_LOCK);
+        let lock_state = Self::resolve(project).await?;
+        ensure!(
+            lock_file_path.exists(),
+            "Twoliter.lock does not exist please run twoliter update first"
```
sam-berning: nit:

```suggestion
            "Twoliter.lock does not exist, please run `twoliter update` first"
```

sam-berning (approved):

cbgbt (approved):

sam-berning (approved):

PR #338, opened 2024-07-28
improve logging in lock generation

cbgbt:
**Description of changes:**
I had a `twoliter update` fail and it was difficult to determine why, so I've added additional logging to lockfile generation. I've also done some minor refactoring of the Kit metadata extraction code that threw the error in question.

This (potentially controversially) adds a dependency on the `tracing` crate. Here's why I like this dependency:
* Makes it easy to emit `trace`-level logs for each function call using the `instrument` macro
* Makes it easy to add context to log messages with helpful variable values
* Integrates with the `log` crate for output, or optionally allows more complex IO via the `subscriber` concept.

Here's an example:

```
[2024-07-28T03:18:49Z INFO  twoliter::lock] Resolving project references to create lock file
[2024-07-28T03:18:50Z ERROR twoliter::lock] Mismatched kit metadata in manifest
list canonical_metadata=<ImageMetadata(decoded) [ImageMetadata { name: "bottlerocket-core-kit", version: Version { major: 2, minor: 0, patch: 0 }, sdk: Image {
name: ValidIdentifier("thar-be-beta-sdk"), version: Version { major: 0, minor: 43, patch: 0 }, vendor: ValidIdentifier("bottlerocket-new") }, kits: [] }]> kit_metadata=<ImageMetadata(decoded) [ImageMetadata { name: "bottlerocket-core-kit",
version: Version { major: 2, minor: 0, patch: 0 }, sdk: Image { name: ValidIdentifier("thar-be-beta-sdk"), version: Version { major: 0, minor: 43, patch: 0 }, vendor: ValidIdentifier("bottlerocket") }, kits: [] }]>
Error: Metadata does not match between images in manifest list
```

**Testing done:**
* `make build` succeeds
* `twoliter update` emits helpful logs on failure
* `twoliter update` succeeds within existing workspaces

**Terms of contribution:**

### DIFF
--- a/twoliter/src/lock.rs
+++ b/twoliter/src/lock.rs
@@ -1,8 +1,10 @@
 use crate::common::fs::{create_dir_all, read, remove_dir_all, write};
 use crate::project::{Image, Project, ValidIdentifier, Vendor};
 use crate::schema_version::SchemaVersion;
-use anyhow::{ensure, Context, Result};
+use anyhow::{bail, ensure, Context, Result};
 use base64::Engine;
+use futures::pin_mut;
+use futures::stream::{self, StreamExt, TryStreamExt};
 use oci_cli_wrapper::{DockerArchitecture, ImageTool};
 use olpc_cjson::CanonicalFormatter as CanonicalJsonFormatter;
 use semver::Version;
@@ -11,13 +13,14 @@ use serde::{Deserialize, Deserializer, Serialize};
 use sha2::Digest;
 use std::cmp::PartialEq;
 use std::collections::{HashMap, HashSet};
-use std::fmt::{Display, Formatter};
+use std::fmt::{Debug, Display, Formatter};
 use std::fs::File;
 use std::hash::{Hash, Hasher};
 use std::mem::take;
 use std::path::{Path, PathBuf};
 use tar::Archive as TarArchive;
 use tokio::fs::read_to_string;
+use tracing::{debug, error, info, instrument, trace};
 
 const TWOLITER_LOCK: &str = "Twoliter.lock";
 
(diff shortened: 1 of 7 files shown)


twoliter/src/lock.rs
```diff
@@ -464,38 +590,75 @@ impl Lock {
...
-                    encoded == metadata,
-                    "metadata does match between images in manifest list"
+        trace!(manifest_list = ?manifest_list, "Deserialized manifest list");
+
+        debug!("Extracting kit metadata from OCI image");
+        let embedded_kit_metadata = stream::iter(manifest_list.manifests)
+            .map(|manifest| async move {
+                let image_uri = format!("{}/{}@{}", vendor.registry, image.name, manifest.digest);
+                EncodedKitMetadata::try_from_image(&image_uri, image_tool).await
+            })
```
bcressey: Is the Docker image tool going to be robust for parallel invocations?

cbgbt: It seems to support it (based on my own testing and a lack of warnings found online), but since a pull can result in many parallel layer pulls, 8 is probably too generous of a limit.

I also don't know if the parallel pulls collapse common layers into a single request (I'm guessing not), but the images won't really have layers in common anyways, since they are different architectures.

My real goal here was to separate the kit metadata fetch from the check that the metadata was consistent across images as a clarity change. The parallelism was added as a bonus. I wouldn't be opposed to removing it. For now, I'll just lower it.

sam-berning (approved):

cbgbt:
^ Remove concurrent image procedures. Potential instability is probably not worth the potential performance improvement.

cbgbt:
^ Force push to rebase

bcressey (approved):

PR #344, opened 2024-08-07
add support for vendor override files

jmt-lab:
**Description of changes:**
* Introduces concept of vendor override files. Users can now define a vendor.toml or multiple toml files in a vendor.d/ directory that specify overrides for vendors.
* Twoliter will now utilized the vendor overrides and maintains its digest check on the actual images match

**Testing done:**
* Tested use of vendor.toml override and ensured twoliter fetch passed and fetched the correct image
* Tested error on if the digests are different
* Tested use of vendor.d/ folder

**Terms of contribution:**


twoliter/src/lock.rs
```diff
@@ -459,9 +505,48 @@ impl Lock {
...
+                let next_body = read_to_string(file).await?;
+                let next: VendorOverrideFile = toml::from_str(next_body.as_str())
+                    .context("failed to deserialize vendor override file")?;
+                for (key, value) in next.vendor.iter() {
+                    overrides.insert(key.clone(), value.clone());
+                }
+            }
+            return Ok(overrides);
+        }
+        let file_path = project.project_dir().join("vendor.toml");
```
bcressey: Convention is for our top-level TOML files to start with an upper-case letter, so `Vendor.toml`.

However, I'm a little put off by how generic that name is. Maybe we could support a `.twoliter` directory with various config files instead - e.g. `vendor.toml` or `override.toml` ?

jmt-lab: Hrm yeah, maybe for vendor it would be .twoliter/vendor.toml or .twoliter/vendor.d Sean had mentioned that we should support multiple vendor files in a directory in addition to single file

cbgbt:
Would it be possible to add integration tests that show how vendor overrides interact with a workspace? I think that would help clarify the behavior and, of course, protect us from regressions.

twoliter/src/lock_image.rs
```diff
@@ -0,0 +1,97 @@
...
+        image: &'lock LockedImage,
+        overrides: &'lock HashMap<String, Vendor>,
+    ) -> Self {
+        Self {
+            image,
+            vendor_overrides: overrides,
+        }
+    }
+}
+// 247242824936.dkr.ecr.us-west-2.amazonaws.com
```
cbgbt:
```suggestion

```

twoliter/src/lock_image.rs
```diff
@@ -0,0 +1,97 @@
+use crate::{lock::LockedImage, project::Vendor};
```
cbgbt: Do you mind documenting this module or the `UniqueLockedImage` struct?

twoliter/src/lock_image.rs
```diff
@@ -0,0 +1,97 @@
...
+        state.write(self.image.name.as_bytes());
+        state.write(self.image.vendor.as_bytes());
+        state.write(self.image.digest.as_bytes());
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use std::collections::HashMap;
+
```
cbgbt: nit: remove whitespace

```suggestion
```

twoliter/src/lock.rs
```diff
@@ -584,9 +626,48 @@ impl Lock {
             schema_version: project.schema_version(),
             sdk: LockedImage::new(&image_tool, vendor, sdk).await?,
             kit: locked,
+            vendor_override: Self::resolve_overrides(project).await?,
         })
     }
 
+    async fn resolve_overrides(project: &Project) -> Result<HashMap<String, Vendor>> {
```
cbgbt: Do you mind adding docstrings to these new functions?

Unit tests on override resolution would also be helpful.

twoliter/src/lock.rs
```diff
@@ -584,9 +626,48 @@ impl Lock {
...
+                }
+            }
+            // Sort the files for override ordering (WalkDir/read_dir does not gaurantee any ordering)
+            files.sort();
+            for file in files.iter() {
+                let next_body = read_to_string(file).await?;
+                let next: VendorOverrideFile = toml::from_str(next_body.as_str())
+                    .context("failed to deserialize vendor override file")?;
+                for (key, value) in next.vendor.iter() {
+                    overrides.insert(key.clone(), value.clone());
```
cbgbt: nit

```suggestion
                for (key, value) in next.vendor.into_iter() {
                    overrides.insert(key, value);
```

cbgbt (commented):
Nice work so far! There has been some offline conversation around modifying how the override interacts with different scenarios.

Apart from that, I have some suggestions for the implementation.

twoliter/src/lock/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+    /// The version of the Twoliter.toml this was generated from
+    pub schema_version: SchemaVersion<1>,
+    /// The resolved bottlerocket sdk
+    pub sdk: LockedImage,
+    /// Resolved kit dependencies
+    pub kit: Vec<LockedImage>,
+    /// Any overrides, these are read from a separate override file
+    #[serde(skip)]
+    pub overrides: BTreeMap<String, BTreeMap<String, Override>>,
+}
```
cbgbt: `#[serde(skip)]` here introduces an invariant that the programmer now has to keep track of: If I want to use overrides, I need to ensure that codepaths which populate this B Tree have been run first.

I'd prefer if we could instead use the type system enforce these invariants. Maybe we could instead create a parent `Workspace` structure which refers to both the ser/de `Lock` and also separately parses or holds the overrides?

twoliter/src/lock/archive.rs
```diff
@@ -0,0 +1,125 @@
+use super::views::{IndexView, ManifestLayoutView};
+use crate::common::fs::{create_dir_all, read, read_to_string, remove_dir_all, write};
+use anyhow::{Context, Result};
+use oci_cli_wrapper::ImageTool;
+use std::fs::File;
+use std::path::{Path, PathBuf};
+use tar::Archive as TarArchive;
+use tracing::{debug, instrument, trace};
+#[derive(Debug)]
```
cbgbt: I would have expected `cargo fmt` to add whitespace here.

twoliter/src/lock/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+
+impl PartialEq for Lock {
+    fn eq(&self, other: &Self) -> bool {
+        self.schema_version == other.schema_version
+            && self.sdk == other.sdk
+            && self.kit == other.kit
```
cbgbt, line 61: We may wish to do a set comparison here.

twoliter/src/lock/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+
+        info!("Resolving project references to check against lock file");
+        let lock_state = Self::resolve(project, &lock.overrides).await?;
+
+        ensure!(lock_state == lock, "changes have occured to Twoliter.toml or the remote kit images that require an update to Twoliter.lock");
+        Ok(lock)
+    }
+
+    #[instrument(level = "trace", skip(project))]
+    async fn load_overrides(
```
cbgbt: Looks like this drops the concept of the `override.d` style directory that we had mentioned.

Just commenting to confirm. I'm ok with incrementally adding that if we need it!

twoliter/src/lock/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+        );
+        for image in self.kit.iter() {
+            let vendor = project
+                .vendor()
+                .get(&ValidIdentifier(image.vendor.clone()))
+                .context(format!(
+                    "failed to find vendor for kit with name '{}' and vendor '{}'",
+                    image.name, image.vendor
+                ))?;
+            let override_ = self
```
cbgbt: I never knew that Rust reserved the `override` keyword :)

twoliter/src/lock/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+        if external_metadata_file.exists() {
+            let existing = read(&external_metadata_file).await.context(format!(
+                "failed to read external kit metadata: {}",
+                external_metadata_file.display()
+            ))?;
+            // If this is the same as what we generated skip the write
+            if existing == kit_list {
+                return Ok(());
+            }
+        }
```
cbgbt, line 163: Could we refactor this out into another function?

twoliter/src/lock/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+        let mut remaining: Vec<Image> = project.kits();
+        let mut sdk_set: HashSet<Image> = HashSet::new();
+        if let Some(sdk) = project.sdk_image() {
+            // We don't scan over the sdk images as they are not kit images and there is no kit metadata to fetch
+            sdk_set.insert(sdk.clone());
+        }
+        while !remaining.is_empty() {
+            let working_set: Vec<_> = take(&mut remaining);
+            for image in working_set.iter() {
+                debug!(%image, "Resolving kit '{}'", image.name);
```
cbgbt, line 189: Not needed for this PR since we aren't actually changing this, but I think it's more common in rust to use `VecDeque` for breadth-first traversal.

e.g. you can do:

```rust
while let Some(image) = remaining.pop_front() {
    /// etc etc
    for dep in metadata.kits {
        remaining.push_back();
    }
}
```

I'd be happy with just cutting an issue to simplify this later.

twoliter/src/lock/mod.rs
```diff
@@ -0,0 +1,289 @@
+pub mod archive;
+pub mod image;
+pub mod views;
+
```
cbgbt: Do you think you could add module-level docstrings explaining what functionality is present in each file under `lock`?

twoliter/src/lock/image.rs
```diff
@@ -0,0 +1,434 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_fmt(format_args!(
+            "{}-{}@{} ({})",
+            self.name, self.version, self.vendor, self.source,
+        ))
+    }
+}
+
+/// The hash should not contain the source to allow for collision detection
+impl Hash for LockedImage {
```
cbgbt: We chatted in person and I think this `Hash` implementation doesn't quite handle what the comment claims.

In hash tables, the `Hash` function is only used to determine which hash bucket the key is stored in. Since hash collisions can occur, each bucket stores elements in a list. The `Eq` implementation is used to determine whether or not the key is already present in that list.

[Here's a Rust playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f62b12c9ab5a3e963c222280a92a7ec0) which demonstrates the issue. Even though both `a` and `b` have the same `Hash` result, they are both present in the set due to the `Eq` implementation.

Finally, [the Rust docs](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html) also mention the following:

> It is required that the keys implement the [Eq](https://doc.rust-lang.org/stable/std/cmp/trait.Eq.html) and [Hash](https://doc.rust-lang.org/stable/std/hash/trait.Hash.html) traits, although this can frequently be achieved by using #[derive(PartialEq, Eq, Hash)]. If you implement these yourself, it is important that the following property holds:
>> k1 == k2 -> hash(k1) == hash(k2)

We should probably just `derive(PartialEq, Hash)` or else take a step back here and look for another approach for the problem we're trying to solve. It would be great if we could add tests to illuminate the hash collision problem.

twoliter/src/lock/image.rs
```diff
@@ -0,0 +1,434 @@
...
+        let manifest_list = self.get_manifest(image_tool).await?;
+        let registry = uri
+            .registry
+            .as_ref()
+            .context("no registry found for image")?;
+
+        let locked_image = LockedImage {
+            name: self.image_resolver_impl.name(),
+            version: self.image_resolver_impl.version()?,
+            vendor: vendor.to_string(),
```
cbgbt: Do you think it makes sense for the `image_resolver_impl` to also be able to tell us the vendor?

twoliter/src/lock/image.rs
```diff
@@ -0,0 +1,434 @@
...
+            digest: self.calculate_digest(image_tool).await?,
+        };
+
+        Ok((
+            locked_image,
+            if skip_metadata {
+                None
+            } else {
+                debug!("Extracting kit metadata from OCI image");
+                let embedded_kit_metadata =
```
cbgbt: The indent level here is a bit hard to follow. Do you think we could do an early return if `skip_metadata` is `true` to de-dent here?

cbgbt (commented):
Nice work! The refactor really makes things much cleaner, thanks for doing that! I think as overall feedback, I'd like to see some test cases showing the expected behavior of the overrides.

twoliter/src/lock/mod.rs
jmt-lab: I actually think this should move to Project which kind of acts as our parent Workspace structure.

twoliter/src/lock/image.rs
```diff
@@ -0,0 +1,434 @@
...
+                .clone()
+                .unwrap_or(self.base_uri.repo.clone()),
+            tag: self.base_uri.tag.clone(),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct ImageResolver {
+    image_resolver_impl: Box<dyn ImageResolverImpl>,
```
sam-berning, line 206: why not add vendor as a struct member here? it feels a bit odd that you pass vendor to initialize the resolver, and then pass it again when you actually resolve an image

twoliter/src/lock/mod.rs
cbgbt: That makes sense to me!

twoliter/src/lock/image.rs
```diff
@@ -0,0 +1,434 @@
...
+        self.version.hash(state);
+        self.vendor.hash(state);
+    }
+}
+
+#[derive(Deserialize, Debug, Clone)]
+pub(crate) struct ImageMetadata {
+    /// The name of the kit
+    #[allow(dead_code)]
+    pub name: String,
```
sam-berning, line 53: I think as long as name is part of the image metadata overrides with a different names won't behave as expected.

Say I have a kit dependency at `public.ecr.aws/bottlerocket/foo:v1.0.0` and I override it with `public.ecr.aws/bottlerocket/bar:v1.0.1`, the name in the image metadata (and thus the config labels) for the former will be `foo` and the latter will be `bar`. So when I push my changes for v1.0.1 to `foo`, regardless of whether the rest of the content is the same, the manifests will never have the same digests, unless I copy the image directly from `bar` to `foo`, in which case `foo` will have invalid metadata.

Can the name field be removed? If not, it might make sense to enforce that overrides must have the same name as the thing they are overriding.

sam-berning: I talked this over with @cbgbt and I don't think we need to block the PR on this. It's not an issue for the SDK since it doesn't have kit metadata, and it would require specific effort to hit this issue on kits, since we publish kits to a repository matching the name of the directory that the kit is defined in. So you'd have to rename `kits/foo` to `kits/bar` in your repository to even publish the kit to a repository that has a different name than the default

twoliter/src/lock/mod.rs
jmt-lab: I love when you leave an improvement comment on your own code :D

twoliter/src/lock/image.rs
```diff
@@ -0,0 +1,442 @@
...
+        f.write_fmt(format_args!(
+            "{}-{}@{} ({})",
+            self.name, self.version, self.vendor, self.source,
+        ))
+    }
+}
+
+/// The hash should not contain the source to allow for collision detection
+// impl Hash for LockedImage {
+//     fn hash<H: Hasher>(&self, state: &mut H) {
```
cbgbt: Remove?

twoliter/src/project.rs
```diff
@@ -56,6 +59,10 @@ pub(crate) struct Project {
 
     /// Set of kit dependencies
     kit: Vec<Image>,
+
+    /// Set of overrides read from Twoliter.override
+    #[serde(skip)]
+    overrides: BTreeMap<String, BTreeMap<String, Override>>,
```
cbgbt: Since we have `UnvalidatedProject`, should we just remove the `serde` tag from this and do away with `serde(skip)`?

twoliter/src/lock/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+                .overrides()
+                .get(&image.vendor)
+                .and_then(|x| x.get(&image.name));
+            let resolver =
+                ImageResolver::from_locked_image(image, image.vendor.as_str(), vendor, override_);
+            resolver
+                .extract(&image_tool, &project.external_kits_dir(), arch)
+                .await?;
+        }
+        let mut kit_list = Vec::new();
```
cbgbt, line 146: I'd still prefer if we refactor this to minimize the overall complexity of this function.

twoliter/src/lock/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+                .get(&image.vendor)
+                .and_then(|x| x.get(&image.name));
+            let resolver =
+                ImageResolver::from_locked_image(image, image.vendor.as_str(), vendor, override_);
+            resolver
+                .extract(&image_tool, &project.external_kits_dir(), arch)
+                .await?;
+        }
+
+        self.synchronize_metadata(project).await
```
cbgbt, line 142: Nice!

cbgbt (approved):
Nice work, @jmt-lab!

As a followup on the ask for integration testing -- @jmt-lab hit some road blocks with Docker misbehaving while interacting with a secure registry on localhost (and also seemingly making nonsense of the image digest...).

I have seen his integration tests pass for crane and otherwise seen how they fail due to docker "strangeness."

We'll cut an issue to followup on cleaning up the integration tests, since twoliter overrides blocks other important work in Twoliter.

@jmt-lab mind linking a remote branch with the current progress, even if the code is hack quality?

jmt-lab:
It's currently living in my fork at: https://github.com/jmt-lab/twoliter/tree/integ-save

sam-berning (approved):

PR #345, opened 2024-08-13
buildsys: add CI/CD hack, fix mtime for external files

bcressey:
**Issue number:**
N/A

**Description of changes:**
Two `buildsys` changes, one potentially controversial, one hopefully not.

The first is the `BUILDSYS_CICD_HACK` variable - naming suggestions wanted! - to allow `cargo` to recreate crates without actually running expensive package builds. It's a hack because it assumes the `build` directory is fully populated; if it's not, then bad things will happen the next time a package is rebuilt.

The second fixes an edge case related to external files or bundled Go modules; these can end up being tracked by `cargo` for changes, so we need to ensure that their modification time is no later than the start of the build. `Cargo.toml` is already tracked for changes and is a useful sentinel for this purpose.

**Testing done:**
Tested in https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/67. With both changes, I was able to reuse cached package builds, touch the files modified by a pull request, and observe that only the affected packages were rebuilt.

**Terms of contribution:**

### DIFF
--- a/tools/buildsys/src/args.rs
+++ b/tools/buildsys/src/args.rs
@@ -92,6 +92,14 @@ pub(crate) struct Common {
 
     #[arg(long, env = "TWOLITER_TOOLS_DIR")]
     pub(crate) tools_dir: PathBuf,
+
+    /// cicd_hack is used to suppress builds from running after all the cargo-related metadata is
+    /// emitted. This allows cargo to create a fresh crate, and assumes that the corresponding
+    /// build artifacts are already present. It is intended for use in a CI/CD scenario where some
+    /// other process populates the build directory from a cache. Other uses may lead to unexpected
+    /// build failures that are difficult to troubleshoot.
+    #[arg(long, env = "BUILDSYS_CICD_HACK")]
+    pub(crate) cicd_hack: bool,
 }
 
 /// Build RPMs from a spec file and sources.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -760,6 +760,7 @@ dependencies = [
  "buildsys-config",
  "clap",
  "duct",
+ "filetime",
  "guppy",
  "hex",
  "lazy_static",
--- a/tools/buildsys/Cargo.toml
+++ b/tools/buildsys/Cargo.toml
@@ -13,6 +13,7 @@ bottlerocket-variant.workspace = true
 buildsys-config.workspace = true
 clap = { workspace = true, features = ["derive", "env"] }
 duct.workspace = true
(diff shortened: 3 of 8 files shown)


tools/buildsys/src/args.rs
```diff
@@ -92,6 +92,9 @@ pub(crate) struct Common {
 
     #[arg(long, env = "TWOLITER_TOOLS_DIR")]
     pub(crate) tools_dir: PathBuf,
+
+    #[arg(long, env = "BUILDSYS_CICD_HACK")]
+    pub(crate) cicd_hack: Option<String>,
```
cbgbt: I think I'm OK with this variable name so long as we document its purpose in the code as well.

That said, it is a great hack 😄

cbgbt (approved):
🚀

mgsharm (approved):

bcressey:
First force push handles the rebase.

Second force push adds some documentation, and adjusts the argument parsing to take a bool instead of an string value that's ignored.

```
Usage: buildsys build-package ...
...
      --cicd-hack
          cicd_hack is used to suppress builds from running after all the cargo-related metadata is emitted. This allows cargo to create a fresh crate, and assumes that the corresponding build artifacts are already present. It is intended for use in a CI/CD scenario where some other process populates the build directory from a cache. Other uses may lead to unexpected build failures that are difficult to troubleshoot [env: BUILDSYS_CICD_HACK=]
```

Testing:
```
❯ cargo run --bin buildsys -- build-package ...
[tools/buildsys/src/main.rs:131:5] args = BuildPackageArgs {
...
    common: Common {
        ...,
        cicd_hack: false,
    },
}

❯ cargo run --bin buildsys -- build-package --cicd-hack ...
[tools/buildsys/src/main.rs:131:5] args = BuildPackageArgs {
...
    common: Common {
        ...,
        cicd_hack: true,
    },
}

❯ BUILDSYS_CICD_HACK=true cargo run --bin buildsys -- build-package ...
[tools/buildsys/src/main.rs:131:5] args = BuildPackageArgs {
...
    common: Common {
        ...,
        cicd_hack: true,
    },
}

❯ BUILDSYS_CICD_HACK=false cargo run --bin buildsys -- build-package ...
[tools/buildsys/src/main.rs:131:5] args = BuildPackageArgs {
...
    common: Common {
        ...,
        cicd_hack: false,
    },
}
```

cbgbt (approved):

sam-berning (approved):

PR #348, opened 2024-08-19
pubsys: improve stability of SSM parameter validation

### LINKED ISSUE #346: pubsys: Improve error message on SSM validation failure
If SSM validation fails after a pubsys promotion, there error message is:

```
Failed to promote SSM: ValidateSsm
```

This occurs because [there's no `display` implementation specifically for the `ValidateSsm` error variant](https://github.com/bottlerocket-os/twoliter/blob/efd17d998a6b59cb1d26866d155f8c4db3f6bcc2/tools/pubsys/src/aws/promote_ssm/mod.rs#L344). In snafu 0.8, the [default Display implementation was removed](https://github.com/shepmaster/snafu/blob/main/src/guide/upgrading.md#version-07--08), so this variant no longer prints the error stack that caused it to fail.

Adding a `#[snafu(display` line to that variant should resolve it.

### LINKED ISSUE #347: pubsys: SSM validation is flaky
SSM validation after SSM variable changes often fails. The error messages aren't very helpful (#346); however, we believe it is either due to throttling or to the validation occuring too quickly and encountering eventual consistency struggles.

Here are some tips on resolving the bug:
* We can use the [governor crate](https://crates.io/crates/governor) for a client-side rate-limiting implementation.
* We get SSM parameters in two spots, both should probably be rate-limited using governor:
  * [Here](https://github.com/bottlerocket-os/twoliter/blob/efd17d998a6b59cb1d26866d155f8c4db3f6bcc2/tools/pubsys/src/aws/ssm/ssm.rs#L40-L43) we call `get_parameters()` in chunks of 10
  * [Here ](https://github.com/bottlerocket-os/twoliter/blob/efd17d998a6b59cb1d26866d155f8c4db3f6bcc2/tools/pubsys/src/aws/ssm/ssm.rs#L173-L181)we use a paginator and request new pages in a loop
* The [published rate limits are here](https://docs.aws.amazon.com/general/latest/gr/ssm.html). Seems like 40 TPS is what we should aim for -- I think it would be wise to use lazy_static! to make a process-wide shared rate limiter.
* Retries: AWS SDK does some retrying to overcome transient errors, and possibly also 500 server errors. There's a hunch that our reliability problems could be caused by eventual consistency -- we call validate immediately after set, so the set operation may not actually be consistent with the get view. If this is the case, we're possibly getting 403 responses from the API.
  * We should [use `tokio_retry` for this](https://docs.rs/tokio-retry/latest/tokio_retry/). We could add this at the top-level [here ](https://github.com/bottlerocket-os/twoliter/blob/efd17d998a6b59cb1d26866d155f8c4db3f6bcc2/tools/pubsys/src/aws/ssm/mod.rs#L220-L223) and retry the entire validate operation on any failure. it's important to use exponential backoff here.

cbgbt:
**Issue number:**

Closes #347
Closes #346

**Description of changes:**
```
    workspace: homogenize dependencies
    
    This changes helps prepare the workspace to use workspace dependencies
    by using the same version requirements and feature sets for common
    dependencies of crates in the workspace.
```

```
    workspace: move to workspace dependencies
```

```
    pubsys: implement Display for ValidateSsm error
```

```
    pubsys: rate-limit SSM GetParameter calls
```

```
    pubsys: retry SSM validation on any failure
```

**Testing done:**
* [x] published and validated AMIs with SSM parameters

**Terms of contribution:**


Cargo.toml
```diff
@@ -36,6 +36,95 @@ installers = []
...
 targets = ["x86_64-unknown-linux-musl", "aarch64-unknown-linux-musl"]
 
+[workspace.dependencies]
+bottlerocket-types = { version = "0.0.14", git = "https://github.com/bottlerocket-os/bottlerocket-test-system", tag = "v0.0.14" }
+bottlerocket-variant = { version = "0.1", path = "tools/bottlerocket-variant" }
+buildsys = { version = "0.1", path = "tools/buildsys", lib = true, artifact = [ "bin:buildsys", "bin:bottlerocket-variant" ] }
+buildsys-config = { version = "0.1", path = "tools/buildsys-config" }
+oci-cli-wrapper = { version = "0.1", path = "tools/oci-cli-wrapper" }
+parse-datetime = { version = "0.1", path = "tools/parse-datetime" }
+pipesys = { version = "0.1", path = "tools/pipesys", lib = true, artifact = [ "bin:pipesys" ] }
```
bcressey, line 46: What does `lib = true` do here? It's sort of surprising since these are all binaries.

Can we move the artifact dependency stuff to `twoliter/Cargo.toml` and leave just the version and path here? It seems a little weird for cases where we have a cross crate dependency (like `pubsys` depends on `buildsys`).

tools/pubsys/src/aws/ssm/ssm.rs
```diff
@@ -1,23 +1,74 @@
...
+    static ref GET_PARAMETERS_MAX_JITTER: Jitter = Jitter::up_to(Duration::from_millis(20));
+    static ref GET_PARAMETERS_RATE_LIMITER: RegionKeyRateLimiter = RateLimiter::keyed(
+        Quota::per_second(nonzero!(GET_PARAMETERS_RATE_LIMIT_PER_SEC))
+            .allow_burst(nonzero!(GET_PARAMETERS_BURST_LIMIT))
+    );
+}
+
+/// Async throttling function to be called before calling SSM GetParameter* functions.
+///
+/// Returns when the rate limit is satisifed.
```
bcressey: nit:
```suggestion
/// Returns when the rate limit is satisfied.
```

bcressey (approved):

Cargo.toml
cbgbt: > What does lib = true do here? It's sort of surprising since these are all binaries.

`buildsys` depends on `pipesys` as a library, and you need to add `lib = true` if you use artifact dependencies to tell Cargo to build both.
```
   Compiling buildsys v0.1.0 (/home/ec2-user/twoliter/tools/buildsys)
error[E0433]: failed to resolve: use of undeclared crate or module `pipesys`
  --> tools/buildsys/src/builder.rs:20:5
   |
20 | use pipesys::server::Server as PipesysServer;
   |     ^^^^^^^ use of undeclared crate or module `pipesys`
```

---

> Can we move the artifact dependency stuff to `twoliter/Cargo.toml` and leave just the version and path here?

I really wanted to do it this way, but Cargo wouldn't cooperate. e.g. if you move the `artifact` declaration for `pubsys` to `twoliter/Cargo.toml`, `cargo build` says:

```
cargo clippy --locked -- -D warnings --no-deps
warning: /home/ec2-user/twoliter/twoliter/Cargo.toml: unused manifest key: dependencies.pubsys.artifact
warning: twoliter v0.4.4 (/home/ec2-user/twoliter/twoliter) ignoring invalid dependency `pubsys` which is missing a lib target
```

... and then eventually

```
    Checking pubsys v0.1.0 (/home/ec2-user/twoliter/tools/pubsys)
error: environment variable `CARGO_BIN_FILE_PUBSYS` not defined at compile time
  --> twoliter/src/tools.rs:17:38
   |
17 | const PUBSYS: &[u8] = include_bytes!(env!("CARGO_BIN_FILE_PUBSYS"));
   |                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: Cargo sets build script variables at run time. Use `std::env::var("CARGO_BIN_FILE_PUBSYS")` instead
   = note: this error originates in the macro `env` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `twoliter` (bin "twoliter") due to 1 previous error
make: *** [Makefile:14: clippy] Error 101
```

jpculp (approved):

PR #358, opened 2024-08-30
release: update release.yml to use env variables instead of context v…

sumukhballal:
**Description of changes:**

Updates the current release flow to use env variables instead of GitHub Actions context values directly.

**Terms of contribution:**

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,6 +55,7 @@ jobs:
       has-releases: ${{ steps.create-release.outputs.has-releases }}
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      GITHUB_REF_NAME: ${{ github.ref_name }}
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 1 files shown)


mgsharm (approved):

.github/workflows/release.yml
```diff
@@ -133,15 +137,15 @@ jobs:
         run: |
           export PATH=$CARGO_HOME/bin:$PATH
           # Actually do builds and make zips and whatnot
-          cargo dist build --tag=${{ github.ref_name }} --output-format=json ${{ matrix.dist-args }} > dist-manifest.json
+          cargo dist build --tag="${GITHUB_REF_NAME}" --output-format=json "${MATRIX_DIST_ARGS}" > dist-manifest.json
```
cbgbt: I believe we want `MATRIX_DIST_ARGS` to remain exploded. This will have the shell treat `MATRIX_DIST_ARGS` as a single string containing whitespace, but the data inside is actually several distinct argument strings.

sumukhballal:
Addressed @cbgbt 's comments.

cbgbt (approved):

PR #357, opened 2024-08-30
add support for a "build-all" task

bcressey:
**Issue number:**
N/A

**Description of changes:**
This is a follow-up to https://github.com/bottlerocket-os/twoliter/pull/282 that removes the last of `buildsys`'s dependencies on `BUILDSYS_VARIANT` and friends. That in turn makes it possible to add a "build-all" target to `Makefile.toml`, which will help speed up CI runs.

(Note that there's still no "publish-all" target; `pubsys` continues to rely on `BUILDSYS_VARIANT` to know what to publish.)

**Testing done:**
For x86_64:
```
❯ cargo make build-all
...
[cargo-make][1] INFO - Running Task: build-all
   Compiling settings-plugins v0.1.0 (/home/fedora/brdev/packages/settings-plugins)
   Compiling settings-defaults v0.1.0 (/home/fedora/brdev/packages/settings-defaults)
   Compiling settings-migrations v0.1.0 (/home/fedora/brdev/packages/settings-migrations)
   Compiling common v0.0.0 (/home/fedora/brdev/packages/common)
   Compiling metal-k8s-1_28 v0.1.0 (/home/fedora/brdev/variants/metal-k8s-1.28)
   Compiling aws-k8s-1_27-nvidia v0.1.0 (/home/fedora/brdev/variants/aws-k8s-1.27-nvidia)
   Compiling metal-k8s-1_29 v0.1.0 (/home/fedora/brdev/variants/metal-k8s-1.29)
   Compiling vmware-dev v0.1.0 (/home/fedora/brdev/variants/vmware-dev)
   Compiling aws-k8s-1_28 v0.1.0 (/home/fedora/brdev/variants/aws-k8s-1.28)
   Compiling metal-dev v0.1.0 (/home/fedora/brdev/variants/metal-dev)
   Compiling vmware-k8s-1_27 v0.1.0 (/home/fedora/brdev/variants/vmware-k8s-1.27)
   Compiling vmware-k8s-1_28 v0.1.0 (/home/fedora/brdev/variants/vmware-k8s-1.28)
   Compiling aws-k8s-1_24-nvidia v0.1.0 (/home/fedora/brdev/variants/aws-k8s-1.24-nvidia)
   Compiling aws-k8s-1_27 v0.1.0 (/home/fedora/brdev/variants/aws-k8s-1.27)
   Compiling aws-k8s-1_28-nvidia v0.1.0 (/home/fedora/brdev/variants/aws-k8s-1.28-nvidia)
[...]

PR #361, opened 2024-09-06
twoliter: allow partial lockfile validation in some scenarios

cbgbt:
**Description of changes:**
```
    twoliter: allow partial lockfile validation in some scenarios
    
    In some CI scenarios, it's useful to drop previously-built variant
    images into a Twoliter project's build directory and use Twoliter to
    publish those variant images.
    
    In these cases, Kit dependencies are not necessary, and it can be useful
    to avoid resolving them and comparing them against Twoliter.lock.
    
    This change allows skipping Kit lockfile verification only when
    executing certain `twoliter make` targets.
```

**Testing done:**
* All provided tests pass

**Terms of contribution:**


cbgbt:
Sorry for the churn. I have another TL interface change I'd like to make. I can save it for a subsequent PR, though.

twoliter/embedded/Makefile.toml
```diff
@@ -1398,7 +1410,6 @@ pubsys \
 # Rather than depend on "build", which currently rebuilds images each run, we
 # depend on publish-tools and check for the input file below to save time.
 # This does mean that `cargo make ami` must be run before `cargo make validate-ami`.
-dependencies = ["fetch-sources"]
```
bcressey, line 1408: Seems like we can get rid of most of these 'Rather than depend on "build"' comments now, since these tasks don't depend on "publish-tools" or anything else.

cbgbt:
I disabled some additional, unnecessary artifact fetching. Some make targets needlessly requested workspace sources as an artifact of previously being part of the monorepo.

I need to test that specific change a bit more before merging, though.

twoliter/src/cmd/make.rs
```diff
@@ -1,11 +1,21 @@
...
 
+// Some subcommands do not require kits and thus do not need to resolve and verify them against the
+// lockfile.
+//
+// Avoiding that resolution can be useful in CI situations where images are already built and we
+// need to perform additional operations using the SDK.
+//
+// `twoliter make` targets in the following list will skip kit validation IFF they also explicitly
+// list their SDK dependency in Twoliter.toml.
+const SKIP_KIT_VALIDATION_TARGETS: &[&str] = &["repack-variant", "repo", "fetch-sdk"];
```
bcressey: Seems like we would want to list all of the repo, AMI, ssm, and OVA operations in this list:
* repo
* validate-repo
* check-repo-expirations
* refresh-repo
* ami
* ami-public
* ami-private
* grant-ami
* revoke-ami
* validate-ami
* ssm
* promote-ssm
* validate-ssm

Optionally:
* upload-ova
* vmware-template

Unsure:
* all the test related tasks
* all the clean related tasks

I know allowlists are preferred over denylists, but we only really need kit validation to occur in the places where we're building something - `build-package`, `build-variant`, and `build-all`. So it might be better to just enforce it there?

cbgbt:
Commenting so that GitHub will provide a useful comparison view.

cbgbt:
^ addresses feedback from @bcressey

sumukhballal:
LGTM! 👍

sumukhballal (approved):

cbgbt:
Force push to remove erroneous `println!`

bcressey (approved):

jmt-lab (approved):

cbgbt:
`build` and `default` are effectively aliases to `build-variant` and are used in the `bottlerocket` repo. These targets now require kit verification.

cbgbt:
^ rebase (without changes) on to develop

cbgbt:
^ make changes to work with #357

cbgbt:
^ force push because I missed `build-all` 😅

bcressey (approved):

PR #363, opened 2024-09-07
twoliter: refactor lock & project interfaces

cbgbt:
**Description of changes:**
While working on #361 I started to consider how to better close-off some of these verification abstractions. Separately, I had an idea on how to make it so that `ImageResolver` could be less aware of how overrides are stored.

This PR contains two minor refactors to existing APIs to try and more-tightly control usage of those APIs via privacy and type constraints.

```
    twoliter: refactor lock to be a private project submodule
    
    The lock abstractions force the project interface to perform the
    appropriate validations when resolving and using dependencies.
    This change moves the `lock` module under `project` so that we can
    more-tightly control access to that functionality.
```

---

```
    twoliter: force vendor lookups to deal with overrides
    
    This funnels vendor lookups into receiving an enum, which forces
    callers handling an image's vendor to explicitly handle overrides.
```

---

```
    twoliter: consistently provide image override view
    
    Twoliter had a bug in which overridden SDK images were resolved, but then
    the wrong SDK was being passed into builds.
    
    This change:
    * Uses the the sealed trait/typestate pattern to modify the Project with
      interfaces to fetch images based on the lock level.
    * Provides clarified public interfaces for getting images from pre/post
      lock resolution.
    * Forces public and private callers to deal with possibly-overridden
      images.
```

**Testing done:**
* Unit tests pass
* built `bottlerocket-core-kit`
* built `bottlerocket` variants

**Terms of contribution:**


bcressey (approved):

jmt-lab (approved):
LGTM

twoliter/src/project/mod.rs
```diff
@@ -215,12 +270,101 @@ impl Project {
...
+            .map(|kit| self.as_project_image(kit))
+            .collect::<Result<_>>()
+            .expect("Could not find kit vendor despite lock resolution succeeding?")
+    }
+
+    pub(crate) fn sdk_image(&self) -> ProjectImage {
+        let Locked(lock) = &self.lock;
+        self.as_project_image(&lock.sdk)
+            .expect("Could not find kit vendor despite lock resolution succeeding?")
+    }
```
bcressey: Is this the right error message?
```suggestion
    pub(crate) fn sdk_image(&self) -> ProjectImage {
        let Locked(lock) = &self.lock;
        self.as_project_image(&lock.sdk)
            .expect("Could not find SDK vendor despite lock resolution succeeding?")
    }
```

bcressey (approved):

cbgbt:
^ force push for feedback from @bcressey

bcressey (approved):

PR #376, opened 2024-09-17
makefile: remove dependencies from build-variant task

rpkelly:
**Issue number:**

Closes #

**Description of changes:**
Due to the changes in build requirements created by the switch to building from kits, only the sdk and the kits should now be required to build a botlerocket variant

**Testing done:**
Used resulting twoliter binary to build a bottlerocket variant from the bottlerocket repo.

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/Makefile.toml
+++ b/twoliter/embedded/Makefile.toml
@@ -857,7 +857,7 @@ cargo build \
 ]
 
 [tasks.build-variant]
-dependencies = ["fetch", "build-sbkeys", "publish-setup", "validate-kits"]
+dependencies = ["fetch-sdk", "build-sbkeys", "publish-setup", "validate-kits"]
 script = [
 '''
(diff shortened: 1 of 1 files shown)


twoliter/embedded/Makefile.toml
```diff
@@ -857,7 +857,7 @@ cargo build \
 ]
 
 [tasks.build-variant]
-dependencies = ["fetch", "build-sbkeys", "publish-setup", "validate-kits"]
+dependencies = ["fetch-sdk", "build-sbkeys", "publish-setup", "validate-kits"]
```
bcressey, line 860: `build-variant` can also trigger the equivalent of `build-package` tasks, and for those we do need to run `fetch-sources` (for first-party sources under `sources/`) and `fetch-vendored` (for packages that need bundled Go modules prepared).

PR #379, opened 2024-09-23
add support for erofs as root filesystem format

bcressey:
**Issue number:**
N/A

**Description of changes:**
Add the `erofs-root-partition` image feature. This is still considered experimental, but using erofs for the root filesystem has shown significant improvements in overall boot time. I'd like to land this implementation, and the associated changes in the core kit, to facilitate further experiments.

I tried many different settings for `mkfs.erofs` and eventually settled on the ones here. The physical cluster size made the largest difference. I tested various extended options - `dedupe`, `fragments`, `all-fragments`, and `ztailpacking` - but left those out to simplify the initial implementation, and to avoid features that the 6.1 kernel labels as [experimental](https://elixir.bootlin.com/linux/v6.1.111/source/fs/erofs/super.c#L359). 

**Testing done:**
Built `aws-dev` and `aws-k8s-1.30` images with the new feature flag enabled. Confirmed that `repack-variant` also worked.

Using some tools provided by @cbgbt, I compared the boot times for `aws-k8s-1.30` AMIs as of the v1.22.0 release, with the current ext4 root filesystem and the experimental erofs root filesystem.
erofs-boottime

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/img2img
+++ b/twoliter/embedded/img2img
@@ -6,6 +6,7 @@ shopt -qs failglob
 OUTPUT_FMT="raw"
 OVF_TEMPLATE=""
 
+EROFS_ROOT_PARTITION="no"
 UEFI_SECURE_BOOT="no"
 IN_PLACE_UPDATES="no"
 
@@ -21,6 +22,7 @@ for opt in "$@"; do
   --data-image-publish-size-gib=*) DATA_IMAGE_PUBLISH_SIZE_GIB="${optarg}" ;;
   --partition-plan=*) PARTITION_PLAN="${optarg}" ;;
   --ovf-template=*) OVF_TEMPLATE="${optarg}" ;;
+  --with-erofs-root-partition=*) EROFS_ROOT_PARTITION="${optarg}" ;;
   --with-uefi-secure-boot=*) UEFI_SECURE_BOOT="${optarg}" ;;
   --with-in-place-updates=*) IN_PLACE_UPDATES="${optarg}" ;;
   *)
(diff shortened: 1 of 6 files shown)


twoliter/embedded/img2img
```diff
@@ -149,25 +162,29 @@ install_root_json "${ROOT_MOUNT}"
...
+      cat <<EOF | debugfs -w -f - "${ROOT_IMAGE}" 2>>"${ROOT_DEBUGFS_STDERR}"
 rm ${artifact#"${ROOT_MOUNT}"}
 write ${artifact} ${artifact#"${ROOT_MOUNT}"}
 ea_set ${artifact#"${ROOT_MOUNT}"} security.selinux system_u:object_r:os_t:s0
 EOF
-  done
-  check_debugfs_errors "${ROOT_DEBUGFS_STDERR}"
+    done
+    check_debugfs_errors "${ROOT_DEBUGFS_STDERR}"
+  fi
```
jpculp, line 187: Since we're just completely remaking the root filesystem if it's erofs, should we follow the same workflow for ext4?

jpculp (approved):

twoliter/embedded/img2img
bcressey: For erofs, extracting the filesystem is a supported feature of `fsck.erofs`. For ext4, I'm more confident that `debugfs` will do what we want, since I'm not aware of a way to do full filesystem extraction using tools from e2fsprogs.

bcressey: The [erofs paper](https://www.usenix.org/system/files/atc19-gao.pdf) mentions support for "image patching" in 4.3, but I didn't find a way to do that with `erofs-utils`.

arnaldo2792 (approved):

arnaldo2792:
One question tho, what was the final size of the root filesystem?

cbgbt (approved):

yeazelm (approved):
Very cool!

bcressey:
> One question tho, what was the final size of the root filesystem?

For `aws-k8s-1.30`, it was 337 MiB with erofs and 700 MiB without. With some of the experimental options like `dedupe`, and a 1 MiB physical cluster size, I got the size below 300 MiB, though the image took quite a lot longer to build.

PR #384, opened 2024-10-02
rpm2img: use package version, release, and epoch

ginglis13:
When building application inventory, use a package's actual version and use the <timestamp>.<commit>.br1 for all packages, not just first party packages. Additionally, expose a package's Epoch - set to 0 if "(none)".

<!--
Tips:
- Please read CONTRIBUTING.md to understand our process and our requests for PRs.
- Please file an issue before creating a PR so we can discuss the change and confirm it's not already being worked on.
-->

**Issue number:**

Closes #364 

**Description of changes:**

When building application inventory, use a package's actual version
and use the `<timestamp>.<commit>.br1` for all packages, not just
first party packages. Additionally, expose a package's Epoch -
set to 0 if "(none)".

⚠️ **This change is dependent on changes to https://github.com/bottlerocket-os/bottlerocket-core-kit to add an `Epoch: 1` to all package spec defns for which the package version is < any core kit release versions.**


**Testing done:**

1. `make test`

2. Build a Bottlerocket k8s-1.29 variant with these changes; in the local Bottlerocket repo, I modified the first party packages to have an `Epoch: 1` to ensure this is passed properly.

```
cargo make \                                                                                                                                                                                               
          -e BUILDSYS_ARCH=aarch64 \
          -e=TWOLITER_REPO=https://github.com/ginglis13/twoliter \
          -e=TWOLITER_VERSION=d1f98f6dd873144be9a6d076662fd3de4a943e29 \
          -e=TWOLITER_ALLOW_SOURCE_INSTALL=true \
          -e=TWOLITER_ALLOW_BINARY_INSTALL=false \
          -e=TWOLITER_SKIP_VERSION_CHECK=true \
          -e BUILDSYS_VARIANT=aws-k8s-1.29
```

results in an application inventory file below:
<details>

```
{
  "Content": [
    {
      "Name": "acpid",
      "Publisher": "Bottlerocket",
      "Version": "2.0.34",
      "Release": "1.1727554942.ab18bc69.br1",
      "Epoch": "0",
      "InstalledTime": "2024-10-02T20:45:29Z",
      "ApplicationType": "Unspecified",
      "Architecture": "aarch64",
      "Url": "http://sourceforge.net/projects/acpid2/",
      "Summary": "ACPI event daemon"
    },
    {
      "Name": "apiclient",
      "Publisher": "Bottlerocket",
      "Version": "0.0",
      "Release": "0.1727554942.ab18bc69.br1",
      "Epoch": "0",
[...]

PR #387, opened 2024-10-07
Bump the Kit metadata label version, bundle krane with Twoliter

### LINKED ISSUE #352: Bundle `crane` with `twoliter`
Twoliter can use `docker` or `crane` in order to interact with Kit OCI repositories. To use `docker`, the docker daemon must have the `containerd-snapshotter` enabled. This is not always possible in environments where the user does not have access to the Docker daemon configuration.

Given that `crane` also has more desirable performance for kit publication, we should consider bundling crane with twoliter.

### LINKED ISSUE #362: CI: Add integration tests for `Twoliter.override`
Twoliter overrides allow you to fetch Twoliter artifacts from an alternative OCI repository during the build, son long as the fetched contents match the digest in Twoliter.lock.

@jmt-lab did some great initial work writing some integration tests for the feature, but they were hung up on unfortunate docker behavior around locally hosted container registries.

The existing code lives in https://github.com/jmt-lab/twoliter/tree/integ-save

This issue is to track merging those tests, or some subset of them (even if they only use `crane`) into Twoliter.

cbgbt:
**Issue number:**

Closes #362 
Closes #352

**Description of changes:**
The Kit metadata version bump is motivated by #384, which will begin to introduce application inventory that includes an RPM packages "Epoch" value and refrain from special-casing `bottlerocket-core-kit`. This change will cause unexpected advisory evaluation behavior for downstream workspaces which depend on core-kit but use an older version of Twoliter when core-kit bumps the epoch of all packages.

When introducing the version bump, Twoliter's integration tests began to fail because they depend on existing published kits, so this PR additionally:
* Bundles the `krane` tool with Twoliter, obviating the need to use system-provided OCI tooling
* Rewrites the integration tests to build and publish kits to a local registry

**Testing done:**
* Provided unit and integration tests pass
* If I try to build `bottlerocket-os/bottlerocket` with this change directly, I get the following message:

```
...
[2024-10-07T06:43:37Z INFO  twoliter::project::lock::image] Resolving dependency image dependency 'bottlerocket-core-kit-2.8.4@public.ecr.aws/bottlerocket/bottlerocket-core-kit:v2.8.4'.
Error: kit appears to be built with metadata version 'v1', possibly by an older version of twoliter with unsupported incompatibilities. This version of twoliter supports metadata version 'v2'.
```

* If I build my own core-kit, then proceed to build bottlerocket, the build proceeds as usual.
* If I try to build with this new core-kit, but the previous Twoliter, I get the following:

```
[2024-10-07T07:21:39Z INFO  twoliter::project::lock::image] Resolving dependency image dependency 'bottlerocket-core-kit-2.8.5@public.ecr.aws/bottlerocket/bottlerocket-core-kit:v2.8.5 (overridden-to: public.ecr.aws/seankell/bottlerocket-core-kit:v2.8.5)'.
Error: no metadata stored on image, this image appears to not be a kit
```

**Terms of contribution:**


tools/krane/hashes/crane
```diff
@@ -0,0 +1 @@
+SHA512 (go-containerregistry-v0.20.1.tar.gz) = c323c5b78c35fb7af67641fa4ef1802b944f8bd908163ff40a952a0c190e2dd210100efba1fbc2064495cff28a60d5bc7ee98e510ec116522c7897539b02fad8
```
bcressey, line 2: It'd be nice to have a URL in a comment above this, so it's easy to double-check hashes. Also in this case, there's both the GitHub-generated archive and a release artifact of the sources, so the URL would help indicate which one is used.

bcressey (approved):

cbgbt:
^ force push to address comment from @bcressey

cbgbt:
^ force push to rebase

bcressey (approved):

jmt-lab (approved):

sam-berning (approved):
Nice, this looks good!

Now that krane is bundled in to twoliter, there's some cleanup that we can do on `oci-cli-wrapper` (or maybe we can just eliminate it) because we needed to provide an interface that worked for both docker and krane and they didn't play nice in some cases. Not a problem for this PR though, we can just open an issue.

cbgbt:
> Now that krane is bundled in to twoliter, there's some cleanup that we can do on oci-cli-wrapper (or maybe we can just eliminate it) because we needed to provide an interface that worked for both docker and krane and they didn't play nice in some cases

I was thinking about this too. For now I've opted to keep the interface in case we regret the decision to use krane or want to shop around, or re-introduce the ability to override it at runtime.

I'm definitely open to removing it or favoring a more pleasant API in the future though!

PR #405, opened 2024-11-08
krane: write krane binary to a tempfile

### LINKED ISSUE #403: Twoliter's build of krane appears to be linking against libc
Twoliter fails to interact with OCI images on hosts with a less-recent libc than our cross-build environment (defined to some degree [here](https://github.com/bottlerocket-os/twoliter/blob/dbba40736e1a42c4f2715107f5baba45dfc8baab/Cargo.toml#L39). The build of `krane` occurs [here](https://github.com/bottlerocket-os/twoliter/blob/dbba40736e1a42c4f2715107f5baba45dfc8baab/tools/krane/build.rs#L37)).

When the error occurs, it looks like this:

```
[2024-10-21T13:16:37Z INFO  twoliter::project::lock::image] Resolving dependency image dependency 'bottlerocket-core-kit-3.0.0@public.ecr.aws/bottlerocket/bottlerocket-core-kit:v3.0.0'.
Error: Failed to run operation with image tool:
 command: /proc/9302/fd/9 manifest public.ecr.aws/bottlerocket/bottlerocket-core-kit:v3.0.0
```

We need to:
* Fix the `go build` of krane to link against `musl`
* Fix the logging on the `oci-tool` so that errors with executing the tool give us stdout/stderr as logs

sam-berning:
**Issue number:**

Closes #403 

**Description of changes:**

While looking into #403, I noticed that the `krane` binary produced in the build works as expected if I just copy the binary over to an affected system.

This PR changes the way that the `krane-bundle` lib installs krane. Instead of keeping the file in memory and sealing it, we create a temp directory and write the krane binary to that directory.

This PR also improves the error message when krane fails to run.

**Testing done:**

Built twoliter with

```bash
cross build --target=x86_64-unknown-linux-musl --release
```

and copied the produced binary over to a system that was affected by the bug in #403. I ran `twoliter update` in the bottlerocket repo without issue.

**Terms of contribution:**

### DIFF
--- a/tools/krane/Cargo.toml
+++ b/tools/krane/Cargo.toml
@@ -10,7 +10,7 @@ publish = false
 anyhow.workspace = true
 flate2.workspace = true
 lazy_static.workspace = true
-pentacle.workspace = true
+tempfile.workspace = true
 
 [build-dependencies]
 flate2.workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1902,8 +1902,8 @@ dependencies = [
  "anyhow",
  "flate2",
  "lazy_static",
- "pentacle",
  "tar",
+ "tempfile",
 ]
 
 [[package]]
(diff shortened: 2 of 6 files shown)


tools/krane/Cargo.toml
```diff
@@ -11,6 +11,7 @@ anyhow.workspace = true
 flate2.workspace = true
 lazy_static.workspace = true
 pentacle.workspace = true
```
cbgbt: I believe you can remove this dependency.

cbgbt (approved):

cbgbt:
Too bad that the sealed file bit struggles on some systems -- it's still not clear to me why. Nice work!

sam-berning:
^ address comments from @cbgbt and update krane README

cbgbt (approved):

sumukhballal (approved):

PR #420, opened 2024-12-18
twoliter: ensure krane tmpfile is deleted

### LINKED ISSUE #417: `krane` tempfiles are not properly cleaned up
### *Issue Description*
When using `twoliter`, my root volume became full due to `/tmp` usage. It turns out that there were many `.tmp` directories, each with their own unpacked copy of `krane`. It seems the `twoliter` is not properly cleaning up its `krane` binaries on exit.

After investigating, I believe it's due to the fact that we have a public static reference to a singular `Krane` struct that we use in `twoliter` (See [here](https://github.com/bottlerocket-os/twoliter/blob/77bf67e44c472159e8ff9356ebeba688a2eca0ce/tools/krane/src/lib.rs#L12)).

The `Krane` struct [uses `TempDir` to store its binary](https://github.com/bottlerocket-os/twoliter/blob/77bf67e44c472159e8ff9356ebeba688a2eca0ce/tools/krane/src/lib.rs#L18); however, the tempfile documentation [calls out that this will result in resource leaking](https://docs.rs/tempfile/3.14.0/tempfile/#resource-leaking) -- Rust must not be calling the `Drop` implementation for this static.

> `tempfile` will (almost) never fail to cleanup temporary resources. However `TempDir` and `NamedTempFile` will fail if their destructors don’t run. This is because `tempfile` relies on the OS to cleanup the underlying file, while `TempDir` and `NamedTempFile` rely on rust destructors to do so. Destructors may fail to run if the process exits through an unhandled signal interrupt (like SIGINT), or if the instance is declared statically (like with [`lazy_static`](https://github.com/rust-lang-nursery/lazy-static.rs/issues/62)), among other possible reasons.

That documentation also says:

> When choosing between the temporary file variants, prefer `tempfile` unless you either need to know the file’s path or to be able to persist it.

We *do* need to know the filepath to execute the binary that we write and thus must use one of the named options.

So rather than relying on a `KRANE` static, I think we need to rework the `krane-bundle` crate and users to create their own instance of `Krane` that they are guaranteed to drop.

cbgbt:
**Issue number:**

Closes #417

**Change Description:**
When using named tempfiles or tempdirs, the `tempfile` crate relies on Rust calling `drop()` in order to delete the created temporary files.

We were using a static reference to a tempfile of the krane binary, which is not guaranteed to drop when the program completes. This caused twoliter to leak somewhat large krane binary tempfiles.

This commit moves away from a static reference to the tempfile, instead improving ergonomics around ImageTool owning the tempfile reference. This allows the Rust compiler to more-reliably drop() the reference on exit.

**Testing done:**
* [x] Unit tests for drop behavior
* [x] Integration tests pass
* [ ] Manual checks that practical `twoliter` executions do not leak files 

**Terms of contribution:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1146,6 +1146,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "derivative"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fcc3dd5e9e9c0b295d6e1e4d811fb6f157d5ffd784b8d202fc62eac8035a770b"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+]
+
+[[package]]
 name = "digest"
 version = "0.10.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1904,6 +1915,7 @@ dependencies = [
  "lazy_static",
  "tar",
  "tempfile",
+ "tokio",
  "which",
 ]
 
@@ -3868,6 +3880,7 @@ dependencies = [
  "buildsys-config",
  "bytes",
  "clap",
+ "derivative",
  "env_logger",
  "filetime",
  "flate2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,6 +85,7 @@ chrono = { version = "0.4", default-features = false }
 clap = "4"
 coldsnap = { version = "0.6", default-features = false }
 daemonize = "0.5"
+derivative = "2"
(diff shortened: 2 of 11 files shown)


cbgbt:
`drop()` isn't called on `SIGTERM` so this doesn't actually protect us in the common scenario that someone uses ctrl-c while `twoliter` is running.

I think the solution should be to unpack `krane` to a well-known location like we do for other tools. I'll take a look at that.

cbgbt:
Closed in favor of #423

PR #423, opened 2024-12-23
Require Docker 23, Statically link Krane

### LINKED ISSUE #417: `krane` tempfiles are not properly cleaned up

cbgbt:
**Issue number:**

Closes #417

**Description of changes:**

---

    krane: statically link krane into twoliter

    This change implements a small FFI library in golang which emulates the
    `krane` program from google/go-containerregistry.

    Prior to this change, we'd experimented with two approaches to using
    `krane` within twoliter, but both approaches had unfortunate
    complications:

    * Writing `krane` to a sealed anonymous file and executing it -- this
      approach would cause twoliter to be killed by some threat detection
      software. Executing sealed anonymous files is commonly restricted on
      Linux systems
    * Writing `krane` to a tempfile and executing it -- it's difficult to
      get Rust to reliably delete named tempfiles, which led to twoliter
      leaking tempfiles of the `krane` binary.

---

    twoliter: require Docker 23

    Docker versions older than 23 no longer receive community-supported
    security patches.

    Additionally, twoliter requires dockerfile syntax 1.4.3. Currently the
    tool pulls this image from the network at build time; however, Docker 23
    bundles this version internally via buildkit. We can further isolate
    twoliter builds from the network by ensuring we are on Docker 23 or
    greater.

---

    twoliter: refactor project to reduce module length

    The project module is large, making it difficult to organize and read.
    This moves much of the vended image/artifact machinery into a new
    module. We also add a new trait to help us write functionality over
    both Projects that are fully locked and Projects with a locked SDK.

---

    twoliter: fetch build SDK prior to cargo-make

    Now that `krane` is run as a static library, we cannot execute it within
    Makefile.toml.

    This change moves the SDK fetch to occur prior to executing build make
    targets.

---

**Testing done:**
* [x] Integration tests
* [x] Manual testing of signal handler (ctrl-c while downloading SDK)
* [x] Manual testing of kit publication
* [x] Test with AWS Codebuild (used by bottlerocket team for release CI)

**Terms of contribution:**


Cargo.toml
```diff
@@ -84,6 +84,7 @@ bytes = "1"
 chrono = { version = "0.4", default-features = false }
 clap = "4"
 coldsnap = { version = "0.6", default-features = false }
+ctrlc = "3"
```
yeazelm, line 87: Do we want `termination` to catch more than `SIGINT`?

tools/krane/go-src/main.go
```diff
@@ -0,0 +1,108 @@
...
+	args := parseCArgs(argc, argv)
+
+	statusCode := kraneMain(args, true, nil, nil)
+
+	return C.int(statusCode)
+}
+
+func parseCArgs(argc C.int, argv **C.char) []string {
+	args := make([]string, 0, argc)
+	for i := 0; i < int(argc); i++ {
```
yeazelm, line 66: nit: I think Go 1.22 will let you simplify this:
```suggestion
	for range int(argc) {
```
For what its worth, it's still idiomatic how you have it.

yeazelm (approved):

cbgbt:
^ addresses feedback from @yeazelm, as well as experimenting with fixes for build failures on other hosts.

Basically, I had originally generated `go.{mod,sum}` using go-1.23. After downgrading the required version of go to `1.22`, I needed to delete my indirect dependencies and re-run `go mod tidy`.

tools/krane/go-src/main.go
cbgbt: The go compiler didn't like this for some reason, and also wasn't happy about binding a name to the iterator that I didn't use, so I've left it as-is.

yeazelm: Hmm, strange, my playground experiment was happy with this but as I said, its perfectly acceptable as you have it.

cbgbt:
I made registration of the cleanup signal only fail with a warning. The signal registration fails in AWS CodeBuild for a reason that isn't quite clear to me.

But cleanup on signals is already best-effort, I think proceeding without it is acceptable.

Cargo.lock
```diff
@@ -4195,7 +4231,7 @@ version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf221c93e13a30d793f7645a0e7762c55d169dbb0a49671918a2319d289b10bb"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.48.0",
```
jpculp, line 4245: Interesting that it dropped.

jpculp (approved):

sam-berning (commented):
Nice work!

sam-berning (approved):

PR #431, opened 2025-01-03
twoliter: disallow std::path::Path::canonicalize

### LINKED ISSUE #427: Use of fs::canonicalize makes Twoliter difficult to wrap in other build systems
I am now trying to wrap Twoliter in Bazel (yay!). Bazel famously uses symlink farms to produce sandboxes efficiently. Since you, in Twoliter here: https://github.com/bottlerocket-os/twoliter/blob/develop/twoliter/src/project/mod.rs#L77C20-L77C36 use `fs::canonicalize`, which is essentially `realpath` - Twoliter tries to escape the sandbox to the real repository and gets write errors since Bazel mounts the real repository into the sandbox with `ro` permissions.

Using something like [fs::absolute](https://doc.rust-lang.org/beta/std/path/fn.absolute.html) would give you similar behaviour, but not break symlink-reliant sandboxing.

I do think people who use symlinks in their repository/build environment do not expect their build output to end up where the destination of their symlink is, so making this change should (hopefully) not negatively impact any existing users.

There are a few more uses (as evidenced by this search: https://github.com/search?q=repo%3Abottlerocket-os%2Ftwoliter+canonicalize&type=code) but the usage seems relatively limited.

The `fs::absolute` however requires a relatively new version of Rust as it was introduced in v1.79 which was released in June, but I do think it should be in the nightly you have pinned Twoliter to.

cbgbt:
**Issue number:**

Closes #427

**Description of changes:**
Twoliter uses the location of `Twoliter.toml` to locate the rest of the Twoliter workspace. Prior to this change, it [used `std::fs::canonicalize`](https://doc.rust-lang.org/std/fs/fn.canonicalize.html) to do so. This function resolves symlinks on the filesystem, which is a problem when you're building in the context of a symlink farm.

This change moves to using [path-absolutize](https://crates.io/crates/path-absolutize) for such canoncicalization, which uses unix separator and path traversal semantics to "canonicalize" a path without resolving symlinks or actually consulting with the host's filesystem.

The change also adds clippy lints to attempt to forbid future use of canonicalization functions that can cause this to regress.

See https://github.com/bottlerocket-os/twoliter/issues/427 for more details.

**Testing done:**
* New integration tests pass
* Created a twoliter workspace where `Twoliter.toml` was a symlink to another workspace. Issuing a build on the old version caused build artifacts to be placed in the `build` directory behind the symlink. With the new change, build artifacts were placed in the new workspace as expected.

**Terms of contribution:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2293,6 +2293,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "path-absolutize"
+version = "3.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e4af381fe79fa195b4909485d99f73a80792331df0625188e707854f0b3383f5"
+dependencies = [
+ "path-dedot",
+]
+
+[[package]]
+name = "path-dedot"
+version = "3.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "07ba0ad7e047712414213ff67533e6dd477af0a4e1d14fb52343e53d30ea9397"
+dependencies = [
+ "once_cell",
+]
+
+[[package]]
 name = "pathdiff"
 version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


sam-berning (approved):

sumukhballal:
Nice!

sumukhballal (approved):

mikn:
You were so quick closing this out I didn't even notice! Thank you @cbgbt :heart:

Issue #378, opened 2024-09-23
pubsys: Allow creation of arbirary additional block device mappings when registering AMI


cbgbt:
It would be useful to be able to attach arbitrary snapshots or create new volumes for AMIs registered via `twoliter`.

We'd need for the feature to ensure that the provided block device mappings do not clash with the ones generated for use with the data volume when using the `split` volume layout.

PR #438, opened 2025-01-10
VOLUME_TYPE: gp2 -> gp3

awoimbee:
**Issue number:** https://github.com/bottlerocket-os/bottlerocket/issues/4000#issuecomment-2342962331
It relates to this issue but I don't think this will change how Bottlerocket is released today, that seems to be closed sources ?

**Description of changes:** Set EBS disk type to gp3 instead of gp2

**Testing done:** on twoliter itself, just cargo test. On Bottlerocket, it's the setup I've been using for a while, benchmarks show no difference.

Note that gp3 was already used in other parts of the code.

**Terms of contribution:**

### DIFF
--- a/tools/pubsys/src/aws/ami/register.rs
+++ b/tools/pubsys/src/aws/ami/register.rs
@@ -15,7 +15,7 @@ const DATA_DEVICE_NAME: &str = "/dev/xvdb";
 
 // Features we assume/enable for the images.
 const VIRT_TYPE: &str = "hvm";
-const VOLUME_TYPE: &str = "gp2";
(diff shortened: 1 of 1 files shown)


cbgbt:
Hi, thanks for your contribution

Bottlerocket does indeed release using `twoliter` today, so we can't make this change directly without impacting that build process. Sorry about that.

I think the path forward here would be to teach `twoliter`/`pubsys` how to take something akin to EC2 Block Device Mappings as input. https://github.com/bottlerocket-os/twoliter/issues/378 describes that concept.

Please let me know if you have any questions!

PR #440, opened 2025-01-14
Revert krane static and prepare for release v0.7.0-rc2

sam-berning:
**Issue number:**

Closes #

**Description of changes:**

Revert krane-static changes for now, as they were causing issues for the release build of twoliter `v0.7.0-rc1`, but we want to release the other changes in `develop`. 

**Testing done:**

Cross built twoliter locally:

```bash
cross build --release --bin twoliter --target x86_64-unknown-linux-musl
```

Then, copied the resulting binary into the bottlerocket repo, and ran

```bash
cargo make -e BUILDSYS_VARIANT=aws-dev
```

**Terms of contribution:**


sam-berning:
The last revert commit had a merge conflict in twoliter/Cargo.toml and Cargo.lock, which was just because krane-static had changed name back to krane-bundle. Also, needed to remove the `cleanup` module added [here](https://github.com/bottlerocket-os/twoliter/commit/cb4d91b630ca4f781085c21866e057e025f0bb35#diff-17637ec1f9ed4d7534d3f20801e890735888e17ae1df2a8ffa8f194e5b1d5b43R6) as the corresponding module wasn't added until [`faac675`](https://github.com/bottlerocket-os/twoliter/commit/faac675f1c594e7c1f037561e032b906ae27534d). Otherwise, the reverts were clean.

sumukhballal (approved):

cbgbt (approved):
I took a deeper look at the issues we were having. gdb uncovered that the golang runtime was de-referencing garbage when [attempting to read `argv`](https://github.com/golang/go/blob/368a9ec99834652ca3f7d8fe24862a7581e12358/src/runtime/runtime1.go#L63). It turns out that this is the result of [a longstanding issue opened against golang](https://github.com/golang/go/issues/13492), which depends on glibc-specific initialization of shared libraries in a way that is not specified by the ELF gABI (and therefor not also done by musl).

The Golang project [does not officially support non-glibc C standard libraries](https://go.dev/wiki/PortingPolicy):

> All Linux first class ports are for systems using glibc only. Linux systems using other C libraries are not fully supported and are not treated as first class.

Thanks for catching this, Sam! I must have mistakenly been testing my changes using a glibc-linked build.

PR #443, opened 2025-01-23
imghelper: set AWS_CA_BUNDLE if cert is available

sumukhballal:
**Issue number:**

If a CA-Bundle is set using a secret mount and present, it will update the SDK to use these certs instead. Also sets the aws variable AWS_CA_BUNDLE if there is a secret mount. The secret mount is created when we pass a CABundle via override using the environment variable `BUILDSYS_CACERTS_BUNDLE_OVERRIDE`.

Note: I tried setting `SSL_CERT_FILE` variable instead but it causes the curl utility in [aws-kms-pkcs11](https://github.com/JackOfMostTrades/aws-kms-pkcs11) to fail.

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/imghelper
+++ b/twoliter/embedded/imghelper
@@ -430,6 +430,11 @@ sbsetup_wrapup() {
 }
 
 sbsetup_aws_profile() {
+  # Use the CA bundle override as the AWS CA cert bundle, if present.
+  if [[ -s "/root/certs/ca-bundle.crt" ]]; then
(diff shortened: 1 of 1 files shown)


bcressey (approved):
You can try setting `CURL_CA_BUNDLE` instead to see if that overrides the libcurl default.

(It's the AWS SDK for C++ behavior you're trying to change, not the KMS proxy.)

This looks OK to me; it's a bit rude to clobber an SDK file, but it should be fine for now.

cbgbt (approved):

PR #467, opened 2025-02-24
twoliter: add check-advisories task to lint BRSAs

### LINKED ISSUE #466: Add linting support for Bottlerocket Security Advisories
https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/381 highlights a need for linting of Bottlerocket Security Advisories (BRSAs) to enforce that advisory content is strictly ASCII characters. 

BRSAs are the source of Bottlerocket's [updateinfo.xml](https://advisories.bottlerocket.aws/updateinfo.xml.gz). If a BRSA encodes special characters, then the encoding of those characters to XML may result in malformed updates in updateinfo.xml. For example, text copied directly from https://nvidia.custhelp.com/app/answers/detail/a_id/5616 for CVE-2025-23359 uses a special "non-breaking hyphen" character for hyphens, resulted in encodings `CVE&#x2011;2025&#x2011;23359` instead of using a normal "-"

Something like https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/381#discussion_r1966305918

```
if find advisories -name '*.toml' -type f >/dev/null 2>&1 ; then
  export LC_ALL=C
  if grep --include '*.toml' -R -l -P ['\x80'-'\xFF'] advisories ; then
    echo "error: found non-ASCII characters in advisories" >&2
    exit 1
  fi
fi
```

may do the trick

ginglis13:
**Issue number:**

Closes #466 

**Description of changes:**

Lint BRSAs for CVE / GHSA IDs in a new task, check-advisories. Also ensure that each directory under "advisories" in a project has an associated tag on the project.

Add this to the list in the meta task "check"

**Testing done:**

[x] Run against kernel-kit with bad CVE IDs

```
╭─fedora ~/kernel-kit  ‹v1.1.3*›
╰─➤  ./tools/twoliter/twoliter make check-advisories --cargo-home /home/fedora/kernel-kit/.cargo --arch $(uname -m)
[2025-02-28T20:20:24Z INFO  twoliter::project::lock] Resolving SDK project reference to check against lock file
[2025-02-28T20:20:25Z INFO  twoliter::project::lock::image] Resolving dependency image dependency 'bottlerocket-sdk-0.50.1@public.ecr.aws/bottlerocket/bottlerocket-sdk:v0.50.1'.
[cargo-make] INFO - cargo make 0.37.24
[cargo-make] INFO -
[cargo-make] INFO - Build File: /home/fedora/kernel-kit/build/tools/Makefile.toml
[cargo-make] INFO - Task: check-advisories
[cargo-make] INFO - Running Task: check-advisories
error: advisory CVE ID did not match expression '^cve\s+=\s+"CVE-\d{4}-(0\d{3}|[1-9]\d{3,})"' and may contain non-ASCII characters
cve = "CVE‑2024‑53869"
cve = "CVE‑2024‑0150"
cve = "CVE‑2024‑0131"
cve = "CVE‑2024‑0149"
cve = "CVE‑2024‑0147"
Error while executing command, exit code: 1
Error: Command was unsuccessful, exit code 105
```

Fixed in https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/55

[x] Run against kernel-kit with bad GHSA ID (mocked advisory)

```
╰─➤  ./tools/twoliter/twoliter make check-advisories --cargo-home /home/fedora/kernel-kit/.cargo --arch $(uname -m)
[2025-02-28T20:22:47Z INFO  twoliter::project::lock] Resolving SDK project reference to check against lock file
[2025-02-28T20:22:47Z INFO  twoliter::project::lock::image] Resolving dependency image dependency 'bottlerocket-sdk-0.50.1@public.ecr.aws/bottlerocket/bottlerocket-sdk:v0.50.1'.
[cargo-make] INFO - cargo make 0.37.24
[cargo-make] INFO -
[cargo-make] INFO - Build File: /home/fedora/kernel-kit/build/tools/Makefile.toml
[cargo-make] INFO - Task: check-advisories
[cargo-make] INFO - Running Task: check-advisories
error: advisory GHSA ID did not match expression '^ghsa\s+=\s+"GHSA(-[23456789cfghjmpqrvwx]{4}){3}"' and may contain non-ASCII characters
ghsa = "GHSA-0000-5224-cccc"
Error while executing command, exit code: 1
Error: Command was unsuccessful, exit code 105
```

[x] Run in a Twoliter project without an `advisories` subdirectory

After ASCII fixes, run against kernel-kit:

```
╭─fedora ~/kernel-kit  ‹develop*›
╰─➤  ./tools/twoliter/twoliter make check-advisories --cargo-home /home/fedora/kernel-kit/.cargo --arch $(uname -m)
[2025-02-25T22:57:43Z INFO  twoliter::project::lock] Resolving SDK project reference to check against lock file
[2025-02-25T22:57:43Z INFO  twoliter::project::lock::image] Resolving dependency image dependency 'bottlerocket-sdk-0.50.1@public.ecr.aws/bottlerocket/bottlerocket-sdk:v0.50.1'.
[cargo-make] INFO - cargo make 0.37.24
[cargo-make] INFO -
[cargo-make] INFO - Build File: /home/fedora/kernel-kit/build/tools/Makefile.toml
[cargo-make] INFO - Task: check-advisories
[cargo-make] INFO - Running Task: check-advisories
[cargo-make] INFO - Build Done in 0.22 seconds.
```

Test case of unexpected directory under `advisories/`:

```
╭─fedora ~/kernel-kit  ‹develop*›
╰─➤  ls advisories
0  1.0.2  1.0.4  1.0.6  1.0.7  1.1.0  staging
╭─fedora ~/kernel-kit  ‹develop*›
╰─➤  ./tools/twoliter/twoliter make check-advisories --cargo-home /home/fedora/kernel-kit/.cargo --arch $(uname -m)
[2025-02-25T22:55:54Z INFO  twoliter::project::lock] Resolving SDK project reference to check against lock file
[2025-02-25T22:55:55Z INFO  twoliter::project::lock::image] Resolving dependency image dependency 'bottlerocket-sdk-0.50.1@public.ecr.aws/bottlerocket/bottlerocket-sdk:v0.50.1'.
[cargo-make] INFO - cargo make 0.37.24
[cargo-make] INFO -
[cargo-make] INFO - Build File: /home/fedora/kernel-kit/build/tools/Makefile.toml
[cargo-make] INFO - Task: check-advisories
[cargo-make] INFO - Running Task: check-advisories
error: no corresponding tag found for advisories/0 directory
Error while executing command, exit code: 1
Error: Command was unsuccessful, exit code 105
```

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/Makefile.toml
+++ b/twoliter/embedded/Makefile.toml
@@ -416,6 +416,7 @@ done
 [tasks.check]
 dependencies = [
    "check-cargo-version",
+   "check-advisories",
    "unit-tests",
    "check-fmt",
    "check-lints",
@@ -536,6 +537,61 @@ fi
 if [ "${rc}" -ne 0 ]; then
   echo "Found lint warnings. First-party shell code is checked with ShellCheck." >&2
   exit $rc
+fi
+'''
+]
+
+# Task to lint Bottlerocket Security Advisories by checking for valid CVE or GHSA IDs
+# and verifying that each versioned directory under "advisories" has a corresponding
+# tag on the Twoliter project this task runs against.
+[tasks.check-advisories]
+script_runner = "bash"
+script = [
+'''
+if find advisories -name '*.toml' -type f >/dev/null 2>&1 ; then
+
+  # Ensure each versioned advisories directory has a corresponding release tag.
+  for version in $(find advisories/* -type d -not -path advisories/staging); do
+    set +e; grep -q v$(basename ${version})$ <(PAGER= git tag); rc="$?"; set -e;
+    if [ "${rc}" -ne 0  ]; then
+      echo "error: no corresponding tag found for ${version} advisories directory" >&2
+      exit 1
+    fi
+  done
+
+
+  # Not all BRSAs might have a CVE; there can be cases where an advisory
+  # is for a GHSA, for example, but no corresponding CVE, and vice versa.
+  # Check separately for GHSA ID regex when a 'ghsa' is included and for
+  # CVE ID regex when a 'cve' is included.
+
+  # 1. If ghsa line exists and GHSA ID does not match regex, error
+
+  # Regex to strictly match a GHSA identifier in an advisory
+  # https://github.com/github/advisory-database?tab=readme-ov-file#ghsa-ids
+  ghsa_regex="^ghsa\s+=\s+\"GHSA(-[23456789cfghjmpqrvwx]{4}){3}\""
+
+  # Find all non-matching GHSA lines and report
+  ghsa_found="$(grep -L --include '*.toml' -R -P ${ghsa_regex} advisories | xargs awk '/ghsa/')"
+  if [ ! -z "${ghsa_found}"  ] ; then
+    echo "error: advisory GHSA ID did not match expression '${ghsa_regex}' and may contain non-ASCII characters" >&2
+    echo "${ghsa_found}" >&2
+    exit 1
+  fi
+
+  # 2. If cve line exists and CVE ID does not match regex, error
+
+  # Regex to strictly match CVE identifier in an advisory
+  # https://cve.mitre.org/cve/identifiers/tech-guidance.html#input_format
+  cve_regex="^cve\s+=\s+\"CVE-\d{4}-(0\d{3}|[1-9]\d{3,})\""
+
+  # Find all non-matching CVE lines and report
+  cve_found="$(grep -L --include '*.toml' -R -P ${cve_regex} advisories | xargs awk '/cve/')"
+  if [ ! -z "${cve_found}"  ] ; then
+    echo "error: advisory CVE ID did not match expression '${cve_regex}' and may contain non-ASCII characters" >&2
+    echo "${cve_found}" >&2
+    exit 1
+  fi
 fi
 '''
 ]
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -20,9 +20,10 @@ jobs:
       - uses: actions/setup-go@v5
         with:
           go-version: "^1.18"
-      # Install `patch`, needed to build `krane-bundle`
-      - run: sudo apt-get install -y patch
-      - run: make build
+      # Install `patch`, needed to build `krane-bundle`, and tools for linking against musl
+      - run: sudo apt-get install -y patch musl-tools musl-dev
+      - run: rustup target add x86_64-unknown-linux-musl
+      - run: CARGO_BUILD_TARGET=x86_64-unknown-linux-musl make build
 
   cross-build:
     runs-on:
--- a/clarify.toml
+++ b/clarify.toml
@@ -5,7 +5,7 @@ license-files = [
     { path = "COPYING", hash = 0x278afbcf },
     { path = "LICENSE-APACHE", hash = 0x24b54f4b },
     { path = "LICENSE-MIT", hash = 0x462dee44 },
-    { path = "src/unicode/data/LICENSE-UNICODE", hash = 0x70f7339 },
+    { path = "src/unicode/data/LICENSE-UNICODE", hash = 0x70f7339  },
 ]
 
 [clarify.petgraph]
@@ -26,6 +26,23 @@ license-files = [
     { path = "LICENSE-MIT", hash = 0xb755395b },
     { path = "src/unicode_tables/LICENSE-UNICODE", hash = 0xa7f28b93 },
 ]
+
+[clarify.rust-fuzzy-search]
+expression = "MIT OR Apache-2.0"
+license-files = [
+    { path = "LICENSE-APACHE", hash = 0xbde481e5 },
+    { path = "LICENSE-MIT", hash = 0xb5a90d39 },
+]
+skip-files = [
+    # these licenses apply to documentation
+    "target/doc/FiraSans-LICENSE.txt",
+    "target/doc/COPYRIGHT.txt",
+    "target/doc/LICENSE-APACHE.txt",
+    "target/doc/LICENSE-MIT.txt",
+    "target/doc/SourceCodePro-LICENSE.txt",
+    "target/doc/SourceSerif4-LICENSE.md",
+]
+
 
 [clarify.typenum]
 expression = "MIT OR Apache-2.0"
--- a/deny.toml
+++ b/deny.toml
@@ -78,8 +78,6 @@ skip = [
     { name = "tabled", version = "0.15.0" },
     # multiple deps are using an older version of tabled_derive
     { name = "tabled_derive", version = "0.7.0" },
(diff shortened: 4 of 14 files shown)


ginglis13:
```
error: cannot install package `cargo-deny 0.18.0`, it requires rustc 1.85.0 or newer, while the currently active rustc version is 1.81.0-nightly
`cargo-deny 0.17.0` supports rustc 1.81.0
```

I'll update to a newer nightly rust compiler in an additional commit to this PR

ginglis13:
^ added a commit for updating nightly rust toolchain to `2025-02-24`. Still working through some issues with cargo-deny. Currently it is flagging all workspace dependencies as unused.

e: working through 

```
Error: -25 01:21:22 [ERROR] failed to interpret `cargo metadata`'s json: expected `,` or `}` at line 1 column 251366: expected `,` or `}` at line 1 column 251366
```

now. This is strange because as far as I can tell, `cargo metadata` is returning valid JSON - `cargo metadata | jq` works fine.

ginglis13:
^ force push adds a check to ensure that each versioned directory under `advisories/` has a corresponding git tag

twoliter/embedded/Makefile.toml
```diff
@@ -540,6 +541,31 @@ fi
...
+if find advisories -name '*.toml' -type f >/dev/null 2>&1 ; then
+  export LC_ALL=C
+  if grep --include '*.toml' -R -l -P ['\x80'-'\xFF'] advisories ; then
+    echo "error: found non-ASCII characters in advisories" >&2
+    exit 1
+  fi
+
+  for version in $(find advisories/* -type d -not -path advisories/staging); do
+    grep v$(basename ${version})$ <(PAGER= git tag)
+    if [ "$?" -ne 0  ]; then
```
ginglis13: This implies a dependency on the project this check runs against having all tags fetched. Curious for feedback on either:
* fetching tags in this check
* or, removing this as a default check and running this check as a new GitHub CI step on projects. Those projects will set `fetch-tags: true` in respective `actions/checkout` steps

ginglis13:
^ force push handles return code from `grep` more gracefully with error message. Testing added to PR description

ginglis13:
^ force push pins `cargo-deny` to v0.17.0 and reverts toolchain upgrade

ginglis13:
^ force push based on some offline feedback from @cbgbt . We should only be linting on structured data, which in this case is a CVE or GHSA ID. Adjusted the lint to validate those IDs rather than lint entire advisory files for non-ASCII chars

Also updated to latest rust nightly toolchain now that [cargo-deny 0.18.1](https://github.com/EmbarkStudios/cargo-deny/releases/tag/0.18.1) is released with a fix for https://github.com/EmbarkStudios/krates/issues/100

ginglis13:
^ force push links against musl in CI for `make build`. integ test failures like

```
 0.166 /host/build/tools/unplug: /lib64/libc.so.6: version `GLIBC_2.39' not found (required by /host/build/tools/unplug)
```

Indicated that builds were failing due to an older glibc coming from the bottlerocket-sdk https://github.com/bottlerocket-os/bottlerocket-sdk/blob/7b29868e87346b1538403509a79c22a67a463bce/Dockerfile#L186

jmt-lab (approved):

twoliter/embedded/Makefile.toml
```diff
@@ -540,6 +541,61 @@ fi
...
 ]
 
+# Task to lint Bottlerocket Security Advisories by checking for valid CVE or GHSA IDs
+# and verifying that each versioned directory under "advisories" has a corresponding
+# tag on the Twoliter project this task runs against.
+[tasks.check-advisories]
+script_runner = "bash"
+script = [
+'''
+if find advisories -name '*.toml' -type f >/dev/null 2>&1 ; then
```
cbgbt, line 551:
```suggestion
if find ${BUILDSYS_ROOT_DIR}/advisories -name '*.toml' -type f >/dev/null 2>&1 ; then
```

We should use full filepaths in case the current working directory is not set as expected. There are a few cases in here that refer to the `advisories` directory that require this.

twoliter/embedded/Makefile.toml
```diff
@@ -540,6 +541,61 @@ fi
...
+  fi
+
+  # 2. If cve line exists and CVE ID does not match regex, error
+
+  # Regex to strictly match CVE identifier in an advisory
+  # https://cve.mitre.org/cve/identifiers/tech-guidance.html#input_format
+  cve_regex="^cve\s+=\s+\"CVE-\d{4}-(0\d{3}|[1-9]\d{3,})\""
+
+  # Find all non-matching CVE lines and report
+  cve_found="$(grep -L --include '*.toml' -R -P ${cve_regex} advisories | xargs awk '/cve/')"
```
cbgbt, line 589: If I understand right, these checks will fail if an advisory document contains the string `cve` or `ghsa` but does not contain a valid CVE or GHSA line.

I think what we actually want to do is semantically parse the advisory TOML document and check that the CVE and GHSA lines are valid, right? I think parsing with regexes like this leads to a lot of potential for surprising and valid inputs to get caught in the crossfire.

Can we at least constrain the mismatch checks to lines that start with `cve`/`ghsa` rather than the entirety of the document? That way we don't misfire on freeform text e.g. in an advisory description that happens to have the character string `cve` in it.

---


In general I want to see code like this in Twoliter hoisted into Rust so that we can get unit test coverage and a higher ability to take advantage of software libraries like a toml parser. We don't really have a "recommended" way to do that. In a pinch we could probably have Twoliter provide a helper binary (`oneliter`?) under `tools/` to run some of these tasks, but in the longer term we probably want to express the task graph currently in Makefile.toml inside Twoliter and drop cargo-make as a dependency.

ginglis13: > Can we at least constrain the mismatch checks to lines that start with cve/ghsa rather than the entirety of the document? That way we don't misfire on freeform text e.g. in an advisory description that happens to have the character string cve in it.

You're right, the `awk` half of this pipeline would break down on any inclusion of "cve"/"ghsa", including advisory descriptions. I'll modify those respective checks to match on line start.

> In general I want to see code like this in Twoliter hoisted into Rust so that we can get unit test coverage and a higher ability to take advantage of software libraries like a toml parser.

I totally agree; that along with including advisory types, tools, and models in twoliter are a more robust long term effort

ginglis13:
Chatted offline with @cbgbt . I'm going to take this approach - add a new `lints` crate under [twoliter/tools](https://github.com/bottlerocket-os/twoliter/tree/develop/tools) that provides multiple binaries. I'll write a binary for the lint I've prepared in this PR, embed that binary in `twoliter`, and invoke it under the `check-advisories` task

ginglis13:
superseded by https://github.com/bottlerocket-os/twoliter/pull/604

PR #469, opened 2025-02-28
pubsys: add --dry-run mode

### LINKED ISSUE #356: pubsys: add dry-run option to SSM commands
## Request

Add a flag to `pubsys ssm` and `pubsys promote-ssm` that causes those commands to output the SSM parameters that would be pushed, but not actually push them.

## Why it's useful

This will allow users to ensure that the parameters they are pushing look correct before pushing them. It also decouples generating the files used to validate the parameters from actually publishing the parameters.

qianxjcraig:
**Issue number:**

Closes #356

**Description of changes:**
Add `--dry-run` flag to `pubsys ssm` subcommand, allowing users to generate a manifest of SSM parameter updates without actually modifying AWS SSM parameters. 

**Testing done:**

- [x] Exit early when `--dry-run` and `-ssm-parameter-output` set.
```
00:20:03 [INFO] Wrote rendered SSM parameters to "/local/home/qianxj/bottlerocket/build/images/x86_64-aws-k8s-1.30/1.25.0-f56453df/bottlerocket-aws-k8s-1.30-x86_64-1.25.0-f56453df-ssm-params.json"
00:20:03 [INFO] Dry-run mode enabled: Exiting after writing parameter manifest.
[cargo-make][1] INFO - Build Done in 0.29 seconds.
```
- [x] Error when `--ssm-parameter-output` not set when `--dry-run` enabled.
```
error: the following required arguments were not provided:
  --ssm-parameter-output <SSM_PARAMETER_OUTPUT>

Usage: pubsys --infra-config-path <INFRA_CONFIG_PATH> ssm --ami-input <AMI_INPUT> --arch <ARCH> --variant <VARIANT> --version <VERSION> --template-path <TEMPLATE_PATH> --ssm-parameter-output <SSM_PARAMETER_OUTPUT> --dry-run

For more information, try '--help'.
```

**Terms of contribution:**

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -15,7 +15,7 @@ jobs:
       labels: bottlerocket_ubuntu-latest_16-core
     steps:
       - uses: actions/checkout@v3
-      - run: cargo install cargo-deny --locked
+      - run: cargo install cargo-deny@0.17.0 --locked
       - run: cargo install cargo-make --locked
       - uses: actions/setup-go@v5
         with:
--- a/tools/pubsys/src/aws/ssm/mod.rs
+++ b/tools/pubsys/src/aws/ssm/mod.rs
@@ -15,7 +15,7 @@ use crate::Args;
 use aws_config::SdkConfig;
 use aws_sdk_ec2::{types::ArchitectureValues, Client as Ec2Client};
 use aws_sdk_ssm::{config::Region, Client as SsmClient};
-use clap::Parser;
+use clap::{ArgGroup, Parser};
 use futures::stream::{StreamExt, TryStreamExt};
 use governor::{prelude::*, Quota, RateLimiter};
 use log::{error, info, trace};
@@ -32,6 +32,11 @@ use std::{
 
 /// Sets SSM parameters based on current build information
 #[derive(Debug, Parser)]
+#[command(group(
+    ArgGroup::new("dry_run_requirements")
+        .arg("dry_run")
+        .requires("ssm_parameter_output")
+))]
 pub(crate) struct SsmArgs {
     // This is JSON output from `pubsys ami` like `{"us-west-2": "ami-123"}`
     /// Path to the JSON file containing regional AMI IDs to modify
@@ -69,6 +74,10 @@ pub(crate) struct SsmArgs {
     /// If set, writes the generated SSM parameters to this path
     #[arg(long)]
     ssm_parameter_output: Option<PathBuf>,
+
+    /// Enables dry-run mode (only generates parameter manifest, does not update SSM)
+    #[arg(long)]
+    dry_run: bool,
 }
 
 /// Wrapper struct over parameter update and AWS clients needed to execute on it.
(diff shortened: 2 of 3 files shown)


.github/workflows/rust.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       labels: bottlerocket_ubuntu-latest_16-core
     steps:
       - uses: actions/checkout@v3
-      - run: cargo install cargo-deny --locked
+      - run: cargo install cargo-deny@0.17.0 --locked
```
jmt-lab, line 18: This may need to be undone when gavin's change merges due to the 1.85 changes

jmt-lab (approved):

tools/pubsys/src/aws/ssm/mod.rs
```diff
@@ -135,6 +139,14 @@ pub(crate) async fn run(args: &Args, ssm_args: &SsmArgs) -> Result<()> {
...
     trace!("Generated templated parameters: {:#?}", new_parameters);
 
+    // Ensure `--ssm-parameter-output` is set when using `--dry-run`
+    ensure!(
+        !(ssm_args.dry_run && ssm_args.ssm_parameter_output.is_none()),
+        error::MissingArgumentSnafu {
+            argument: "--ssm-parameter-output is required when using --dry-run".to_string()
+        }
+    );
+
```
cbgbt: I think the idiomatic way to do this with the `clap` library is [with `ArgGroup`s](https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html#argument-relations). Mind trying that out instead?

twoliter/embedded/Makefile.toml
```diff
@@ -1473,7 +1473,8 @@ pubsys \
    --ssm-parameter-output "${ssm_parameter_output}" \
    \
    ${PUBLISH_REGIONS:+--regions "${PUBLISH_REGIONS}"} \
-   ${ALLOW_CLOBBER:+--allow-clobber}
+   ${ALLOW_CLOBBER:+--allow-clobber} \
```
cbgbt, line 1476: Does `ALLOW_CLOBBER` actually work despite not being prefixed with `PUBLISH_`?

cbgbt (commented):
Nice work! Just a few comments.

twoliter/embedded/Makefile.toml
qianxjcraig: Yes, tested manually. `ALLOW_CLOBBER` works despite not having the `PUBLISH_` prefix because it's explicitly included in the environment variables list. 
```
cargo make -e BUILDSYS_ARCH=x86_64 \
-e BUILDSYS_VARIANT=aws-k8s-1.30 \
-e PUBLISH_SSM_DRY_RUN=true \
-e ALLOW_CLOBBER=true \            
-e TWOLITER_ALLOW_SOURCE_INSTALL=false \
-e TWOLITER_ALLOW_BINARY_INSTALL=false \
-e TWOLITER_SKIP_VERSION_CHECK=true \
ssm
```
While the `is_build_system_env()` function in [`cargo_make.rs`.](https://github.com/bottlerocket-os/twoliter/blob/2519e56182d028f02cbd5d01af6012b0b7c73461/twoliter/src/cargo_make.rs#L191) only checks and filters which shell environment variables get forwarded.

But one thing worth note is setting `-e ALLOW_CLOBBER=false` or `-e ALLOW_CLOBBER=true` will both enable this flag. And if we want to disable it, simply omit the variable entirely. Same for `PUBLISH_SSM_DRY_RUN`.  Because the shell parameter expansion ${ALLOW_CLOBBER:+--allow-clobber} only checks if parameter is unset or null.

cbgbt: Got it, thanks for explaining!

cbgbt (approved):

PR #470, opened 2025-02-28
fix: move vendoring of cargo dependencies to cargo vendor

jmt-lab:
**Description of changes:**
Migrates how we vendor cargo dependencies to use cargo vendor and ensure we do not sync cargo internal directories in .cargo where incompatibilities can occur such as the one we are fixing:

https://github.com/bottlerocket-os/bottlerocket/issues/4415 

**Testing done:**

Built bob with a host using cargo 1.85.0 and verified that it builds sucessfully

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/Makefile.toml
+++ b/twoliter/embedded/Makefile.toml
@@ -354,10 +354,20 @@ dependencies = ["setup"]
 script_runner = "bash"
 script = [
 '''
+declare -a CARGO_MANIFESTS
+# We make sure there is an empty folder at least here
+# to make the docker mount happy later on
+mkdir -p .cargo/vendor
 for ws in sources variants .; do
-  [ -s "${BUILDSYS_ROOT_DIR}/${ws}/Cargo.toml" ] || continue
-  cargo fetch --locked --manifest-path "${BUILDSYS_ROOT_DIR}/${ws}/Cargo.toml"
+  [ -s "${BUILDSYS_ROOT_DIR}/${ws}/Cargo.lock" ] || continue
+  CARGO_MANIFESTS+=(--sync "${BUILDSYS_ROOT_DIR}/${ws}/Cargo.toml")
 done
+cargo vendor --locked --no-delete  --versioned-dirs --respect-source-config  ${CARGO_MANIFESTS[@]} .cargo/vendor > .cargo/config.toml 2> .cargo/vendor.log || {
+   echo "vendoring of cargo dependencies failed"
+   cat .cargo/vendor.log
+   exit 1
+}
+rm .cargo/vendor.log
 
 chmod -R o+r ${CARGO_HOME}
 '''
--- a/twoliter/embedded/build.Dockerfile
+++ b/twoliter/embedded/build.Dockerfile
@@ -109,7 +109,11 @@ RUN --mount=target=/host \
     rm /bypass
 
 USER builder
-RUN --mount=source=.cargo,target=/home/builder/.cargo \
+    # We only mount the config.toml and the vendor directory instead of the whole .cargo directory
+    # because the .git and .registry directories are specific to the version of cargo and can cause
+    # incapatibility between the host cargo and the cargo in the sdk
+RUN --mount=source=.cargo/config.toml,target=/home/builder/rpmbuild/.cargo/config.toml \
+    --mount=source=.cargo/vendor,target=/home/builder/rpmbuild/.cargo/vendor \
     --mount=type=cache,target=/home/builder/.cache,from=cache,source=/cache \
     --mount=source=sources,target=/home/builder/rpmbuild/BUILD/sources \
     --mount=target=/host \


twoliter/embedded/Makefile.toml
```diff
@@ -354,11 +354,25 @@ dependencies = ["setup"]
...
-
+DIGEST="$(echo ${CARGO_DIGESTS[@]} | sha256sum | awk '{ print $1 }')"
+MARKER_FILE=".cargo/vendor-${DIGEST}.marker"
+if [ -f "${MARKER_FILE}" ]; then
+    echo "no changes detected in cargo projects, skipping cargo vendor"
+else
+    cargo vendor --locked ${CARGO_MANIFESTS[@]} .cargo/vendor > .cargo/config.toml 2> .cargo/vendor.log || {
+        echo "vendoring cargo sources failed"
+        cat .cargo/vendor.log
+    }
```
cbgbt: We don't pass `--no-delete` to `cargo vendor` but we also don't clean up older marker files. I think we have to do one or the other.

In order to keep old marker files, I think we have to pass `--versioned-dirs` to `cargo vendor` so that we can keep multiple different versions of vendored deps based on the Bottlerocket version. But I'm not sure! Any chance you could test by doing `git checkouts` on workspaces with different versions of deps in `Cargo.lock` and see what happens?

Additionally, can we pass `--respect-source-config` here? I'm somewhat interested in using kits to ship cargo registries so that they can ship library dependencies. This idea isn't one that I've scoped or has much buy-in yet, but I'd like to keep it available as an option if we can.

cbgbt (commented):
I'm happy with this approach, nice work!

Though I have some concerns about what happens to previously cached data since we aren't cleaning up old marker files. Specifically, I'm wondering if it's possible for us to get into cases where we're accidentally using the wrong version of a dependency -- seems unlikely, since the Cargo metadata in the versioned dep should be different.

twoliter/embedded/Makefile.toml
```diff
@@ -354,11 +354,29 @@ dependencies = ["setup"]
...
+declare -a CARGO_DIGESTS
+mkdir -p .cargo/vendor
+if [ ! -f ".cargo/config.toml" ]; then
+    touch .cargo/config.toml
+fi
 for ws in sources variants .; do
   [ -s "${BUILDSYS_ROOT_DIR}/${ws}/Cargo.toml" ] || continue
-  cargo fetch --locked --manifest-path "${BUILDSYS_ROOT_DIR}/${ws}/Cargo.toml"
+  CARGO_MANIFESTS+=(--sync "${BUILDSYS_ROOT_DIR}/${ws}/Cargo.toml")
+  CARGO_DIGESTS+=("$(sha256sum "${BUILDSYS_ROOT_DIR}/${ws}/Cargo.lock" | awk '{ print $1 }')")
```
cbgbt: I wonder what happens if `Cargo.lock` doesn't exist? I think this is possible if a user does not have different workspaces for each of these directories.

As we discussed we can just use `Cargo.lock` as the marker file instead of `Cargo.toml` at the start of the loop to make it work.

twoliter/embedded/Makefile.toml
```diff
@@ -354,11 +354,29 @@ dependencies = ["setup"]
...
 done
-
+DIGEST="$(echo ${CARGO_DIGESTS[@]} | sha256sum | awk '{ print $1 }')"
+MARKER_FILE=".cargo/vendor-${DIGEST}.marker"
+if [ -f "${MARKER_FILE}" ]; then
+    echo "no changes detected in cargo projects, skipping cargo vendor"
+else
+    cargo vendor --versioned-dirs --respect-source-config --locked ${CARGO_MANIFESTS[@]} .cargo/vendor > .cargo/config.toml 2> .cargo/vendor.log || {
```
cbgbt: Can you check that we don't clobber settings in `.cargo/config.toml`?

You mentioned in person that `--respect-source-config` probably does this and I think you're right, but we should test it.

cbgbt (commented):
Sorry, I should have thought about these the first time I reviewed!

twoliter/embedded/Makefile.toml
bcressey: If the marker and log file logic aren't essential, I'd prefer to omit them and stay closer to the previous implementation:
```
- cargo fetch --locked --manifest-path "${BUILDSYS_ROOT_DIR}/${ws}/Cargo.toml"
+ cargo vendor --locked --sync "${BUILDSYS_ROOT_DIR}/${ws}/Cargo.toml" --no-delete  --versioned-dirs --respect-source-config  --quiet .cargo/vendor
```

twoliter/embedded/Makefile.toml
```diff
@@ -354,11 +354,29 @@ dependencies = ["setup"]
 script_runner = "bash"
 script = [
 '''
+declare -a CARGO_MANIFESTS
+declare -a CARGO_DIGESTS
+mkdir -p .cargo/vendor
+if [ ! -f ".cargo/config.toml" ]; then
+    touch .cargo/config.toml
+fi
```
bcressey: This doesn't seem necessary - `cargo vendor` will create these directories and doesn't complain if the config file is missing.

twoliter/embedded/build.Dockerfile
```diff
@@ -109,7 +109,8 @@ RUN --mount=target=/host \
     rm /bypass
 
 USER builder
-RUN --mount=source=.cargo,target=/home/builder/.cargo \
+RUN --mount=source=.cargo/config.toml,target=/home/builder/rpmbuild/.cargo/config.toml \
+    --mount=source=.cargo/vendor,target=/home/builder/rpmbuild/.cargo/vendor \
```
bcressey, line 116: Why not mount the whole `.cargo` directory? Is there content in there that we're trying to keep the SDK copy of `cargo` from seeing?

jmt-lab: If we mount the whole .cargo directory we'll be back to square one with the error we are hitting. the .registry folder takes precedence over vendor

twoliter/embedded/Makefile.toml
jmt-lab: this is because of the marker file

twoliter/embedded/Makefile.toml
jmt-lab: The markerfile was done to prevent running cargo vendor every time you call twoliter build commands. It was meant to optimize timing as when you cargo vendor it does the whole shebam over.

twoliter/embedded/build.Dockerfile
jmt-lab: and also the internal folders in the .cargo are the cargo version specific folders and files. I chose to put vendor inside .cargo root for cleanliness and to prevent .gitignore clashes.

twoliter/embedded/Makefile.toml
jmt-lab: Removed the marker file

twoliter/embedded/Makefile.toml
jmt-lab: removed

cbgbt (approved):

gthao313 (approved):

twoliter/embedded/build.Dockerfile
bcressey: > If we mount the whole .cargo directory we'll be back to square one with the error we are hitting. the .registry folder takes precedence over vendor

Can you add a comment to this effect? It's valuable context to avoid an otherwise tempting refactor to simplify down to the one mount.

bcressey (approved):

PR #471, opened 2025-03-03
Update rust toolchain to nightly-2025-02-28

cbgbt:
**Description of changes:**
* Upgrade Rust's nightly compiler. This allows us to build against the latest `cargo-deny`.
* Modify CI to link against `musl` to avoid ABI incompatibilities with Bottlerocket's SDK

**Testing done:**
* CI on this commit.

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #473, opened 2025-03-05
fix: add minutes to parse-datetime for waves

### LINKED ISSUE #472: build/tools/waves/ohno.toml does not work (uses minutes in time, but is unsupported by `parse-datetime`)
We recently tried the `ohno.toml` wave timing, and discovered that it doesn't work because it uses minutes here: https://github.com/bottlerocket-os/twoliter/blob/develop/twoliter/embedded/waves/ohno.toml#L4
but it is unsupported here: https://github.com/bottlerocket-os/twoliter/blob/develop/tools/parse-datetime/src/lib.rs#L99

I will attach a PR that adds minutes also to this match statement.

mikn:
**Issue number:**

Closes #472

**Description of changes:**
Introduces minutes also to parse-datetime to allow for the `ohno.toml` to work.

**Testing done:**
Tried building a repository using the ohno.toml wave configuration.

**Terms of contribution:**

### DIFF
--- a/tools/parse-datetime/src/lib.rs
+++ b/tools/parse-datetime/src/lib.rs
@@ -97,6 +97,10 @@ pub fn parse_offset(input: &str) -> Result<Duration> {
         .context(error::DateArgCountSnafu { input })?;
 
     let duration = match unit_str {
+        "minute" | "minutes" => Duration::try_minutes(i64::from(count)).context(error::DateIntSnafu {
+            integer: count,
+            unit: "minutes",
+        })?,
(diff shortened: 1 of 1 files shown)


cbgbt (approved):
Looks good, thanks for submitting this fix!

cbgbt:
~~The CI is failing due to an unrelated issue that's fixed in #471~~

Ah, I just assumed this was due to CI failing on the repository overall, but there's actually a formatting issue. @mikn, do you mind running `cargo fmt`?

bcressey (approved):
Thanks!

cbgbt (approved):

PR #484, opened 2025-03-11
testsys: support custom block mappings for ec2 instance launch

gthao313:
**Issue number:**

Closes #

**Description of changes:**
 testsys: support custom block mappings for ec2 instance launch
Associated PR at TestSys: https://github.com/bottlerocket-os/bottlerocket-test-system/pull/978

**Testing done:**
https://github.com/bottlerocket-os/bottlerocket-test-system/pull/978#issue-2908692952

**Terms of contribution:**

### DIFF
--- a/tools/testsys/src/aws_resources.rs
+++ b/tools/testsys/src/aws_resources.rs
@@ -3,7 +3,7 @@ use crate::error::{self, Result};
 use aws_sdk_ec2::config::Region;
 use aws_sdk_ec2::types::{Filter, Image};
 use bottlerocket_types::agent_config::{
-    ClusterType, CustomUserData, Ec2Config, Ec2KarpenterConfig, KarpenterDeviceMapping,
+    BlockDeviceMappingConfig, ClusterType, CustomUserData, Ec2Config, Ec2KarpenterConfig,
 };
 use maplit::btreemap;
 use serde::Deserialize;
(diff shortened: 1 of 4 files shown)


tools/testsys/src/aws_resources.rs
```diff
@@ -191,6 +180,13 @@ pub(crate) async fn ec2_crd(
                 .testsys_agent_pull_secret
                 .clone(),
         )
+        .device_mappings(Some(
+            bottlerocket_input
+                .crd_input
+                .config
+                .block_device_mapping
+                .clone(),
+        ))
```
yeazelm: This is safe if the config is not set? Does this just result in `Some()` of an empty Vec?

tools/testsys/src/aws_resources.rs
```diff
@@ -191,6 +180,18 @@ pub(crate) async fn ec2_crd(
                 .testsys_agent_pull_secret
                 .clone(),
         )
+        .device_mappings({
```
gthao313, line 183: `device_mappings` is the field of EC2Config struct [here](https://github.com/bottlerocket-os/bottlerocket-test-system/blob/56aecbaf0d2c6229d035fc02af7ce5701abdc940/bottlerocket/types/src/agent_config.rs#L298), so I have to use `device_mappings` instead of `set_device_mappings`.

```
ec2_builder
145 | |         .node_ami(bottlerocket_input.image_id)
146 | |         .instance_count(2)
147 | |         .instance_types::<Vec<String>>(
...   |
183 | |         .set_device_mappings({
    | |_________-^^^^^^^^^^^^^^^^^^^
    |
help: there is a method `device_mappings` with a similar name
    |
183 -         .set_device_mappings({
183 +         .device_mappings({
```

jpculp (changes requested):
Can we swap the order of the commits or squash them? Best to avoid a commit that doesn't work on its own (if we reverted the testsys bump for example).

jpculp (approved):

jpculp (approved):

yeazelm (approved):

yeazelm:
New commit message looks good!

yeazelm (approved):

PR #487, opened 2025-03-12
Adjust use of cargo config file to not clash with host cargo operations

### LINKED ISSUE #486: New .cargo format forces me to remove .cargo to run `cargo generate-lockfile` in sources/
I recently tried to upgrade to Twoliter 0.8.0 where you changed the format of how you vendor dependencies into the .cargo folder to make it more robust against version changes. Inherently great! But sadly if you've built a kit or a variant that has resulted in a `.cargo` folder this gets picked up by cargo (I am unsure how to disable it):
https://internals.rust-lang.org/t/problems-of-cargo-config-files-and-possible-solutions/12987/14
https://github.com/rust-lang/cargo/issues/12738

And for me it creates a recursive loop that gives an error like this:
```
mikn@hal8999 ~/d/m/s/p/b/b/sources ((72abfc6e))> cargo generate-lockfile
error: failed to get `bottlerocket-defaults-helper` as a dependency of package `settings-defaults-metal-dev v0.3.0 (/home/mikn/devel/molnett/src/platform/base/bottlerocket/sources/settings-defaults/metal-dev)`

Caused by:
  failed to load source for dependency `bottlerocket-defaults-helper`

Caused by:
  Unable to update https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-defaults-helper-v0.1.1

Caused by:
  the source git+https://github.com/bottlerocket-os/bottlerocket-settings-sdk?tag=bottlerocket-defaults-helper-v0.1.1 requires a lock file to be present first before it can be
  used against vendored source code

  remove the source replacement configuration, generate a lock file, and then
  restore the source replacement configuration to continue the build
mikn@hal8999 ~/d/m/s/p/b/b/sources ((72abfc6e)) [101]> 
```

I can fix this by removing the .cargo folder, but I do need to do that every time which feels a bit crufty.

The secondary problem is that it is also a lot more sensitive to races, as now it seems to expect to be allowed to do more in this folder than it was previously. For now I am downgrading to 0.7.3 again.

jmt-lab:
**Issue number:** 486

Closes #486 

**Description of changes:**

* Change our call to cargo vendor to store the config file to .cargo/twoliter_cargo_config.toml
* Checks if the user has their own .cargo/config.toml and if so will ensure it is taken into account
* Modify the bind mount in the build context to mount .cargo/twoliter_cargo_config.toml to .cargo/config.toml 

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/Makefile.toml
+++ b/twoliter/embedded/Makefile.toml
@@ -362,13 +362,31 @@ for ws in sources variants .; do
   [ -s "${BUILDSYS_ROOT_DIR}/${ws}/Cargo.lock" ] || continue
   CARGO_MANIFESTS+=(--sync "${BUILDSYS_ROOT_DIR}/${ws}/Cargo.toml")
 done
-cargo vendor --locked --no-delete  --versioned-dirs --respect-source-config  ${CARGO_MANIFESTS[@]} .cargo/vendor > .cargo/config.toml 2> .cargo/vendor.log || {
-   echo "vendoring of cargo dependencies failed"
-   cat .cargo/vendor.log
-   exit 1
-}
-rm .cargo/vendor.log
+exec 9<>.cargo/vendor.lock
+if ! flock -w 90 9; then
+    echo "failed to obtain lock" >&2
+    exit 1
+fi
 
+set +e
+cargo vendor \
+    --locked \
+    --no-delete \
+    --versioned-dirs \
+    --respect-source-config \
+    ${CARGO_MANIFESTS[@]} \
+    .cargo/vendor \
+    > .cargo/twoliter_cargo_config.toml \
+    2> .cargo/vendor.log
+
+if [ "${?}" -ne 0 ]; then
+   echo "vendoring of cargo dependencies failed" >&2
+   cat .cargo/vendor.log >&2
+   exit 1
+fi
+
+set -e
+rm .cargo/vendor.log
 chmod -R o+r ${CARGO_HOME}
 '''
 ]
@@ -459,7 +477,8 @@ done
 if ! docker run --rm \
    -u $(id -u):$(id -g) \
    -e CARGO_HOME="/tmp/.cargo" \
-   -v "${CARGO_HOME}":/tmp/.cargo \
+   -v "${CARGO_HOME}/twoliter_cargo_config.toml":/tmp/.cargo/config.toml \
+   -v "${CARGO_HOME}/vendor":/tmp/.cargo/vendor \
    -v "${BUILDSYS_ROOT_DIR}/sources":/tmp/sources \
    "${TLPRIVATE_SDK_IMAGE}" \
    cargo fmt \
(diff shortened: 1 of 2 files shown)


qianxjcraig (approved):

twoliter/embedded/Makefile.toml
```diff
@@ -23,11 +23,17 @@ BUILDSYS_SBKEYS_DIR = "${BUILDSYS_ROOT_DIR}/sbkeys"
 BUILDSYS_CACERTS_BUNDLE_OVERRIDE = ""
 BUILDSYS_METADATA_DIR = "${BUILDSYS_BUILD_DIR}/metadata"
 BUILDSYS_CARGO_METADATA_PATH = "${BUILDSYS_METADATA_DIR}/cargo_metadata.json"
-BUILDSYS_SBKEYS_PROFILE = { script = ['echo "${BUILDSYS_SBKEYS_PROFILE:-local}"'] }
-BUILDSYS_VERSION_BUILD = { script = ["git describe --always --dirty --exclude '*' --abbrev=8 || echo 00000000"] }
+BUILDSYS_SBKEYS_PROFILE = { script = [
+    'echo "${BUILDSYS_SBKEYS_PROFILE:-local}"',
+] }
```
jmt-lab: Uuuuuuugh my editor did nonsense reformatting hold on reverting.

twoliter/embedded/Makefile.toml
```diff
@@ -362,7 +362,15 @@ for ws in sources variants .; do
...
-cargo vendor --locked --no-delete  --versioned-dirs --respect-source-config  ${CARGO_MANIFESTS[@]} .cargo/vendor > .cargo/config.toml 2> .cargo/vendor.log || {
+if [ -f ".cargo/config.toml" ]; then
+  # if the user has a configuration we should respect it
+  cp ".cargo/config.toml" ".cargo/twoliter_cargo_config.toml"
+elif [ -f ".cargo/twoliter_cargo_config.toml" ]; then
+  # If this file already exists from a previous run remove it to avoid duplication
+  # of vendor configuration
+  rm .cargo/twoliter_cargo_config.toml
+fi
+cargo vendor --locked --no-delete  --versioned-dirs --respect-source-config  ${CARGO_MANIFESTS[@]} .cargo/vendor >> .cargo/twoliter_cargo_config.toml 2> .cargo/vendor.log || {
```
bcressey: This restores functionality from the original PR that I didn't understand the motivation for. What do we mean by "we should respect it"?

In what way are we respecting it? How do we know that what we're respecting won't conflict with what we need to happen here as far as vendoring crates?

It might be better to ignore the edits made by `cargo vendor` to the config and just generate our own configuration with the expected paths whenever we want to use the vendored directory.

But then I have questions around how we can make sure the various `Makefile.toml` tasks use the vendored crates instead of downloading them all again. And how we use those crates when running cargo commands directly under `sources`.

jmt-lab: I mostly just add it by default, in cause an external customer wanted to use some other argument in .cargo/config.toml such as [net]git.fetch_with_cli or something

jmt-lab:
Updated to use the flock approach to get a file lock before vendor to prevent multiple runs of cargo vendor colliding with each other.

bcressey:
Can you fix the commit message to align with conventions?
```
component: short summary of fix

Explanation of issue, 1-2 sentences.

Explanation of fix.

Explanation of any secondary changes e.g. the introduction of locking.
```

bcressey (approved):

PR #509, opened 2025-03-18
feat: add single zone configuration

zongzhidanielhu:
**Issue number:**

**Description of changes:**
Added feature allows configuring a specific zone in Test.toml, so that EC2 instance(s) can be launched in this specific zone. Usually, two zones must specified, but this feature allows one specific single zone for EC2 instance to launch into.

**Testing done:**
See testing at: https://github.com/bottlerocket-os/bottlerocket-test-system/pull/984

**Terms of contribution:**

### DIFF
--- a/tools/testsys/src/run.rs
+++ b/tools/testsys/src/run.rs
@@ -151,6 +151,11 @@ struct CliConfig {
     #[arg(long, env = "TESTSYS_INSTANCE_TYPE")]
     instance_type: Option<String>,
 
+    /// Specify the zone that should be used. This is only applicable for aws-* variants.
+    /// It can be omitted for non-aws variants and can be omitted to use default instance types.
+    #[arg(long, env = "TESTSYS_ZONES")]
+    zones: Option<Vec<String>>,
+
     /// Add secrets to the testsys agents (`--secret awsCredentials=my-secret`)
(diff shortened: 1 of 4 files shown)


tools/testsys/src/run.rs
```diff
@@ -151,6 +151,11 @@ struct CliConfig {
     #[arg(long, env = "TESTSYS_INSTANCE_TYPE")]
     instance_type: Option<String>,
 
+    /// Specify the zone that should be used. This is only applicable for aws-* variants.
+    /// It can be omitted for non-aws variants and can be omitted to use default instance types.
```
yeazelm, line 155:
```suggestion
    /// It can be omitted for non-aws variants and can be omitted to use default availability zones for the vpc.
```
Is this what you meant?

tools/testsys/src/run.rs
```diff
@@ -151,6 +151,11 @@ struct CliConfig {
...
+    #[arg(long, env = "TESTSYS_ZONES")]
```
yeazelm, line 156: Is the intention that we would specify this as TESTSYS_ZONES? I figured this would just be availability zones specific? Can you share how this is intended to be used as an example in the description?

PR #511, opened 2025-04-01
Use Rust's front-end parallel compilation

arnaldo2792:
**Description of changes:**

Use the recommended configuration that enables the front-end parallel compilation for faster builds. In various runs, I observed that the compilation time went down by 30 seconds:

```
# Before
Finished `release` profile [optimized] target(s) in 8m 22s

# After
Finished `release` profile [optimized] target(s) in 7m 40s
```

**Testing done:**

- Build variant, and publish the AMI successfully

**Terms of contribution:**

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,8 @@
 [unstable]
 bindeps = true
+
+[build]
+# This is the recommended configuration that yields the
+# best results:
+#  - https://blog.rust-lang.org/2023/11/09/parallel-rustc.html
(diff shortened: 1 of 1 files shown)


cbgbt (approved):
I'm supportive of enabling this in twoliter, but we should be mindful that the compiler team is still working through a backlog of ICE and deadlock bugs:
https://github.com/rust-lang/rust/labels/WG-compiler-parallel

If we encounter such issues while compiling twoliter, we should revert.

bcressey (approved):

PR #512, opened 2025-04-02
twoliter: add support for kmod kit feature flag

### LINKED ISSUE #74: kernel: Chart out future for kmod-kit

rpkelly:
**Issue number:**

Closes: https://github.com/bottlerocket-os/bottlerocket-kernel-kit/issues/74

**Description of changes:**
Added a feature flag `kmod-kits` which is on by default. Turning this feature flag off, causes builds to not produce kmod kits.

**Testing done:**
Built `aws-ecs-2` variant using current variant definition, saw that kmod kits were created. Added `kmod-kits = false` to the variant definition, rebuilt, and saw that no kmod kits were created. When no kmod kits created, the `.no-kmod-kit` marker file was created. Tested that `cargo make repo` continued to work in this case.

Set `in-place-updates = false` saw that migrations files were not created. Additionally a `.no-migrations` marker is created in order to allow `cargo make repo` to continue to work.

**Terms of contribution:**

### DIFF
--- a/tools/buildsys/src/manifest.rs
+++ b/tools/buildsys/src/manifest.rs
@@ -239,6 +239,14 @@ FIPS-compliant ciphers to be included in the image.
 fips = true
 ```
 
+`external-kmod-development` enables functionality for building custom kernel modules at runtime
+
+```ignore
+[package.metadata.build-variant.image-features]
+external-kmod-development = true
+```
+
+
 */
 
 mod error;
@@ -487,8 +495,11 @@ impl ManifestInfo {
     /// Convenience method to return the enabled image features for this variant.
     pub fn image_features(&self) -> Option<HashSet<ImageFeature>> {
         let variant = self.build_variant()?;
-        let mut features =
-            HashSet::from([ImageFeature::InPlaceUpdates, ImageFeature::HostContainers]);
+        let mut features = HashSet::from([
+            ImageFeature::InPlaceUpdates,
+            ImageFeature::HostContainers,
+            ImageFeature::ExternalKmodDevelopment,
+        ]);
         if let Some(image_features) = &variant.image_features {
             for (feature, enabled) in image_features.iter() {
                 if *enabled {
@@ -800,6 +811,7 @@ pub enum ImageFeature {
     Fips,
     InPlaceUpdates,
     HostContainers,
+    ExternalKmodDevelopment,
 }
 
 const EXPERIMENTAL_IMAGE_FEATURES: [&ImageFeature; 1] = [&ImageFeature::ErofsRootPartition];
@@ -816,6 +828,7 @@ impl TryFrom<String> for ImageFeature {
             "fips" => Ok(ImageFeature::Fips),
             "in-place-updates" => Ok(ImageFeature::InPlaceUpdates),
             "host-containers" => Ok(ImageFeature::HostContainers),
+            "external-kmod-development" => Ok(ImageFeature::ExternalKmodDevelopment),
             _ => error::ParseImageFeatureSnafu { what: s }.fail()?,
         }
     }
@@ -832,6 +845,7 @@ impl fmt::Display for ImageFeature {
             ImageFeature::Fips => write!(f, "FIPS"),
             ImageFeature::InPlaceUpdates => write!(f, "IN_PLACE_UPDATES"),
             ImageFeature::HostContainers => write!(f, "HOST_CONTAINERS"),
+            ImageFeature::ExternalKmodDevelopment => write!(f, "EXTERNAL_KMOD_DEVELOPMENT"),
         }
     }
 }
--- a/twoliter/embedded/build.Dockerfile
+++ b/twoliter/embedded/build.Dockerfile
@@ -212,6 +212,7 @@ ARG EROFS_ROOT_PARTITION
 ARG IN_PLACE_UPDATES
 ARG HOST_CONTAINERS
 ARG FIPS
+ARG EXTERNAL_KMOD_DEVELOPMENT
 
 USER builder
 WORKDIR /home/builder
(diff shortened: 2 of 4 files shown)


tools/buildsys/src/manifest.rs
```diff
@@ -239,6 +239,15 @@ FIPS-compliant ciphers to be included in the image.
 fips = true
 ```
 
+`kmod-kits` enables a feature for users to bring their own custom kernel modules other than the
+modules pre-built into the Bottlerocket kernel.
```
arnaldo2792: Technically, the kernel modules aren't all built into the kernel

```suggestion
`kmod-kits` enables a feature for users to bring their own custom kernel modules other than the
modules provided by Bottlerocket.
```

tools/buildsys/src/manifest.rs
```diff
@@ -487,8 +496,11 @@ impl ManifestInfo {
     /// Convenience method to return the enabled image features for this variant.
     pub fn image_features(&self) -> Option<HashSet<ImageFeature>> {
         let variant = self.build_variant()?;
-        let mut features =
-            HashSet::from([ImageFeature::InPlaceUpdates, ImageFeature::HostContainers]);
+        let mut features = HashSet::from([
+            ImageFeature::InPlaceUpdates,
+            ImageFeature::HostContainers,
+            ImageFeature::KmodKits,
```
arnaldo2792: I see, so this is what makes the feature be on by default.

tools/buildsys/src/manifest.rs
```diff
@@ -239,6 +239,15 @@ FIPS-compliant ciphers to be included in the image.
 fips = true
 ```
 
+`kmod-kits` enables a feature for users to bring their own custom kernel modules other than the
+modules provided by Bottlerocket.
+
+```ignore
+[package.metadata.build-variant.image-features]
+kmod-kits = true
+```
```
bcressey: I'd like to give this a somewhat more generic name since it will also be used to disable the inclusion of the kernel devel archive at runtime.

```suggestion
`kmod-kits` enables a feature for users to bring their own custom kernel modules other than the
modules provided by Bottlerocket.

```ignore
[package.metadata.build-variant.image-features]
external-kmod-development = true
```
```

File: twoliter/embedded/build.Dockerfile [hunk 4]
```diff
@@ -436,11 +438,13 @@ ENV VARIANT=${VARIANT} VERSION_ID=${VERSION_ID} BUILD_ID=${BUILD_ID}
 ARG BYPASS_SOCKET
 ARG OUTPUT_SOCKET
 ARG BUILDER_UID
+ARG KMOD_KITS
 
 USER root
 
 WORKDIR /tmp
 RUN --mount=target=/host \
+    if [[ -n "${KMOD_KITS}" ]]; then \
```
bcressey: See usage of `BUILDSYS_KMOD_KIT` in `Makefile.toml` - if we don't produce a kmod kit then we will need to conditionalize the inclusion in the `repo` task.

twoliter/embedded/build.Dockerfile
```diff
@@ -436,11 +438,13 @@ ENV VARIANT=${VARIANT} VERSION_ID=${VERSION_ID} BUILD_ID=${BUILD_ID}
 ARG BYPASS_SOCKET
 ARG OUTPUT_SOCKET
 ARG BUILDER_UID
+ARG KMOD_KITS
```
bcressey: While we're making this change, it'd be great to also suppress `rpm2migrations` if the in-place updates feature isn't active.

tools/buildsys/src/manifest.rs
```diff
@@ -239,6 +239,15 @@ FIPS-compliant ciphers to be included in the image.
 fips = true
 ```
 
+`external-kmod-develpment` enables a feature for users to bring their own custom kernel modules other than the
+modules provided by Bottlerocket.
```
bcressey:
```suggestion
`external-kmod-development` enables functionality for building custom kernel modules at runtime
```

tools/buildsys/src/manifest.rs
```diff
@@ -800,6 +812,7 @@ pub enum ImageFeature {
     Fips,
     InPlaceUpdates,
     HostContainers,
+    ExternalKmodDevelpment,
```
bcressey:
```suggestion
    ExternalKmodDevelopment,
```

twoliter/embedded/build.Dockerfile
```diff
@@ -401,21 +403,26 @@ ARG VARIANT
...
 ENV VARIANT=${VARIANT} VERSION_ID=${VERSION_ID} BUILD_ID=${BUILD_ID}
 WORKDIR /root
 
 USER root
 RUN --mount=target=/host \
     /host/build/tools/pipesys link --fd-socket "${BYPASS_SOCKET}" --target /bypass && \
     /host/build/tools/pipesys link --fd-socket "${OUTPUT_SOCKET}" --target /output && \
-    mkdir -p /local/migrations && \
+    mkdir -p /local/migrations; \
+    if [[ -n "${IN_PLACE_UPDATES}" ]]; then \
```
bcressey, line 415: nit: it'll make the diff annoying, but let's go ahead and indent the `if` and `else` block contents for readability. Especially the `else` clause suffers from trying to avoid it.

twoliter/embedded/metadata.spec
```diff
@@ -62,7 +62,13 @@ Provides: %{_cross_os}image-feature(fips)
 Provides: %{_cross_os}image-feature(no-fips)
 %endif
 
+%if %{with kmod_kits}
+Provides: %{_cross_os}image-feature(kmod-kits)
+%else
+Provides: %{_cross_os}image-feature(no-kmod-kits)
+%endif
```
bcressey:
```suggestion
%if %{with external_kmod_development}
Provides: %{_cross_os}image-feature(external-kmod-development)
%else
Provides: %{_cross_os}image-feature(no-external-kmod-development)
%endif
```

twoliter/embedded/metadata.spec
```diff
@@ -62,7 +62,13 @@ Provides: %{_cross_os}image-feature(fips)
...
 %description
+
```
bcressey: nit: begone, new newline!
```suggestion
```

twoliter/embedded/metadata.spec
```diff
@@ -62,6 +62,11 @@ Provides: %{_cross_os}image-feature(fips)
 Provides: %{_cross_os}image-feature(no-fips)
 %endif
 
+%if %{with external_kmod_development}
+Provides: %{_cross_os}image-feature(external-kmod-development)
+%else
+Provides: %{_cross_os}image-feature(external-kmod-development)
```
bcressey:
```suggestion
Provides: %{_cross_os}image-feature(no-external-kmod-development)
```

bcressey (commented):
LGTM, just one fix.

bcressey (approved):

arnaldo2792 (approved):

PR #521, opened 2025-04-22
amispec: initial implementation of amispec

cbgbt:
**Issue number:**

Related to https://github.com/bottlerocket-os/bottlerocket/issues/4000

**Description of changes:**
This PR introduces a set of libraries that will be used to allow `pubsys` to create AMIs with configurable AMI properties and EBS block-device-mappings.

While not implemented here, these libraries will allow users to put an `amispec.toml` file next to their variant definition, which will allow users to override default attributes of the AMI (like volume-type, volume-size, tpm-support, etc.)

---
    serde-templated: create serde-templated

    This crate provides primitives for deserializing data types which may be
    handlebars templates that require rendering before finally being parsed
    into the target data structure.
---

    serde-templated-derive: provide macros for making templated types

    This macro provides the `Templated` derive macro, which simplifies the
    process of creating a "template-able" version of a given struct.
---

    amispec: provide initial implementation of amispec

    amispec is a library for parsing AMI properties, including EBS block
    device mappings. The library provides a mechanism for reading
    templatized configurations via serde.
---

**Note to Reviewers:**
This PR is best reviewed commit-by-commit, as the `amispec` library depends on the two preceding libraries in the commit chain.

Additionally, the following AWS documentation pages are helpful for reviewing the exposed features of the AWS APIs:
* https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RegisterImage.html
* https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_BlockDeviceMapping.html
* https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_EbsBlockDevice.html

**Testing done:**
* All provided unit tests pass
* I have pending changes to `pubsys` which take advantage of these changes to register AMIs. I've confirmed that they successfully apply the desired AMI properties.

**Terms of contribution:**


cbgbt:
^ force push to rebase

tools/serde-templated/src/templated.rs
```diff
@@ -0,0 +1,133 @@
...
+impl<T> From<T> for Templated<T>
+where
+    T: FromStr + Clone,
+{
+    fn from(value: T) -> Self {
+        Self::raw(value)
+    }
+}
+
+impl From<&str> for Templated<String> {
```
sam-berning: why implement this just for `Templated<String>` if you're casting it as `Templated::Template`? this should be generalizable to `Templated<T>`

jmt-lab:
For knowledge sharing purposes, why have we introduced a new toml file here versus adding to our existing use of cargo metadata in variant definitions or adding new logic to Twoliter.toml?

tools/serde-templated-derive/src/templated.rs
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+    /// Defines the `TemplateOf` trait for the resulting `Templated` type.
+    pub(crate) fn render_template_of_trait_def(&self) -> Result<proc_macro2::TokenStream> {
+        let Self {
+            ident,
+            generics,
+            data: _data,
+            ..
+        } = self;
```
sam-berning: this is neat

tools/serde-templated-derive/src/templated.rs
```diff
@@ -0,0 +1,374 @@
...
+
+        let generated_struct_name = format_ident!("Templated{}", ident);
```
sam-berning: nit: I'd prefer this in a function since we use this in multiple places

cbgbt:
> For knowledge sharing purposes, why have we introduced a new toml file here versus adding to our existing use of cargo metadata in variant definitions or adding new logic to Twoliter.toml?

Good question! We already do something similar to this for vmware variants, [e.g.](https://github.com/bottlerocket-os/bottlerocket/tree/develop/variants/vmware-k8s-1.32). This lets us segment the platform-specific registration details from the variant itself.

Imagine a variant which had the userspace components to run on multiple cloud platforms -- this setup allows managing details of registration in each cloud platform without modifying the identity of the variant itself.

tools/serde-templated-derive/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{DeriveInput, parse_macro_input};
+
+mod templated;
+
+pub(crate) type Error = darling::Error;
+pub(crate) type Result<T, E = crate::Error> = std::result::Result<T, E>;
+
+#[proc_macro_derive(Templated, attributes(templated))]
```
sam-berning, line 13: nit: the tests provide a decent overview of the usage of this macro, but a doc comment here or a README for this crate would be useful

jmt-lab:
Another question, I do not understand why we need to have this serde-templated stuff just for handlebars replacement. Why not simply pre-resolve the handlebar template, as in when we read amispec.toml before we deserialize we run the handlebar tempalte substitutions then afterwards deserialize the resolved string? That seems like the simpler approach than adding all this serde templated stuff. But let me know if i'm missing something here.

Also if the reason is to defer resolution of the handlebar templates to later, this seems like a lot of code for deferring at a finer grain than necessary. Why not just defer loading the file till when we have everything we need to resolve the template so the flow could be read file -> handlebar resolve -> serde standard deserialize?

cbgbt:
> Another question, I do not understand why we need to have this serde-templated stuff just for handlebars replacement. Why not simply pre-resolve the handlebar template, as in when we read amispec.toml before we deserialize we run the handlebar tempalte substitutions then afterwards deserialize the resolved string?
> ...
> Why not just defer loading the file till when we have everything we need to resolve the template so the flow could be read file -> handlebar resolve -> serde standard deserialize?

I tend to dislike rendering templated configuration files and then parsing them. The problem is that without careful template design, variables in the render context can inject strings that interact with the underlying serialization format.

Consider a template like this:

```toml
hello = "world! and {{ someone }}"
```

Suppose `someone` is set to something like `"\"\nmalicious = \"mwahahaha\""`. Whoops, injection!

Template designers can protect against this with handlebars helpers to encode things, e.g. `{{ toml_encode someone }}` but it's easy to make a mistake.

> Also if the reason is to defer resolution of the handlebar templates to later, this seems like a lot of code for deferring at a finer grain than necessary

I appreciate that there's added complexity in the implementation and am open to that feedback. This solution does defer rendering until after parsing, but the purpose is to avoid the injection issues I mentioned above and not for the sake of the deferred rendering itself.

tools/amispec/src/ebs.rs
```diff
@@ -0,0 +1,585 @@
...
+                mut $builder: EbsBlockDeviceBuilder,
+            ) -> EbsBlockDeviceBuilder {
+                $builder = $builder
+                    .set_delete_on_termination($self.delete_on_termination)
+                    .set_encrypted($self.encrypted)
+                    .set_kms_key_id($self.kms_key_id.clone())
+                    .set_outpost_arn($self.outpost_arn.clone())
+                    .set_snapshot_id($self.snapshot_id.clone());
+                $($impl_body)*
+            }
```
sam-berning: nit: I think there's a slightly confusing bug here if a user provides an empty `build()` in the `macro_rules` pattern. I think we should either add a return after the `impl_body` lines here, or require `build()` to be at least one line with `$($impl_body:tt)+`. I think I prefer the latter.

I tested this by adding a dummy `Gp4` volume type with the following:

```
impl_ebs_volume_type! {
    Gp4 ("gp4") {
        template: TemplatedGp4,
        volume_size: Gp4VolumeSize,

    build(&self, builder) {}
}
````

error with `$($impl_body:tt)*`:
```
error[E0308]: mismatched types
   --> tools/amispec/src/ebs.rs:192:18
    |
189 |               fn configure_ebs_block_device(
    |                  -------------------------- implicitly returns `()` as its body has no tail or `return` expression
...
192 |               ) -> EbsBlockDeviceBuilder {
    |                    ^^^^^^^^^^^^^^^^^^^^^ expected `EbsBlockDeviceBuilder`, found `()`
...
559 | / impl_ebs_volume_type! {
560 | |     Gp4 ("gp4") {
561 | |         template: TemplatedGp4,
562 | |         volume_size: Gp4VolumeSize,
...   |
565 | |     build(&self, builder) {}
566 | | }
    | |_- in this macro invocation
    |
    = note: this error originates in the macro `impl_ebs_volume_type` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider returning the local binding `builder`
    |
198 ~                     .set_snapshot_id($self.snapshot_id.clone());
199 +                 builder
```

error with `$($impl_body:tt)+`:
```
error: no rules expected `}`
   --> tools/amispec/src/ebs.rs:565:28
    |
177 | macro_rules! impl_ebs_volume_type {
    | --------------------------------- when calling this macro
...
565 |     build(&self, builder) {}
    |                            ^ no rules expected this token in macro call
    |
note: while trying to match meta-variable `$impl_body:tt`
   --> tools/amispec/src/ebs.rs:185:15
    |
185 |             $($impl_body:tt)+
    |               ^^^^^^^^^^^^^
```

File: tools/amispec/src/ebs.rs [hunk 6]
```diff
@@ -0,0 +1,585 @@
...
+    })]
+    pub volume_size: Option<StandardVolumeSize>,
+
+    // Common EBS attributes
+    pub delete_on_termination: Option<bool>,
+    pub encrypted: Option<bool>,
+    pub kms_key_id: Option<String>,
+    pub outpost_arn: Option<String>,
+    pub snapshot_id: Option<String>,
+}
```
sam-berning: I'd prefer we generate this struct as part of the `impl_ebs_volume_type!` macro to avoid the duplicated code here and in the macros on each of these types. We could add the additional fields on top of the common attributes as a part of the pattern, e.g.

```
(
        $name:ident ($awsname:expr) {
            template: $templated_ty:ty,
            volume_size: $size_ty:ident,
            $($extra_attr:ident: $extra_ty:ident,)*
        }

        build(&$self:ident, $builder:ident) {
            $($impl_body:tt)+
        }
    )
```

(I wasn't 100% sure this would work, so I tested it out and pushed a working example to https://github.com/sam-berning/twoliter/tree/amispec-impl, if you want to take a look and test it)

tools/amispec/src/lib.rs
```diff
@@ -0,0 +1,662 @@
...
+    X86_64,
+    #[serde(alias = "aarch64")]
+    Arm64,
+    X86_64Mac,
+    #[serde(alias = "aarch64_mac")]
+    Arm64Mac,
+}
+serde_plain::derive_fromstr_from_deserialize!(Architecture);
+
+impl From<Architecture> for ArchitectureValues {
```
sam-berning: nit: I think I prefer using the fully-qualified path for `ArchitectureValues` as you've done for the other types below

sam-berning (commented):
I think I'm aligned that the `serde-templated` approach is better than just rendering the template before deserializing. Long term I'd like to see it separated from `twoliter`, but overall this approach makes sense to me.

tools/serde-templated/src/templated.rs
cbgbt: I was prompted to add this when using `Templated<String>` in combination with [`bon` for builders](https://bon-rs.com/reference/builder/member/into).

In bon, you can let your `Builder` type allow `impl Into<X>` for any field like this:

```rust
use bon::Builder;

#[derive(Builder)]
struct Example {
    #[builder(into)]
    field: Templated<String>,
}
```

Later in `amispec` we are using `bon` and `serde-templated` together to generate `TemplatedAmiSpec` and `TemplatedAmiSpecBuilder` types. This lets you do the following:


```rust
TemplatedExample::builder()
    .field("{{ hello }}") // Uses the `Into` implementation to make `Templated<String>`
    .build()
```

I thought it best to assume the incoming string was a template, since we can't know for sure.

It might be best to drop this though. It could surprise folks who aren't expecting templates. We also have an `impl<T> From<T> for Templated<T>` which could get pretty confusing for `String` types having different behavior than `&str`.

cbgbt: `impl<T> From<T> for Templated<T>` is useful but I also think it's a bit of a trap for `String` types. I'll see if there's some way to exclude them somehow.

cbgbt:
^ addressed feedback from @sam-berning and added as a Co-author. Thanks for your help with the macro definition!

sam-berning (approved):

cbgbt:
I realized that recursing into `Result<T, E>` in `serde-templated-derive` was somewhat nonsensical, so I've removed that.

jmt-lab (approved):

tools/serde-templated-derive/src/container_visitor.rs
```diff
@@ -0,0 +1,179 @@
...
+/// * HashMap<K, V> -> `K` and `V` are contained types
+/// * Option<HashMap<K, V>> -> `K` and `V` are contained types
+///
+/// Because macros do not have access to type information, this is done on a
+/// best-effort using the `idents` present in our macro input.
+///
+/// Container types include:
+/// * std::option::Option
+/// * std::vec::Vec
+/// * std::collections{BTreeMap, BTreeSet, BinaryHeap, HashMap, HashSet, LinkedList, VecDeque}
```
bcressey:
```suggestion
/// * std::collections::{BTreeMap, BTreeSet, BinaryHeap, HashMap, HashSet, LinkedList, VecDeque}
```

tools/serde-templated/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+    use super::*;
+    use bounded_integer::BoundedU64;
+    use maplit::btreemap;
+    use serde_json::json;
+
+    #[test]
+    fn test_maybe_templated_parse_u64() {
+        let maybe_templated: Templated<u32> = serde_json::from_value(json!(1)).unwrap();
+        assert_eq!(maybe_templated.render(&()).unwrap(), 1);
+    }
```
bcressey: nit: since the test seems to cover parsing a JSON number as a `u32`:
```suggestion
    fn test_maybe_templated_parse_u32() {
        let maybe_templated: Templated<u32> = serde_json::from_value(json!(1)).unwrap();
        assert_eq!(maybe_templated.render(&()).unwrap(), 1);
    }
```

tools/serde-templated-derive/README.md
```diff
@@ -0,0 +1,199 @@
...
+using the provided type.
+
+Due to a limitation of syntax parsing for macros, this is given as a string.
+
+```rust
+#[derive(Templated)]
+struct Example {
+    #[templated(template_as = "Option<Templated<String>>")]
+    value: String,
+}
```
bcressey, line 175: Not really a doc question, but how would this work in practice if the template ended up with a `None` and the struct requires something in the `value` field?

tools/amispec/src/ebs.rs
```diff
@@ -0,0 +1,468 @@
...
+    Io2(Io2),
+    Gp2(Gp2),
+    Gp3(Gp3),
+    St1(St1),
+    Sc1(Sc1),
+}
+
+impl EbsBlockDevice {
+    pub(crate) fn create_sdk_ebs_block_device(&self) -> SdkEbsBlockDevice {
+        let builder = EbsBlockDeviceBuilder::default();
```
bcressey: I'd personally add a macro to cut down the repetition but I know that's not to everyone's taste 😀 
```suggestion
        let builder = EbsBlockDeviceBuilder::default();
        macro_rules! ebs_volume {
            ($device:expr, $voltype:ident) => {
                $device.configure_ebs_block_device(builder).volume_type(SdkVolumeType::$voltype)
            };
        }
```

tools/amispec/src/lib.rs
```diff
@@ -0,0 +1,656 @@
...
+            tpm_support,
+            uefi_data,
+            virtualization_type: _, // Only HVM is supported
+        } = self;
+
+        RegisterImageInput::builder()
+            .name(name)
+            .virtualization_type(VirtualizationType::Hvm.to_string())
+            // `amispec` only supports hvm images.
+            // `paravirtual` is the RegisterImage default, so we just always set it to `hvm` here.
```
bcressey: nit:
```suggestion
            // `amispec` only supports hvm images.
            // `paravirtual` is the RegisterImage default, so we just always set it to `hvm` here.
            .virtualization_type(VirtualizationType::Hvm.to_string())
```

bcressey (approved):
Very nice!

tools/serde-templated-derive/README.md
cbgbt: This would result in code that doesn't compile, unless you also provide a `render_with` attribute to `templated` that tells it how to do the conversion back from the template to the parent.

cbgbt:
^ Addresses feedback from @bcressey.

Your feedback inspired me to shift even more of the `EBSBlockDevice` type definition into macros, so new volumes should only require a new item to be added to the macro invocation.

cbgbt:
^ Rebase on `develop`

jmt-lab (approved):

cbgbt:
^ rebase

sam-berning (approved):

PR #522, opened 2025-04-22
test: add integration test for application-inventory generation

jmt-lab:
**Description of changes:**
Introduces an integration test that will build bottlerocket twice. First it will build bottlerocket with the latest released version of twoliter. Then it will build bottlerocket with the current commits build. Afterwards it does a comparison to make sure the application-inventory.json file is generated correctly and matches the released version.

**Testing done:**
Ran `make integ`

**Terms of contribution:**


tests/integration-tests/Cargo.toml
```diff
@@ -5,8 +5,18 @@ edition = "2021"
 license = "MIT OR Apache-2.0"
 
 [dev-dependencies]
+duct.workspace = true
+flate2.workspace = true
 libc.workspace = true
+lzma-rs = "0.3"
+once_cell = "1.21"
```
sam-berning: nit: I'd prefer we keep all of the dependencies workspace dependencies

tests/integration-tests/src/appinventory.rs
```diff
@@ -0,0 +1,204 @@
...
+
+#[derive(Serialize, Deserialize)]
+pub struct GithubRelease {
+    tag_name: String,
+}
+
+async fn find_latest_version(repository: &str) -> String {
+    // We make a get request to fetch the github releases of twoliter
+    let url = Url::parse(&format!(
+        "https://api.github.com/repos/bottlerocket-os/{repository}/releases"
```
sam-berning: Can we just use https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release to get the latest? Should be fine as long as we don't mark release candidates in Twoliter as "latest"

tests/integration-tests/Cargo.toml
cbgbt: `cargo-deny` is supposed to be forbidding this in CI, but I think it failed here because our `cargo-deny` call doesn't turn on tests? It's not really clear why it didn't flag this.

tests/integration-tests/src/appinventory.rs
```diff
@@ -0,0 +1,204 @@
...
+    ))
+    .expect("invalid url");
+    let client = reqwest::Client::new();
+    let response = client
+        .get(url)
+        .header("Accept", "application/vnd.github+json")
+        .header("X-GitHub-Api-Version", "2022-11-28")
+        .header("User-Agent", "curl/8.9.1") // Pretend to be curl
```
cbgbt: Is it problematic to use a more accurate `User-Agent`?

tests/integration-tests/src/appinventory.rs
```diff
@@ -0,0 +1,204 @@
...
+async fn find_latest_version(repository: &str) -> String {
+    // We make a get request to fetch the github releases of twoliter
+    let url = Url::parse(&format!(
+        "https://api.github.com/repos/bottlerocket-os/{repository}/releases"
+    ))
+    .expect("invalid url");
+    let client = reqwest::Client::new();
+    let response = client
+        .get(url)
+        .header("Accept", "application/vnd.github+json")
```
cbgbt: We should probably optionally authenticate here if we're running in CI so that we don't get throttled. Using `octocrab` would probably make this a lot easier (and also do the nice things like settings the right headers).

tests/integration-tests/src/appinventory.rs
```diff
@@ -0,0 +1,204 @@
...
+    // Make the file executable
+    cmd!("chmod", "+x", &twoliter_bin)
+        .run()
+        .expect("failed to make twoliter executable");
+    (tmp_dir, twoliter_bin)
+}
+
+#[derive(Deserialize, Eq, Debug)]
+#[serde(rename_all = "PascalCase")]
+struct ContentView {
```
cbgbt: Not a hard blocker, but I'd prefer to hoist this as `pub` into something that we could re-use to perform inventory generation in Rust.

tests/integration-tests/src/appinventory.rs
```diff
@@ -0,0 +1,204 @@
...
+
+#[derive(Deserialize, PartialEq, Eq, Debug)]
+#[serde(rename_all = "PascalCase")]
+struct InventoryView {
+    content: Vec<ContentView>,
+}
+
+impl PartialEq for ContentView {
+    fn eq(&self, other: &Self) -> bool {
+        self.name == other.name
```
cbgbt: Can we remove this? It's the same as the `derive` implementation.

tests/integration-tests/src/appinventory.rs
```diff
@@ -0,0 +1,204 @@
...
+        ],
+        cmd_env,
+    );
+    assert!(
+        output.status.success(),
+        "failed to run twoliter build variant"
+    );
+    // Now we should have the application inventory file
+    let aif = tmp_dir.path().join(format!(
+        "build/images/x86_64-aws-ecs-2/latest/application-inventory.json",
```
cbgbt: nits:
* Use a `CONSTANT` for the target variant
* I think `x86_64` is correct since you've mentioned in person that it's `twoliter`'s default if no arch is specified, but we should probably explicitly provide an ARCH and then check for it here. This would protect us against an architecture-specific test breakage if we modify the default arch selection behavior.

tests/integration-tests/src/twoliter_update.rs
```diff
@@ -1,13 +1,15 @@
+use crate::twoliter_build::copy_project_to_temp_dir;
```
cbgbt, line 1: Thanks for these improvements to the integration tests! Do you mind putting them into a separate commit?

cbgbt (commented):
This is great, thanks for writing this! Just a few minor suggestions.

tests/integration-tests/src/appinventory.rs
```diff
@@ -0,0 +1,198 @@
...
+        "https://api.github.com/repos/bottlerocket-os/{repository}/releases/latest"
+    ))
+    .expect("invalid url");
+    let client = reqwest::Client::new();
+    let mut request = client
+        .get(url)
+        .header("Accept", "application/vnd.github+json")
+        .header("X-GitHub-Api-Version", "2022-11-28")
+        // Pretend to be curl
+        .header("User-Agent", "curl/8.9.1");
```
cbgbt: We really ought to use something more accurate here, like `"twoliter-ci"` or similar.

tests/integration-tests/src/appinventory.rs
```diff
@@ -0,0 +1,198 @@
...
+    tmp_dir
+}
+
+// Installs twoliter from the latest non-rc release on github
+async fn install_twoliter_from_release() -> (TempDir, PathBuf) {
+    let tmp_dir = TempDir::new().expect("failed to create temporary directory");
+    let twoliter_bin = tmp_dir.path().join("twoliter");
+    let latest = find_latest_version("twoliter").await;
+    let arch = env::consts::ARCH;
+    let response = reqwest::get(format!("https://github.com/bottlerocket-os/twoliter/releases/download/{latest}/twoliter-{arch}-unknown-linux-musl.tar.xz")).await.expect("failed to download twoliter release archive");
```
cbgbt: We ought to set our user-agent for this request as well.

cbgbt (approved):
Nice, thanks!

We chatted offline and `octocrab` can't be added to our workspace due to an unfortunate issue having to do with `snafu`, the `backtrace` crate, `std::backtrace::Backtrace`, and feature unification.

Basically, `octocrab` has a hard dependency on `snafu` using `std::backtrace::Backtrace`; however, our workspace configures `snafu` to use the `backtrace` crate via `tough` and `parse-datetime`.

To fix this, we need to make the `backtrace` usage in `tough` configurable (or just move to `std`, which has been stable since, I believe, Rust 1.61).

sam-berning (approved):

cbgbt (approved):

PR #524, opened 2025-05-03
pubsys: integrate with amispec

cbgbt:
**Description of changes:**
This change allows `twoliter` users to place an `amispec.toml` template alongside the `Cargo.toml` for their variant, overriding properties of the registered AMI.

Large diff, but a lot of it is tests of the `amispec` generation.

---

    pubsys: integrate with amispec

    Users can now define an `amispec.toml` file alongside their variant's
    Cargo.toml file, specifying overrides for the default AMI properties
    used when registering that variant as an AMI with pubsys.

---

    amispec: add Display implementations to amispec types

---

    amispec: allow u64 inputs to bounded ebs volume fields

**Testing done:**
* Unit tests pass
* Registered an AMI prior to this change, then with this change but only a `name` overridden in `amispec.toml`. This is the diff of `aws describe-images`

```
11c11
<                         "SnapshotId": "snap-001b6ffe9125c9a00",
---
>                         "SnapshotId": "snap-0d2dfc19d789f0725",
21c21
<                         "SnapshotId": "snap-065f575ec94538730",
---
>                         "SnapshotId": "snap-09409c6cf48e306ae",
28c28
<             "CreationDate": "2025-05-02T23:49:33.000Z",
---
>             "CreationDate": "2025-05-02T23:55:12.000Z",
33,34c33,34
<             "ImageId": "REDACTED",
<             "ImageLocation": "REDACTED/bottlerocket-aws-k8s-1.31-aarch64-v1.38.0-78594e2e",
---
>             "ImageId": "REDACTED",
>             "ImageLocation": "REDACTED/bottlerocket-aws-k8s-1.31-aarch64-v1.38.0-78594e2e-extended",
36c36
<             "Name": "bottlerocket-aws-k8s-1.31-aarch64-v1.38.0-78594e2e",
---
>             "Name": "bottlerocket-aws-k8s-1.31-aarch64-v1.38.0-78594e2e-extended"
```
* Registered an AMI with no `amispec.toml` and ensured it used the same default values as prior.
* Registered an AMI using the `unified` partition-plan with the following `amispec.toml`:

```toml
name = "lol what"

[block-device-mappings."/dev/xvda".ebs]
volume-type = "gp3"
iops = "3000"
volume-size = "32"

[block-device-mappings."/dev/xvdb".ebs]
volume-type = "io2"
iops = "800"
volume-size = "16"

[block-device-mappings."/dev/xvdc".ebs]
volume-type = "gp2"
volume-size = "3"
```
Confirmed that the name was `"Name": "lol what",` and that three different EBS volumes were created:

```json
{
    "Ebs": {
        "DeleteOnTermination": true,
        "Iops": 3000,
        "SnapshotId": "REDACTED",
        "VolumeSize": 32,
        "VolumeType": "gp3",
        "Encrypted": false
    },
    "DeviceName": "/dev/xvda"
},
{
    "Ebs": {
        "DeleteOnTermination": true,
        "Iops": 800,
        "VolumeSize": 16,
        "VolumeType": "io2",
        "Encrypted": false
    },
    "DeviceName": "/dev/xvdb"
},
{
    "Ebs": {
        "DeleteOnTermination": true,
        "VolumeSize": 3,
        "VolumeType": "gp2",
        "Encrypted": false
    },
    "DeviceName": "/dev/xvdc"
}
```
* Registered an AMI using the `unified` partition plan, and tried to make the root volume way too small:

```toml
name = "{{ ami.unique_name }} bad size"

[block-device-mappings."/dev/xvda".ebs]
volume-type = "gp3"
volume-size = "1"
```

Confirmed that the name was overridden, only one block device was configured, and my poor `volume-size` directive was ignored:

```json
"BlockDeviceMappings": [
    {
        "Ebs": {
            "DeleteOnTermination": true,
            "SnapshotId": "REDACTED",
            "VolumeSize": 22,
            "VolumeType": "gp3",
            "Encrypted": false
        },
        "DeviceName": "/dev/xvda"
    }
]
```

During this registration, `pubsys` said:

```
00:22:10 [INFO] Merging user-defined amispec template for AMI properties
00:22:10 [WARN] EBS volume 'Gp3(Gp3 { volume_size: Some(Bounded(1)), iops: None, throughput: None, delete_on_termination: Some(true), encrypted: None, kms_key_id: None, outpost_arn: None, snapshot_id: Some("Redacted") })' was configured with size '1', but size hints dictate that it should be larger. Using volume size '22'
00:22:10 [INFO] Making register image call in us-west-2
```

* Confirmed that `uefi` data is properly set for secureboot AMIs.

* Confirmed that AMIs with overridden names are properly detected as being already-registered
* Confirmed that copied AMIs maintain overridden features like `name` and `arch`

**Terms of contribution:**

### DIFF
--- a/tools/pubsys/src/aws/ami/register/merge_toml.rs
+++ b/tools/pubsys/src/aws/ami/register/merge_toml.rs
@@ -1,0 +1,99 @@
+//! Provides a utility for merging toml tables
+
+/// Represents a type that can be "merged" with another version of itself.
+///
+/// This is primarily intended to be used with Map-like types, where keys can be recursively
+/// upserted into `self` from `other`.
+pub(crate) trait Merge {
+    fn merge(&mut self, other: &Self);
+}
+
+impl Merge for toml::Value {
+    fn merge(&mut self, other: &Self) {
+        match (self, other) {
+            (toml::Value::Table(a), toml::Value::Table(b)) => a.merge(b),
+            (a, b) => {
+                *a = b.clone();
+            }
+        }
+    }
+}
+
+impl Merge for toml::Table {
+    fn merge(&mut self, other: &Self) {
+        for (k, v) in other {
+            if let Some(a_val) = self.get_mut(k) {
+                a_val.merge(v);
+            } else {
+                self.insert(k.clone(), v.clone());
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::Merge;
+    use toml::toml;
+
+    #[test]
+    fn test_flat_merge() {
+        let mut a = toml! {
+            a = 1
+            b = 2
+            c = 3
+        };
+
+        let b = toml! {
+            b = 4
+            c = 5
+            d = 6
+        };
+
+        let expected = toml! {
+            a = 1
+            b = 4
+            c = 5
+            d = 6
+        };
+
+        a.merge(&b);
+
+        assert_eq!(a, expected);
+    }
+
+    #[test]
+    fn test_nested_merge() {
+        let mut a = toml! {
+            a = 1
+            b = { c = 2, d = 3 }
+        };
+
+        let b = toml! {
+            b = { d = { g = 6 }, e = 4 }
+            f = 5
+        };
+
+        let expected = toml! {
+            a = 1
+            b = { c = 2, d = { g = 6 }, e = 4 }
+            f = 5
+        };
+
+        a.merge(&b);
+
+        assert_eq!(a, expected);
+    }
+
+    #[test]
+    fn test_empty_merge() {
+        let mut a = toml! {
+            a = 1
+            b = 2
+        };
+        let expected = a.clone();
+
+        a.merge(&toml::map::Map::new());
+        assert_eq!(a, expected);
+    }
+}
--- a/tools/pubsys/src/aws/ami/register/mk_amispec.rs
+++ b/tools/pubsys/src/aws/ami/register/mk_amispec.rs
@@ -1,0 +1,966 @@
+//! This module generates an `amispec` to be used during the AMI registration process.
+//!
+//! An `amispec` defines the arguments that are passed to the EC2 RegisterImage call, including the
+//! EBS block device mappings which control EBS volumes and their properties.
+//!
+//! `amispec`s can be defined using handlebars-templated values for any field, allowing pubsys to
+//! inject AMI registration settings into the user-provided template.
+//!
+//! The `amispec` used to register an AMI is created using the following process:
+//! * A default `amispec` template is defined as a TOML table within pubsys
+//! * The default TOML table is modified to support features of the variant that is being registered
+//!     * Default EBS block device mappings are created for each snapshot
+//!     * If the variant supports secure boot, then the template has uefi-data updated, and sets
+//!       the default `boot-mode` to `uefi-preferred`
+//! * The default TOML template is merged with the user-provided TOML document, using a recursive
+//!   "upsert" procedure.
+//! * The resulting TOML document is parsed as a `TemplatedAmiSpec`
+//! * The `TemplatedAmiSpec` is finally rendered into the final `AmiSpec`.
+//!
+//! The handlebars render context is defined in [`AmispecRenderContext`] and includes basic
+//! AMI properties like the desired architecture and name, as well as information regarding
+//! EBS block devices required for the volumes based on the variant definition.
+use super::merge_toml::Merge;
+use super::BottlerocketSnapshots;
+use crate::aws::ami::VariantManifest;
+use amispec::{
+    ebs::{EbsBlockDeviceKind, InvalidValueError},
+    AmiSpec, TemplateOf, TemplatedAmiSpec,
+};
+use buildsys::manifest::{ImageFeature, PartitionPlan};
+use log::{debug, info, warn};
+use serde::Serialize;
+use snafu::{OptionExt, ResultExt, Snafu};
+use std::collections::HashMap;
+
+use crate::aws::ami::AmiArgs;
+
+type Result<T, E = AmispecError> = std::result::Result<T, E>;
+
+const ROOT_DEVICE_NAME: &str = "/dev/xvda";
+const DATA_DEVICE_NAME: &str = "/dev/xvdb";
+pub(crate) const VIRT_TYPE: &str = "hvm";
+
+lazy_static::lazy_static! {
+    /// The default amispec template.
+    ///
+    /// pubsys modifies this structure based on attributes of the variant being registered:
+    /// * If the variant has Secure Boot enabled,
+    ///     * The boot-mode is set to uefi-preferred
+    ///     * uefi-data is added to the template
+    static ref DEFAULT_AMISPEC_TEMPLATE: toml::Table = toml::toml! {
+        name = "{{ ami.unique_name }}"
+        description = "{{ ami.description }}"
+        architecture = "{{ ami.arch }}"
+        root-device-name = "/dev/xvda"
+        sriov-net-support = "simple"
+        virtualization-type = VIRT_TYPE
+        ena-support = true
+    };
+
+    static ref DEFAULT_ROOT_DEVICE_TEMPLATE: toml::Table = toml::toml! {
+        ["/dev/xvda".ebs]
+        volume-type = "gp2"
+        volume-size = "{{ block_devices.root.volume_size }}"
+        snapshot-id = "{{ block_devices.root.snapshot_id }}"
+        delete-on-termination = true
+    };
+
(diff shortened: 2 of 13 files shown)


twoliter/embedded/Makefile.toml
```diff
@@ -1307,6 +1310,10 @@ if [ "${is_split}" == "yes" ] ; then
    data_volume_args+=(--data-image "${data_image}")
 fi
 
+if [ -f "${BUILDSYS_AMISPEC_TEMPLATE}" ]; then
```
bcressey: nit: generally `-s` is a better check unless a completely empty file is tolerated:
```suggestion
if [ -s "${BUILDSYS_AMISPEC_TEMPLATE}" ]; then
```

tools/pubsys/src/aws/ami/register/mod.rs
```diff
@@ -162,17 +53,85 @@ async fn _register_image(
...
+        region: &Region,
+        ebs_client: EbsClient,
+        ec2_client: &Ec2Client,
+        cleanup_snapshot_ids: &mut Vec<String>,
+    ) -> Result<BottlerocketSnapshots> {
+        debug!("Uploading images into EBS snapshots in {}", region);
+        let uploader = SnapshotUploader::new(ebs_client);
+        let os_snapshot =
+            snapshot_from_image(&ami_args.os_image, &uploader, None, ami_args.no_progress)
+                .await
```
bcressey: nit: can be left for later, but the refactor makes it more obvious that we could create and wait for both snapshots in parallel instead of serially

tools/pubsys/src/aws/ami/register/merge_toml.rs
```diff
@@ -0,0 +1,99 @@
+//! Provides a utility for merging toml tables
```
sam-berning, line 1: nit: we've implemented something similar already in bottlerocket-settings-sdk ([link][settings-sdk]) 

doesn't make sense to take a dependency on the settings-sdk here, but maybe worth considering pulling this out into a crate that both can depend on? given it's so short and simple, definitely non-blocking

[settings-sdk]: https://github.com/bottlerocket-os/bottlerocket-settings-sdk/blob/0d825fcc0a199e233bed88ea9d9d1685c6d966bc/bottlerocket-defaults-helper/src/merge_toml.rs

tools/pubsys/src/aws/ami/register/mk_amispec.rs
```diff
@@ -0,0 +1,833 @@
...
+        AmiArgs {
+            os_image: PathBuf::new(),
+            data_image: None,
+            variant_manifest: VariantManifest::new(
+                toml::toml! {
+                    [package]
+                    name = "test-variant"
+
+                    [package.metadata.build-variant.image-layout]
+                    os-image-size-gib = 100
```
bcressey: With `partition-plan` unspecified, this should default to the "split" layout, so I'm surprised that it ended up with "unified".

tools/pubsys/src/aws/ami/register/mk_amispec.rs
```diff
@@ -0,0 +1,833 @@
...
+    pub volume_size: u64,
+}
+
+impl BottlerocketSnapshots {
+    /// Creates the block device handlebars render context from a set of created EBS snapshots
+    fn block_device_render_context(
+        &self,
+        image_layout: &ImageLayout,
+    ) -> HashMap<String, BlockDeviceRenderContext> {
+        let (os_volume_size, data_volume_size) = image_layout.publish_image_sizes_gib();
```
bcressey: It seems like we could additionally validate the partition plan from the image layout; if it's "split" and we only have one snapshot, then we should throw an error?

tools/pubsys/src/aws/ami/register/mod.rs
cbgbt: I want to do it this way, but I was wondering if `coldsnap`'s progress output would do the right thing. I haven't tested, but I was thinking about circling back to `coldsnap` and possibly making a concurrent API there.

I should just try it and see what happens though.

tools/pubsys/src/aws/ami/register/mk_amispec.rs
```diff
@@ -0,0 +1,833 @@
...
+        AmiArgs {
+            os_image: PathBuf::new(),
+            data_image: None,
+            variant_manifest: VariantManifest::new(
+                toml::toml! {
+                    [package]
+                    name = "test-variant"
+
+                    [package.metadata.build-variant.image-features]
+                    uefi-secure-boot = true
```
bcressey, line 482: A different test case relevant to my interests would be to set `uefi-secure-boot = true`, and pair it with an ami spec template like:
```
tpm-support = "v2.0"
boot-mode = "uefi"
```

To confirm that the default `boot-mode = "uefi-preferred"` gets replaced.

tools/pubsys/src/aws/ami/register/mod.rs
cbgbt: Oh, I see now that `coldsnap` defers the progress implementation to the caller. Good design!

This is a simple enough change to make, I'm going to do it.

tools/pubsys/src/aws/ami/register/mk_amispec.rs
cbgbt: The data flow in `pubsys` is a bit hard to follow. It needs information about the variant being published to register correctly; however, `pubsys'` CLI arguments are in many cases redundant with the manifest. So now `pubsys` has to decide which to respect.

Both before and after this change, `pubsys` uses its CLI arguments to determine properties of the variant (only falling back to using the manifest itself when it can't, like when trying to determine what to do about Secure Boot.)

I'd really like to refactor it so that the decision of how to derive this data is more clear from the way that the software is written, rather than implied by reading some of the implementation details. It's not clear how long that would take me though, so I've stopped short of doing that here.

tools/pubsys/src/aws/ami/register/mk_amispec.rs
```diff
@@ -0,0 +1,833 @@
...
+                    ebs::Gp3::builder()
+                        .snapshot_id("some-random-snapshot")
+                        .volume_size(1).unwrap()
+                        .build()
+                )
+                .build()
+                .with_device_name("/dev/xvda"))
+
+            .build()
+        ; "snapshots with no hints are ignored"
```
bcressey, line 713: What does "snapshots with no hints" mean? Are the two snapshots in the hash map ignored because neither of them has a matching snapshot ID? Otherwise, they both seem to have a volume size.

bcressey (approved):

tools/pubsys/src/aws/ami/register/mk_amispec.rs
```diff
@@ -0,0 +1,833 @@
...
+        description = "{{ ami.description }}"
+        architecture = "{{ ami.arch }}"
+        root-device-name = "/dev/xvda"
+        sriov-net-support = "simple"
+        virtualization-type = VIRT_TYPE
+        ena-support = true
+    };
+
+    static ref DEFAULT_ROOT_DEVICE_TEMPLATE: toml::Table = toml::toml! {
+        ["/dev/xvda".ebs]
```
sam-berning, line 62: nit: we can use the `*_DEVICE_NAME` constants here

tools/pubsys/src/aws/ami/register/mk_amispec.rs
cbgbt: At this point in the process, we have an `amispec` with a series of block device mappings. We also have a set of snapshots that we've registered based on block devices from the variant definition.

The variant's Manifest can be used to get the desired size of the block device (referred to as "size hint" here.)

In most cases, the `amispec`'s device mappings are a superset of the variant-derived block devices, but technically the sets could be completely disjoint at this point.

So basically, this function uses the snapshot ID of a block device mapping to attempt to determine whether or not it was derived from one of the Variant-defined block devices. If it is, it "corrects" the volume size.

In this test, the AMI has a block device mapping which refers to some tangential snapshot. In this case, it isn't "variant-derived" and therefor has no size hints -- `pubsys` will not attempt to alter the size of this volume.


Possibly using `variant-derived` as a descriptor in these symbol names would make it more clear how things relate?

sam-berning (approved):

tools/pubsys/src/aws/ami/register/mk_amispec.rs
cbgbt: Unfortunately the `toml!` macro really hates to have interpolated variables inside :(.

sam-berning: oh I guess that makes sense... would be difficult to know what was meant to be a literal string and what was meant to be an identifier

sam-berning (approved):

PR #532, opened 2025-05-12
fix: make sure to read env var at build script exec time, not at build time

mikn:
**Description of changes:**
We briefly discussed this during the community meeting. I encountered this when building twoliter using Bazel, which I understand don't follow maybe the implicit expectations of the regular cargo build process, but the "problem" is that you rely on `DATA_INPUT_DIR` in runtime being derived from `CARGO_MANIFEST_DIR` at build time. If this is different between the build script being built (when the `env!` macro is expanded into the const) and the build script being executed (when it produces the `twoliter` binary), the current code will break.
(this is something Bazel, with its parallelization and sandboxing, does)

**Testing done:**
Tested building using both Bazel and the regular build process and it worked fine. :)

**Terms of contribution:**

### DIFF
--- a/twoliter/build.rs
+++ b/twoliter/build.rs
@@ -14,8 +14,6 @@ use std::path::{Path, PathBuf};
 use std::time::SystemTime;
 use std::{env, fs};
 
-const DATA_INPUT_DIR: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/embedded");
-
 fn main() {
     let paths = Paths::new();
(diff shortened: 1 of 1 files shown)


mikn:
@cbgbt don't think you were in that meeting, but would appreciate your review nonetheless :pray:

cbgbt (approved):
Thanks for the fix!

Note to self for GH cross-reference: rebase atop this for #537

jmt-lab (approved):

PR #534, opened 2025-05-13
pubsys: allow ami copies when uefi-data not provided

cbgbt:
**Description of changes:**
#524 introduced a bug.

* Previously, `uefi-data` was always expected by `pubsys ami`; however, if the command would only result in AMI copies, the file was never parsed.
* With #524, the file is parsed at the same time as CLI arguments. This would lead to an issue in which sbkey profiles aren't present while attempting to copy an image.

This change makes `uefi-data` optional, and only requires that it be present when needed to register an AMI that should have `uefi-secure-boot` enabled.

**Testing done:**
* Registered an AMI with UEFI secure boot
* Removed the UEFI data from my workspace and attempted to copy the AMI (succeeded)

**Terms of contribution:**

### DIFF
--- a/tools/pubsys/src/aws/ami/mod.rs
+++ b/tools/pubsys/src/aws/ami/mod.rs
@@ -55,7 +55,7 @@ pub(crate) struct AmiArgs {
 
     /// Path to the UEFI data
     #[arg(short = 'e', long, value_parser = parse_uefi_data)]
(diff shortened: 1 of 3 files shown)


jmt-lab (approved):
LGTM

sam-berning (approved):

PR #537, opened 2025-05-13
twoliter: compress included tool archives

cbgbt:
**Description of changes:**
This reduces the binary size of `twoliter` built with `--profile=dist` from ~120MB to ~58MB. Unpacking the tools takes less than a second.

---
    tools: implement include-env-compressed
    
    Implements a macro for including binary assets that are zstd-compressed
    when compiled for release builds.
---
    twoliter: split and compress included tools
    
    Twoliter's "tools" are now included within their own crates to minimize
    the number incremental builds when some archives change.
    
    Uses include-env-compressed to compress the included archives.
---
    workspace: mark twoliter as the default build target
---

**Testing done:**
* Unit and integration tests pass
* Built & published core-kit
* Built & registered AMI

**Terms of contribution:**


twoliter/src/tool-crates/buildsys/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
```
jmt-lab, line 1: Is there no way to combine these crates with the actual subcrates for the tool itself? Why not add the buildscript to the actual buildsys crate?

cbgbt: I may not be understanding, do you mean add a sub-crate to `tools/buildsys` that does this, then pull it in from `twoliter`?

Or do you mean modify the existing `buildsys` crate? If that's what you mean, I'm pretty sure this would be a circular dependency -- We can't have buildsys requiring a pre-compiled version of itself in order to compile.

tools/include-env-compressed/tests/include_build_script.rs
```diff
@@ -0,0 +1,29 @@
+use include_env_compressed::ArchiveKind;
+
+const BUILD_SCRIPT_CONTENTS: &str = r#"
```
sam-berning: why not use `include_bytes!` here or otherwise read the file contents to a string as a `lazy_static` or in the test? seems brittle to have a copy of the file contents as a literal string in the test.

twoliter/src/tool-crates/embedded-bundle/build.rs
```diff
@@ -14,7 +14,7 @@ use std::path::{Path, PathBuf};
 use std::time::SystemTime;
 use std::{env, fs};
 
-const DATA_INPUT_DIR: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/embedded");
+const DATA_INPUT_DIR: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/../../../embedded");
```
sam-berning: nit: I don't *love* having this path point outside of the crate, how would you feel about moving the embedded files into this crate?

sam-berning (approved):

twoliter/src/tool-crates/embedded-bundle/build.rs
cbgbt: I'm OK with doing that, or even symlinking it. I just know that I end up opening `twoliter/embedded` very frequently, so I'm reluctant to bury it.

sam-berning: yeah I agree with that, it's a path I'm pretty used to as well. I'd be in favor of moving the directory into this crate and symlinking to it at `twoliter/embedded`

cbgbt: I agree that that's the correct ownership of the data, but I may make the symlink point the other way so that the GitHub UI also continues to show `twoliter/embedded`

cbgbt:
^ rebase on develop, address feedback from @sam-berning 

Sorry for messy inter-diff -- I forgot to separate my rebase into a different push.

sam-berning (approved):

jmt-lab (approved):

PR #545, opened 2025-06-09
testsys: bump to 0.0.17 version

zongzhidanielhu:
Issue number:

Closes #

Description of changes:
testsys: bump to 0.0.17 version

Terms of contribution:

### DIFF
--- a/clarify.toml
+++ b/clarify.toml
@@ -8,9 +8,17 @@ ignore-licenses = [
 [clarify.aws-lc-sys]
 expression = "ISC AND (Apache-2.0 OR ISC) AND OpenSSL AND MIT"
 license-files = [
-  { path = "aws-lc/LICENSE", hash = 0xb6d14686 },
-  { path = "LICENSE", hash = 0xf308ccd7 },
-  { path = "aws-lc/third_party/fiat/LICENSE", hash = 0x75829ee2 },
+    { path = "LICENSE", hash = 0xf308ccd7 },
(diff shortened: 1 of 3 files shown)


clarify.toml
```diff
@@ -8,9 +8,17 @@ ignore-licenses = [
...
 license-files = [
-  { path = "aws-lc/LICENSE", hash = 0xb6d14686 },
-  { path = "LICENSE", hash = 0xf308ccd7 },
-  { path = "aws-lc/third_party/fiat/LICENSE", hash = 0x75829ee2 },
+    { path = "LICENSE", hash = 0xf308ccd7 },
+    { path = "aws-lc/LICENSE", hash = 0xb6d14686 },
+    { path = "aws-lc/third_party/fiat/LICENSE", hash = 0x75829ee2 },
+]
+
+[clarify.aws-lc-fips-sys]
```
jpculp, line 16: This seems right, but can we keep the clarified crates alphabetized?

jpculp (commented):
nit. The branch name can be a point of confusion. Better to say something like `testsys-0.0.17`. Not necessarily a blocker, but for next time.

yeazelm (approved):

gthao313 (approved):

jpculp (changes requested):
It looks like your last push undid the ordering fix.

zongzhidanielhu:
> It looks like your last push undid the ordering fix.

ironically, if you change the order, it won't build. Strange thing

PR #549, opened 2025-06-21
testsys: add support to delete individual tests

### LINKED ISSUE #548: Add support for specific test deletion in testsys
Currently, testsys (in `twoliter/tools` and in the actual [bottlerocket-test-system](https://github.com/bottlerocket-os/bottlerocket-test-system) repo) has the capability to delete tests+resources based on arch and variant filters -- besides deleting them all or based on status (`running`, `failed`, `passed`).

While this is ideal for production environments, for development it is cumbersome to not be able to delete specific tests as you cannot easily remove certain tests -- especially those stuck in `waiting` status -- while preserving desired ones.

I would like support for a feature in testsys -- either in `twoliter/tools` or [bottlerocket-test-system](https://github.com/bottlerocket-os/bottlerocket-test-system) repo) -- to allow for specific test deletion.

koooosh:
**Issue number:**

Closes #548 

**Description of changes:**

This PR adds the capability for testsys to delete individual tests and its resources (`reset-single-test` task).

This is implemented by label filtering (similar to how it's done for aarch and variant) thru the use of a new label `test-name` which is comprised of the `cluster_resource_name` + `name_suffix` fields   that is unique to each test.

The various test CRDs in the following test type files are updated to add this new label during creation so they can be filtered for later:
- `aws-ecs.rs`
- `migration.rs`
- `sonobuoy.rs`

Ideally, I wanted to add support for a list of tests being passed rather than a single one, but that seems to require additional changes in the TestManager logic in [bottlerocket-test-system](https://github.com/bottlerocket-os/bottlerocket-test-system), so I wanted to release this in the meantime. Users can invoke this command in a script to delete multiple specified tests.

**Testing done:**
- Twoliter builds with changes
- `test-name` label gets created for test CRDs and matches the actual name of the test:
```
$ kd test x86-64-aws-k8s-130-nvidia-quick -n testsys
Name:         x86-64-aws-k8s-130-nvidia-quick
Namespace:    testsys
Labels:       testsys/arch=x86_64
              ...
              testsys/test-name=x86-64-aws-k8s-130-nvidia-quick

$ kd test x86-64-aws-k8s-130-nvidia-conformance -n testsys
Name:         x86-64-aws-k8s-130-nvidia-conformance
Namespace:    testsys
Labels:       testsys/arch=x86_64
              ...
              testsys/test-name=x86-64-aws-k8s-130-nvidia-conformance

$ kd test x86-64-aws-k8s-130-nvidia-test -n testsys
Name:         x86-64-aws-k8s-130-nvidia-test
Namespace:    testsys
Labels:       testsys/arch=x86_64
              ...
              testsys/test-name=x86-64-aws-k8s-130-nvidia-test

$ kd test x86-64-vmware-k8s-129-conformance  -n testsys
Name:         x86-64-vmware-k8s-129-conformance
Namespace:    testsys
Labels:       testsys/arch=x86_64
              ...
              testsys/test-name=x86-64-vmware-k8s-129-conformance
```

- `reset-single-test` task deletes the test and its related resources:
```
$ cargo make     -e=TWOLITER_ALLOW_SOURCE_INSTALL=false     -e=TWOLITER_ALLOW_BINARY_INSTALL=false     -e=TWOLITER_SKIP_VERSION_CHECK=true reset-single-test x86-64-aws-k8s-130-nvidia-test
....
[cargo-make][1] INFO - Running Task: reset-single-test
Starting delete for x86-64-aws-k8s-130-nvidia-test
Delete finished for x86-64-aws-k8s-130-nvidia-test
Starting delete for x86-64-aws-k8s-130-nvidia-instances-unuj
Delete finished for x86-64-aws-k8s-130-nvidia-instances-unuj
Starting delete for x86-64-aws-k8s-130-nvidia
Delete finished for x86-64-aws-k8s-130-nvidia
[2025-06-21T00:33:21Z INFO  testsys::delete] Delete finished
```

**Terms of contribution:**

### DIFF
--- a/tools/testsys/src/aws_ecs.rs
+++ b/tools/testsys/src/aws_ecs.rs
@@ -146,6 +146,11 @@ impl CrdCreator for AwsEcsCreator {
         let labels = test_input.crd_input.labels(btreemap! {
             "testsys/type".to_string() => test_input.test_type.to_string(),
             "testsys/cluster".to_string() => cluster_resource_name.to_string(),
+            "testsys/test-name".to_string() => format!(
+                "{}-{}",
+                cluster_resource_name,
+                test_input.name_suffix.unwrap_or(test_input.crd_input.test_flavor.as_str())
+            ),
         });
 
         let test_crd = EcsTestConfig::builder()
@@ -222,6 +227,11 @@ pub(crate) fn workload_crd(region: &str, test_input: TestInput) -> Result<Test> 
     let labels = test_input.crd_input.labels(btreemap! {
         "testsys/type".to_string() => test_input.test_type.to_string(),
         "testsys/cluster".to_string() => cluster_resource_name.to_string(),
+        "testsys/test-name".to_string() => format!(
+            "{}-{}",
+            cluster_resource_name,
+            test_input.name_suffix.unwrap_or("test")
+        ),
     });
     let gpu = test_input.crd_input.variant.variant_flavor() == Some("nvidia");
     let plugins: Vec<_> = test_input
--- a/tools/testsys/src/delete.rs
+++ b/tools/testsys/src/delete.rs
@@ -11,6 +11,10 @@ pub(crate) struct Delete {
     #[clap(long)]
     test: bool,
 
+    /// Delete a particular test
+    #[clap(long)]
+    test_name: Option<String>,
+
     /// Focus status on a particular arch
     #[clap(long)]
     arch: Option<String>,
@@ -49,12 +53,16 @@ impl Delete {
         };
         let crd_type = self.test.then_some(CrdType::Test);
         let mut labels = Vec::new();
+        if let Some(test_name) = self.test_name {
+            labels.push(format!("testsys/test-name={}", test_name))
+        };
         if let Some(arch) = self.arch {
             labels.push(format!("testsys/arch={}", arch))
         };
         if let Some(variant) = self.variant {
             labels.push(format!("testsys/variant={}", variant))
         };
+
         let mut stream = client
             .delete(
                 &SelectionParams {
@@ -63,7 +71,7 @@ impl Delete {
                     crd_type,
                     ..Default::default()
                 },
-                false,
+                !self.test,
             )
             .await?;
 
--- a/twoliter/embedded/Makefile.toml
+++ b/twoliter/embedded/Makefile.toml
@@ -1875,6 +1875,16 @@ script = [
     '''
 ]
 
+# This task will clear a specific test by name and its resources from the testsys cluster.
+[tasks.reset-single-test]
+script = [
+    '''
+    set -eu
+    export PATH="${TWOLITER_TOOLS_DIR}:${PATH}"
+    testsys --log-level=${TESTSYS_LOG_LEVEL} ${CARGO_MAKE_TESTSYS_ARGS} delete --test-name ${@}
+    '''
+]
+
 # This task will clear all testsys components from the testsys cluster.
 [tasks.uninstall-test]
 script = [
--- a/tools/testsys/src/migration.rs
+++ b/tools/testsys/src/migration.rs
@@ -21,6 +21,11 @@ pub(crate) fn migration_crd(
     let labels = migration_input.crd_input.labels(btreemap! {
         "testsys/type".to_string() => "migration".to_string(),
         "testsys/cluster".to_string() => cluster_resource_name.to_string(),
+        "testsys/test-name".to_string() => format!(
+            "{}-{}",
+            cluster_resource_name,
+            migration_input.name_suffix.unwrap_or(migration_input.crd_input.test_flavor.as_str())
+        ),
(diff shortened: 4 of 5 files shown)


tools/testsys/src/aws_ecs.rs
```diff
@@ -146,6 +146,11 @@ impl CrdCreator for AwsEcsCreator {
         let labels = test_input.crd_input.labels(btreemap! {
             "testsys/type".to_string() => test_input.test_type.to_string(),
             "testsys/cluster".to_string() => cluster_resource_name.to_string(),
+            "testsys/test-name".to_string() => format!(
+                "{}-{}",
+                cluster_resource_name,
+                test_input.name_suffix.unwrap_or(test_input.crd_input.test_flavor.as_str())
```
koooosh, line 152: The fallback value for `name_suffix` during label creation matches [the fallback during CRD creation](https://github.com/bottlerocket-os/twoliter/blob/ed80cb09a3a9f90d4b58d131421a635e25899809/tools/testsys/src/aws_ecs.rs#L195).

tools/testsys/src/delete.rs
```diff
@@ -63,7 +71,7 @@ impl Delete {
                     crd_type,
                     ..Default::default()
                 },
-                false,
+                !self.test,
```
koooosh, line 74: This is a flag to [include_dependencies](https://github.com/bottlerocket-os/bottlerocket-test-system/blob/develop/model/src/test_manager/manager.rs#L254C9-L254C29) which is used to get the resources of the test to delete.

Although this was set to `false` previously, resources were still able to be queried since the `crd_type` field in `SelectionParams` was set to None, meaning it returned both Test and Resource CRDs. In our case now, since only Test CRDs have the `test-name` label, we need to change this flag so that it is also able to get the Resources CRDs associated with it for `reset-single-test`

twoliter/embedded/Makefile.toml
```diff
@@ -1865,6 +1865,16 @@ script = [
     '''
 ]
 
+# This task will clear a specific test by name from the testsys cluster.
+[tasks.clean-single-test]
```
koooosh: Ideally, this new task name would be `clean-test` and the existing `clean-test` task would be renamed to `clean-tests` plural, but I'm afraid that might introduce a breaking change for users who have automation/scripts for this.

Same applies for `reset-test`

gthao313:
Nice work! I'm pretty surprised we can add single test/resource deletion without huge code change! 

I have few questions 
1.  Why do we need both new task? `clean-single-test task` and `reset-single-test task`? Can you provide a scenario that we need both? Base on my understanding, TestSys would re-create the instance resource for each test, so If we don't clean up the instance resource along with the test, and next new test would create a new instance resource with it. Therefore, I don't see the point why we need to keep the resource? `reset-single-test task` looks pretty sufficient. 
2. I don't see this commit updates vmware tests. Should we also update it as well?

koooosh:
1. Yep, you're right that a new test will create a new instance resource, but I was thinking about the cluster resource. For example, in the following output:
```
[cargo-make][1] INFO - Running Task: reset-single-test
Starting delete for x86-64-aws-k8s-130-nvidia-test
Delete finished for x86-64-aws-k8s-130-nvidia-test
Starting delete for x86-64-aws-k8s-130-nvidia-instances-unuj
Delete finished for x86-64-aws-k8s-130-nvidia-instances-unuj
Starting delete for x86-64-aws-k8s-130-nvidia
Delete finished for x86-64-aws-k8s-130-nvidia
```
I'm referring to `x86-64-aws-k8s-130-nvidia`. AFAIK, a new test doesn't create a new cluster resource if it already exists, so that was my reasoning for including the `clean-single-test` task as well. However, if you believe the cluster resource is not necessary to preserve, I'm fine with removing the `clean-single-test` task.

2. I don't believe the `vmware_k8s.rs`  file needs updating as [the vmware test/workload/migration CRD methods use the methods in the sonobouy and migration.rs files under the hood.](https://github.com/bottlerocket-os/twoliter/blob/develop/tools/testsys/src/vmware_k8s.rs#L272-L295)

gthao313:
1. For cluster resource, if we don't specify   `cluster-destruction-policy = "onTestSuccess"`, the cluster resource won't be deleted. But I agree we should keep both commands.
2. could please test a VMware variant to check if it would be deleted properly?

koooosh:
1. After experimenting a bit more, I now actually agree with your initial suggestion to remove the `clean` task as it leaves behind the instance resource for the test. This resource cannot be removed unless you run `reset-test` which this feature is attempting to avoid!

2. Launched a vmware test and verified that its `test-name` label is correct (added to description) and that it's able to be deleted

koooosh:
^ Force push removes the `clean-single-test` task as discussed

gthao313 (approved):
LGTM!

jmt-lab (approved):

PR #551, opened 2025-06-26
twoliter:Add project-vendor as a new field in Twoliter.toml

### LINKED ISSUE #562: Add project-vendor field and schema migration system to Twoliter
## Overview
This issue tracks the implementation of a new `project-vendor` field in Twoliter.toml and a schema migration system to handle backward compatibility with existing projects.

## Background
Currently, Twoliter.toml files don't have a way to specify the vendor name that should appear in the OS release information. We need to add a `project-vendor` field and ensure it's propagated through the build system to the final OS image.

sky1122:
**Issue number:**

Closes #562

**Description of changes:**
Adding a project-vendor field to Twoliter.toml to set the VENDOR_NAME value in os-release.

To maintain backward compatibility for Twoliter.toml, a migration module will be implemented that handles the conversion between schema version 1 and version 2 of Twoliter.toml. This ensures that customers using schema version 1 can continue to use their existing Twoliter.toml files when updating to the newest version.

For both schema version 1 Twoliter.toml files and the VENDOR_NAME field in os-release, the default value will be set to "Bottlerocket".

**Testing done:**
```
bash-5.1# cd os-release
bash: cd: os-release: Not a directory
bash-5.1# cat os-release
NAME=Bottlerocket
ID=bottlerocket
VERSION="1.42.0 (aws-k8s-1.32-fips)"
PRETTY_NAME="Bottlerocket OS 1.42.0 (aws-k8s-1.32-fips)"
VARIANT_ID=aws-k8s-1.32-fips
VERSION_ID=1.42.0
BUILD_ID=8c0f7c83-dirty
VENDOR_NAME=xxxxxxx
HOME_URL="https://github.com/bottlerocket-os/bottlerocket"
SUPPORT_URL="https://github.com/bottlerocket-os/bottlerocket/discussions"
BUG_REPORT_URL="https://github.com/bottlerocket-os/bottlerocket/issues"
DOCUMENTATION_URL="https://bottlerocket.dev"
bash-5.1#
```

**Terms of contribution:**


sky1122:
force pushed to drop unnecessary build artifacts

sky1122:
force pushed to make the clippy happy

tools/buildsys/src/builder.rs
```diff
@@ -469,9 +471,16 @@ impl DockerBuild {
...
+                external_kit_dependencies: ExternalKitMetadataView::load(
+                    args.common.root_dir.clone(),
+                )
+                .context(error::GraphSnafu)?
+                .list(),
+                project_vendor: {
+                    ExternalKitMetadataView::load(args.common.root_dir)
+                        .context(error::GraphSnafu)?
+                        .get_project_vendor()
+                },
```
cbgbt: nit: we can drop the curly braces around this expression.

tools/buildsys/src/builder.rs
```diff
@@ -469,9 +471,16 @@ impl DockerBuild {
...
-                    .context(error::GraphSnafu)?
-                    .list(),
+                external_kit_dependencies: ExternalKitMetadataView::load(
+                    args.common.root_dir.clone(),
+                )
+                .context(error::GraphSnafu)?
+                .list(),
+                project_vendor: {
+                    ExternalKitMetadataView::load(args.common.root_dir)
+                        .context(error::GraphSnafu)?
```
cbgbt: Rather than parse this twice, we should probably hoist the loading of `ExternalKitMetadataView` out of this constructor.

tools/buildsys/src/manifest.rs
```diff
@@ -356,6 +356,9 @@ impl Manifest {
 pub struct ExternalKitMetadataView {
     #[serde(rename = "kit")]
     kits: Vec<ImageView>,
+    /// set thie project_vendor optional because older version of twoliter
+    /// doesn't have this field, and the default value will be `Bottlerocket`
```
cbgbt: I think these comments are stale, we can probably just remove them.

tools/buildsys/src/manifest.rs
```diff
@@ -382,6 +385,10 @@ impl ExternalKitMetadataView {
             .map(|x| format!("{}/{}", x.vendor, x.name))
             .collect()
     }
+
+    pub fn get_project_vendor(&self) -> String {
+        self.project_vendor.clone()
+    }
```
cbgbt:
```suggestion
    pub fn get_project_vendor(&self) -> &str {
        self.project_vendor.as_ref()
    }
```

twoliter/embedded/rpm2img
```diff
@@ -375,7 +375,7 @@ PRETTY_NAME="${PRETTY_NAME} ${VERSION}"
 VARIANT_ID=${VARIANT}
 VERSION_ID=${VERSION_ID}
 BUILD_ID=${BUILD_ID}
-VENDOR_NAME="Bottlerocket"
+VENDOR_NAME=${PROJECT_VENDOR}
 HOME_URL="https://github.com/bottlerocket-os/bottlerocket"
```
cbgbt, line 379: This change seems correct but I think it might be in the wrong commit (it's in the `buildsys` commit)

twoliter/src/project/lock/mod.rs
```diff
@@ -122,13 +122,16 @@ pub(crate) struct Lock {
     pub sdk: LockedImage,
     /// Resolved kit dependencies
     pub kit: Vec<LockedImage>,
+    /// The project vendor
+    pub project_vendor: String,
 }
```
cbgbt, line 131: I think we need to bump the schema version of the lock if we add this field.

twoliter/src/project/lock/mod.rs
```diff
@@ -122,13 +122,16 @@ pub(crate) struct Lock {
...
 
 impl PartialEq for Lock {
     fn eq(&self, other: &Self) -> bool {
         self.schema_version == other.schema_version
             && self.sdk == other.sdk
             && self.kit == other.kit
+            && self.project_vendor == other.project_vendor
     }
 }
```
cbgbt: This problem existed before your PR, but I think this `impl PartialEq` should actually just be derived?

e.g. if we add `#[derive(PartialEq)]` to `Lock` I am pretty sure it would generate an equivalent implementation of this trait.

twoliter/src/project/mod.rs
```diff
@@ -500,8 +504,11 @@ impl VendedArtifact for Image {
 #[serde(rename_all = "kebab-case")]
 struct UnvalidatedProject {
     schema_version: SchemaVersion<1>,
+    project_vendor: String,
     release_version: String,
     sdk: Option<Image>,
+    // this ValidIdentifier is actually the vendor name
+    // Vendor is the registry
```
cbgbt: You've done a lot of wading through the concept of "vendor" in twoliter. Any chance you'd be willing to add a docstring comment or a README.md segment explaining how these concepts interact?

This comment here makes me wonder if we should be considering renaming some of these symbols (or using type aliases) to make it more clear what's going on.

twoliter/src/project/lock/mod.rs
```diff
@@ -116,8 +118,8 @@ impl LockedSDK {
 #[derive(Debug, Clone, Eq, Ord, PartialOrd, Serialize, Deserialize)]
 #[serde(rename_all = "kebab-case")]
 pub(crate) struct Lock {
-    /// The version of the Twoliter.toml this was generated from
-    pub schema_version: SchemaVersion<1>,
+    /// This schema version is seprately with the Twoliter.toml schema_version
```
cbgbt: This comment is helpful, but I think it's best to start docstring comments with an objective statement of what something is or it's behavior. you can then compare it to other programming constructs later.

e.g. you could say:

```rust
/// The supported version of the Twoliter.lock format.
///
/// This version is independent of the Twoliter.toml schema version.
```

twoliter/src/compatibility.rs
```diff
@@ -2,7 +2,9 @@
 //! own artifacts.
 
 /// Defines the exact supported schema version of Twoliter.toml supported by twoliter
-pub const SUPPORTED_TWOLITER_PROJECT_SCHEMA_VERSION: u32 = 1;
+pub const SUPPORTED_TWOLITER_PROJECT_SCHEMA_VERSION: u32 = 2;
```
cbgbt: We called this `SUPPORTED` when we only supported one version of the `Twoliter.toml` file. Does it make sense to change this to `PREFERRED` or `LATEST`?

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,553 @@
+//! Migration system for Twoliter project files.
+//!
+//! This module provides functionality to automatically migrate Twoliter.toml files between different schema versions,
+//! ensuring backward compatibility when new fields are added or the schema changes.
+//! Supports both forward (v1->v2) and backward (v2->v1) migrations.
```
cbgbt: Do we actually want backwards migrations? I think the case for backward migrations would be older versions of twoliter reading a new format; however, they won't have the benefit of knowing of the new format.

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,553 @@
...
+        registry.register(Box::new(v1_to_v2::V1ToV2Migrator::new()));
+
+        registry
+    }
+
+    /// Register a new migrator.
+    pub fn register(&mut self, migrator: Box<dyn ProjectMigrator>) {
+        self.migrators.push(migrator);
+    }
+
```
cbgbt: This is fine, but I think avoiding mutability can be a very powerful strategy. You might consider instead just making `new()` take the valid list of migrators:

```rust
pub fn new(migrators: Vec<Box<dyn ProjectMigrator>>) -> Self {
    Self { migrators }
}
```

You could then `impl Default` to get the builtin set, like this:

```rust
impl Default for MigrationRegistry {
    fn default() -> Self {
        MigrationRegistry::new(vec![
            Box::new(V1ToV2Migrator) as _, // you can use this `as _` to do the type coercion.
        ])
        
    }
}
```

Then for unit tests, you could test more intricate cases via `new()` directly.

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,553 @@
...
+    pub fn find_migrator(&self, from: u32, to: u32) -> Option<&dyn ProjectMigrator> {
+        self.migrators
+            .iter()
+            .find(|m| m.can_migrate(from, to))
+            .map(|m| m.as_ref())
+    }
+
+    pub fn can_migrate(&self, from: u32, to: u32) -> bool {
+        from == to || self.find_migrator(from, to).is_some()
+    }
```
cbgbt: I might refrain from making these `pub`. Seems like we only want callers to interact with `migrate_project`.

I might also change these to names like:

* `find_direct_migrator`
* `can_directly_migrate`

Since it's easy to read these function names and assume that they are referring to the broader step-by-step implementation below.

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,553 @@
...
+        if let Some(migrator) = self.find_migrator(from, to) {
+            let result = migrator
+                .migrate(content)
+                .with_context(|| format!("Failed to migrate from version {} to {}", from, to))?;
+            info!(
+                "Successfully migrated project from version {} to version {}",
+                from, to
+            );
+            return Ok(result);
+        }
```
cbgbt: This seems behaviorally equivalent to `migrate_step_by_step`, right? Unless I'm missing something, it's probably best for us to fall through directly to that implementation. Eliminating program branches can be a good way to mitigate the risk of bugs in "lesser-used" codepaths.

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,553 @@
...
+    /// Perform step-by-step migration when no direct path exists.
+    fn migrate_step_by_step(&self, content: &str, from: u32, to: u32) -> Result<String> {
+        let mut current_content = content.to_string();
+        let mut current_version = from;
+
+        if from < to {
+            // Forward migration (step by step)
+            while current_version < to {
+                let next_version = current_version + 1;
+                if let Some(migrator) = self.find_migrator(current_version, next_version) {
```
cbgbt: As a tip, you can avoid the `else { anyhow::bail!(...) }` branch by doing this:

```rust
let migrator = self.find_direct_migrator(current_version, next_version)
    .with_context(|| format!(
        "No migration path found from version {current_version} to version {next_version}"
    ))?;
```

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,553 @@
...
+    }
+
+    /// Perform step-by-step migration when no direct path exists.
+    fn migrate_step_by_step(&self, content: &str, from: u32, to: u32) -> Result<String> {
+        let mut current_content = content.to_string();
+        let mut current_version = from;
+
+        if from < to {
+            // Forward migration (step by step)
+            while current_version < to {
```
cbgbt: I think the opening `if` is redundant with this loop condition.

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,553 @@
...
+/// Trait for migrating project configurations between schema versions.
+pub trait ProjectMigrator: Send + Sync {
+    /// The source schema version this migrator handles.
+    fn current_version(&self) -> u32;
+
+    /// The target schema version this migrator produces.
+    fn to_version(&self) -> u32;
+
+    /// Perform the migration from source to target version.
+    fn migrate(&self, input: &str) -> Result<String>;
```
cbgbt: A weakness of this interface is that we're forced to de-serialize and re-serialize the incoming `Twoliter.toml` file every time we migrate.

One simple way to improve this would be for this method to take and emit a `toml::Table` directly:

```rust
migrate(&self, input: toml::Table) -> Result<toml::Table>;
```

---

It's more advanced Rust, but in general we prefer to work with strongly-modeled types rather than directly modifying "toml" or "json" types as is done in the v1->v2 migrator implementation you have here. One way to solve this could be to use pointers to fully-modeled types using [the `Any` type](https://doc.rust-lang.org/std/any/index.html), like this:

```rust
fn parse_source_version(&self, input: &str) -> Box<dyn Any>;
fn migrate(&self, input: &dyn Any) -> Result<Box<dyn Any>>;
```

Somewhere in this `twoliter::project::migrate` module, you'd need a representation of each supported Twoliter config type:

```rust
#[derive(Debug, Clone, Deserialize)]
struct ProjectV1 {
    schema_version: SchemaVersion<1>,
    release_version: String,
    sdk: Option<Image>,
    vendor: Option<BTreeMap<ValidIdentifier, Vendor>>,
    kit: Option<Vec<Image>>,
}

#[derive(Debug, Clone, Deserialize)]
struct ProjectV2 {
    schema_version: SchemaVersion<1>,
    release_version: String,
    sdk: Option<Image>,
    vendor: Option<BTreeMap<ValidIdentifier, Vendor>>,
    kit: Option<Vec<Image>>,
    project_vendor: String,
}
```

Then the v1_to_v2 could do something like this:

```rust
fn parse_source_version(&self, input: &str) -> Result<Box<dyn Any>> {
    let source_version: ProjectV1 = toml::from_str(input).context("yada yada")?;
    Ok(Box::new(source_version) as _)
}

fn migrate(&self, input: &dyn Any) -> Result<Box<dyn Any>> {
    let source: ProjectV1 = input.downcast_ref().context("Attempted to call v1->v2 migrator on something that is not ProjectV1")?;
    // Do the migration, make a ProjectV2 type
    // ...
    Ok(Box::new(project_v2_type) as _) // Erase the real type again
}
```

You could make it so that the `migrate_project` function was then generic over the return type:

```rust
fn migrate_project<Project>(&self, content: &str, from: u32, to: u32) -> Result<Project> 
    // Do the migrations between "Box<dyn Any>" types
    // ...

    // this assumes you have a "curr_project" which is a Box<dyn Any>
    let curr_project: Box<Project> = curr_project.downcast().context("Migrator returned incorrect project type")?;
    Ok(*curr_project)
}
```

I'd somewhat prefer we structure the migrator this way, as it would make it such that we were much more strict in recognizing a valid `v1` Twoliter.toml while also eliminating the need for us to continuously re-parse the intermediate objects from a string or from a generic `toml` type.

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,553 @@
...
+        .as_integer()
+        .context("schema-version field is not an integer")?;
+
+    Ok(version as u32)
+}
+
+/// Migrate project content to the target version.
+/// This is the main function used by the project loading system to automatically
+/// handle migration when loading Twoliter.toml files.
+pub fn migrate_project_content(content: &str, target_version: u32) -> Result<String> {
```
cbgbt: Per my comment above, regardless of how we change the migrator's intermediate representation, I think we should return something much more structured than `String` here.

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,553 @@
...
+        debug!(
+            "Project file is already at target version {}",
+            target_version
+        );
+        return Ok(content.to_string());
+    }
+
+    let registry = MigrationRegistry::new();
+
+    if !registry.can_migrate(current_version, target_version) {
```
cbgbt: It seems like you implemented a step-wise migration but this function only checks for first-degree migrations, and so will cause deeper ones not to run.

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,553 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // Mock migrator for testing
+    struct MockMigrator {
+        from: u32,
+        to: u32,
+        transform: fn(&str) -> Result<String>,
+    }
```
cbgbt: As a tip: use `mockall!` to dramatically simplify the creation of mock objects:
https://crates.io/crates/mockall

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,553 @@
...
+        }
+
+        fn migrate(&self, input: &str) -> Result<String> {
+            (self.transform)(input)
+        }
+    }
+
+    #[test]
+    fn test_project_migrator_can_migrate() {
+        let migrator = MockMigrator::new(1, 2, |s| Ok(s.to_string()));
```
cbgbt: Given the bug identified above, we should expand this test.

Consider using the `test_case` macro here, it lets you defined parameterized test cases on a given test function. There are other examples of its usage in the Twoliter repository.

twoliter/src/project/lock/mod.rs
sky1122: Yes, I created a new variable in [compatibility.rs](https://github.com/bottlerocket-os/twoliter/pull/551/files#diff-f37f92a6817efecff5354aa288b6a35b9b5c0fa3fb6e879d7b6f187c69c07d9a) `pub const SUPPORTED_TWOLITER_LOCK_SCHEMA_VERSION: u32 = 2;`  and in the `Lock` struct use that variable to bump the schema version
```
pub schema_version: SchemaVersion<SUPPORTED_TWOLITER_LOCK_SCHEMA_VERSION>,
```

twoliter/src/compatibility.rs
sky1122: I change to `LATEST` based on the slightly different between `PREFERRED` or `LATEST` and the current behavior is tend to be `LATEST`

twoliter/src/project/migrate/mod.rs
sky1122: That is exact my confusion. I think I confused by other terms. I left that comment because I understand for forward migration is let newer version of the twoliter to understand a old format, but not 100% sure why it doesn't need the backward. Thank you for explaining! I don't think we need the backward since user will update to get the new features, and forward also make sure older format won't effect.

twoliter/src/project/migrate/mod.rs
sky1122: I was thinking we can both do the migration from v1_to_v3.rs(directly adding changes target to version 1 to version 3) and also could migrate step by step. I will delete this logic branch and only follow the step by step implementation.

cbgbt: What you are saying makes sense, and I was wondering if that's why you did it.

When thinking about it, I realized that this code would only work if the first migration was a full hop, but wouldn't catch the case where you have to do something like:
* v1 -> v2
* v2 -> v4


In order to do this "properly" I think we'd need to do breadth-first search of the migration graph. But we could also refrain from complicating this solution until we need it, and only support incremental migrations for now.

sky1122: Thanks for the detail suggestion! yes, I will have these two method in the schema_version and make sure all the missing gap will have the migration. And change the `can_migrate` method to make sure that.
```
    /// Get sorted version (ascending order)
    pub(crate) fn get_sorted_versions() -> Vec<u32> {
        Self::iter()
            .map(|v| v.get())
            .collect()
    }

    /// Returns consecutive version pairs
    pub(crate) fn get_version_pairs(from:u32, to:u32) -> Vec<(u32, u32)> {
        let sorted_versions = Self::get_sorted_versions();
        let versions_in_range: Vec<(u32, u32)> = sorted_versions
            .iter()
            .zip(sorted_versions.iter().skip(1))
            .filter(|(&v1, &v2)| {
                v1 >= from && v2 <= to
            })
            .map(|(&v1, &v2)| (v1, v2))
            .collect();
            versions_in_range
    }
```

sky1122:
I have add a new dependency `mockall = "0.12"` that is why `cargo clippy --locked -- -D warnings --no-deps` failed. Will address that with other potential feedbacks.

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,586 @@
...
+        }
+    }
+}
+
+/// Detect the schema version of a project file without fully parsing it.
+pub fn detect_schema_version(content: &Box<dyn Any>) -> Result<u32> {
+    let content = content
+        .downcast_ref::<String>()
+        .context("Failed to downcast content to String")?
+        .as_str();
```
cbgbt: So far as I can tell, this does not need to use `Any` and should just take a string.

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,586 @@
...
+        .as_integer()
+        .context("schema-version field is not an integer")?;
+
+    Ok(version as u32)
+}
+
+/// Migrate project content to the target version.
+/// This is the main function used by the project loading system to automatically
+/// handle migration when loading Twoliter.toml files.
+pub fn migrate_project_content(content: Box<dyn Any>, target_version: u32) -> Result<Box<dyn Any>> {
```
cbgbt: I'd rather this weren't the public interface. The migrator can use `Any` internally -- that's fine. But I think the `pub` interface should be something more like this:


```rust
pub fn migrate_project_content(content: &str) -> Result<UnvalidatedProject> {
    // This calls a private function which does the dynamic migration behavior
    let migrated = migrate_project_content_to(content, LATEST_VERSION)?
        .downcast()
        .context("Failed to migrate Twoliter.toml to supported version")
}
```

In this way, we provide callers a strongly typed interface that takes a string and returns what they expect. The dynamic nature of the migrator becomes hidden by the public interface.


---

Something that could really help with this review would be to make it such that `mod.rs` only contains the public interface to the migrator, including the trait.

I think the specific implementation could be moved to a different submodule. This would make it much clearer what is exposed to the caller vs what is hidden.

twoliter/src/project/unvalidated_project_versions.rs
```diff
@@ -0,0 +1,31 @@
...
+    pub(crate) release_version: String,
+    pub(crate) sdk: Option<Image>,
+    pub(crate) vendor: Option<BTreeMap<ValidIdentifier, Vendor>>,
+    pub(crate) kit: Option<Vec<Image>>,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+pub(crate) struct UnvalidatedProjectV2 {
+    pub(crate) schema_version: SupportedSchemaVersion,
```
cbgbt: Can we keep it so that we use const generics for `schema_version`? This makes it impossible to construct this type unless the `schema_version` field is correct:

```
schema_version: SchemaVersion<2>
```

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,586 @@
...
+    // For string input, we need to parse it first using the appropriate migrator
+    if let Some(content_str) = content.downcast_ref::<String>() {
+        // Find the first migrator in the chain to parse the source version
+        let version_pairs =
+            SupportedSchemaVersion::get_version_pairs(current_version, target_schema_version);
+        if let Some((first_from, first_to)) = version_pairs.first() {
+            let migrator = registry
+                .find_migrator(*first_from, *first_to)
+                .context("No migration path available")?;
+
```
cbgbt: Can we hide this behavior behind something `MigrationRegistry::parse` so that readers don't need to know the details here?

An alternative would be to separate the `parse` functionality into a separate trait.

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,586 @@
...
+    fn default() -> Self {
+        MigrationRegistry::new(vec![Box::new(V1ToV2Migrator) as _])
+    }
+}
+
+impl MigrationRegistry {
+    /// Create a new migration registry with all available migrators.
+    pub fn new(migrators: Vec<Box<dyn ProjectMigrator>>) -> Self {
+        Self { migrators }
+    }
```
cbgbt: We discussed in person: We probably want to detect duplicated migration intervals here. (e.g. don't have two migrations starting from `1`, don't have gaps of versions greater than 1.

cbgbt (commented):
We talked in person but I'll try to summarize the discussion here so that we can revisit it when convenient:

* Consider keeping `mod.rs` as just the public API for `migrate` and moving implementation details to submodules (e.g. `registry.rs` and maybe `parsers.rs`?)
* We should probably go back to using `u32` to represent versions.
* We maybe don't need parsers as a separate abstraction if we hide the parse implementation within the `MigrationRegistry`
* Prefix to next point: `MigrationRegistry` is a graph represented by a collection of edges. Even though we don't do complex graph traversals, it's useful to think/talk about it like that.
* Right now the SupportedVersions enum ensures that input versions are valid, we decided that we could perhaps just ensure that there are graph nodes for each version in range [START_VERSION, END_VERSION].

sky1122:
force pushed draft changes to get feedbacks

twoliter/Cargo.toml
```diff
@@ -58,6 +58,7 @@ tar.workspace = true
 
 [dev-dependencies]
 test-case.workspace = true
+mockall = "0.12"
```
cbgbt: Use `mockall.workspace = true` here

twoliter/src/project/migrate/parser.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/// Detects the schema version from Any type
+// pub fn detect_schema_version_from_any(content: &Box<dyn Any>) -> Result<u32> {
+//     match content {
+//         box_any if box_any.is::<UnvalidatedProjectV1>() => Ok(1),
+//         box_any if box_any.is::<UnvalidatedProjectV2>() => Ok(2),
+//         _ => Err(anyhow!("Failed to detect schema version from content")),
+//     }
+// }
```
cbgbt: Should remove this.

twoliter/src/project/migrate/parser.rs
```diff
@@ -0,0 +1,75 @@
...
+/// Parses a TOML string into the appropriate project type based on schema version
+pub fn parse_project_toml(content: &str) -> Result<Box<dyn Any>> {
+    let version = detect_schema_version(&content).unwrap_or(0);
+
+    match version {
+        1 => {
+            let project: UnvalidatedProjectV1 =
+                toml::from_str(&content).context("Failed to parse TOML as UnvalidatedProjectV1")?;
+            Ok(Box::new(project))
+        }
```
cbgbt: nit: this is fine, but most folks look for constants like this near defined as `const`. In this case you'd need a `HashMap` or `BTreeMap` so you can't use `const` since those allocate, so you need `static`. You can use `LazyLock` for this: https://doc.rust-lang.org/beta/std/sync/struct.LazyLock.html

Older rust code (including twoliter) use `lazy_static!` so you might see that around, but LazyLock is the "new way".

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,745 @@
...
+use std::any::Any;
+
+use crate::{
+    compatibility::SUPPORTED_TWOLITER_LOCK_SCHEMA_VERSION,
+    project::migrate::{
+        migrator::MigrationRegistry,
+        parser::{detect_schema_version, parse_project_toml, UnvalidatedProject},
+    },
+};
+
```
cbgbt: Rust tip: Consider `pub use self::migrator::MigrationRegistry` to re-expose the type at the top level.

twoliter/src/project/migrate/registry.rs
```diff
@@ -0,0 +1,165 @@
...
+    any::Any,
+    collections::{HashMap, HashSet, VecDeque},
+};
+
+use crate::project::migrate::{
+    implementations::V1ToV2Migrator, migrator::MigrationRegistry, ProjectMigrator,
+};
+use anyhow::{anyhow, Context, Result};
+impl Default for MigrationRegistry {
+    fn default() -> Self {
```
cbgbt: It's a bit odd to define the implementation of `MigrationRegistry` separate to the struct definition. Mind putting them together?

twoliter/src/project/migrate/registry.rs
```diff
@@ -0,0 +1,165 @@
...
+        }
+    }
+
+    /// Register a new migrator.
+    pub fn register(&mut self, key: (u32, u32), migrator: Box<dyn ProjectMigrator>) {
+        self.migrators.insert(key, migrator);
+    }
+
+    /// Find a migrator that can handle the given version transition.
+    pub fn find_migrator(
```
cbgbt: If you ignore the implementation and only look at the signature and docstring, this function and `find_migration_path()` are hard to tell apart. So we should change the name for this (and also make it private if we can)

```suggestion
    /// Find a migrator that can handle the given version transition.
    fn find_migration_edge(
```

twoliter/src/project/migrate/registry.rs
```diff
@@ -0,0 +1,165 @@
...
+        from_version: u32,
+        to_version: u32,
+    ) -> Option<&dyn ProjectMigrator> {
+        self.migrators
+            .get(&(from_version, to_version))
+            .map(|m| m.as_ref())
+    }
+
+    /// Find the shortest migration path from source version to target version using BFS
+    pub fn find_migration_path(&self, source: u32, target: u32) -> Result<Vec<(u32, u32)>> {
```
cbgbt: I think `Option` fits more cleanly than `Result`

```suggestion
    pub fn find_migration_path(&self, source: u32, target: u32) -> Option<Vec<(u32, u32)>> {
```

twoliter/src/project/migrate/registry.rs
```diff
@@ -0,0 +1,165 @@
...
+        source: u32,
+        target: u32,
+    ) -> Result<Box<dyn Any>> {
+        if source == target {
+            return Ok(content);
+        }
+
+        let path = self.find_migration_path(source, target)?;
+
+        self.can_migrate(&path)?;
```
cbgbt: I think this check isn't necessary since the route was just returned internally in the `MigrationRegistry`.

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,745 @@
...
+            .map(|boxed| *boxed)
+            .map_err(|e| anyhow!("Failed to downcast: {:?}", e));
+    }
+
+    // Parse the content into the appropriate type based on current version
+    let parsed_content =
+        parser::parse_project_toml(content).context("Failed to parse project content")?;
+
+    // Create a migration registry with default migrators
+    let registry = MigrationRegistry::default();
```
cbgbt: MigrationRegistry should probably remain private, but it would be interesting to make `migrate_project_content` a thin wrapper around a private function which allows passing non-default registries. This would let you test your graph traversal against more complicated graphs ahead of time.

You could also just confine the graph traversal tests to `find_migration_path`, although you'll still want to add a test or two of this function to ensure that the downcasting logic is correct.

twoliter/src/project/migrate/parser.rs
```diff
@@ -0,0 +1,75 @@
...
+        toml::from_str(&content).map_err(|e| anyhow!("Failed to parse TOML content: {}", e))?;
+
+    let version = value
+        .get("schema-version")
+        .or_else(|| value.get("schema_version"))
+        .ok_or_else(|| anyhow!("Failed to detect schema version: missing schema-version field"))?;
+
+    let version = version
+        .as_integer()
+        .ok_or_else(|| anyhow!("Failed to detect schema version: not an integer"))?;
```
cbgbt: Nit (current code is fine). Serde can help a lot for things like this:

```rust
fn detect_schema_version(...) {
   #[derive(Deserialize)]
   struct MaybeVersionedProject {
       schema_version: i32,
   }

   let project: MaybeVersionedProject = toml::from_str(content).context("Failed to extract schema version")?;
   Ok(project.schema_version as u32)
}
```

twoliter/src/project/mod.rs
```diff
@@ -525,8 +514,12 @@ impl UnvalidatedProject {
         Ok(Project {
             filepath,
             project_dir: project_dir.clone(),
-            schema_version: self.schema_version,
+            schema_version: self
+                .schema_version
+                .try_into()
+                .expect("Schema version conversion failed"),
```
cbgbt: I think this can just be `self.schema_version` again.

cbgbt (commented):
This is great! Only a few minor things.

sky1122:
force pushed to run cargo clippy fix and cargo fmt

sky1122:
force pushed to make build happy

sky1122:
previous push a wrong branch. force pushed

twoliter/src/project/migrate/migrator.rs
```diff
@@ -0,0 +1,409 @@
...
+    /// Create a new migration registry with all available migrators.
+    pub fn new(migrators: Vec<Box<dyn ProjectMigrator>>) -> Self {
+        let mut registry_map = HashMap::new();
+
+        // Convert the Vec to a HashMap indexed by (source_version, target_version)
+        for migrator in migrators {
+            let key = (migrator.current_version(), migrator.to_version());
+            registry_map.insert(key, migrator);
+        }
+
```
cbgbt: nit: Use `HashMap::with_capacity()` when the sizes of the data is known. This prevents spurious allocations.

If you use the iterator API, this Rust does this for you for free:

```rust
let registry_map = migrators.iter()
    .map(|migrator| (migrator.current_version(), migrator.to_version()))
    .collect();
```

Here rust knows that the capacity should match the input collection, and you also don't need to specify types because it deduces the type later when it tries to create `Self`.

twoliter/src/project/migrate/migrator.rs
```diff
@@ -0,0 +1,409 @@
...
+        &self,
+        parent: &HashMap<u32, u32>,
+        source: u32,
+        target: u32,
+    ) -> Vec<(u32, u32)> {
+        let mut path = Vec::new();
+        let mut current = target;
+
+        while current != source {
+            let prev = *parent.get(&current).unwrap(); // Safe because we know the path exists
```
cbgbt: This is only safe because we assume `reconstruct_path` is always called by `find_migration_path`.

I might suggest defining `fn reconstruct_path` within `find_migration_path` to make it harder to pass the wrong preconditions to this function.

You could also change `unwrap()` to `expect()`:

```suggestion
            let prev = *parent.get(&current).expect("parents of the discovered path to be populated correctly"); // Safe because we know the path exists
```

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+    fn to_version(&self) -> u32;
+
+    /// Perform the migration from source to target version.
+    fn migrate(&self, input: &dyn Any) -> Result<Box<dyn Any>>;
+}
+
+pub fn migrate_project_content(content: &str) -> Result<UnvalidatedProject> {
+    // Create the default registry
+    let mut registry = MigrationRegistry::new(vec![]);
+    V1ToV2Migrator::register(&mut registry);
```
cbgbt: This should probably be defined using the `Default` trait, right? I thought this had been done already.

twoliter/src/project/migrate/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+        .map_err(|e| anyhow!("Failed to downcast: {:?}", e))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    
+    
+    
+
```
cbgbt: Odd newlines left here. Surprised `rustfmt` allowed this.

twoliter/src/project/migrate/parser.rs
```diff
@@ -0,0 +1,236 @@
...
+        Ok(Box::new(project))
+    });
+    map.insert(1, parser1);
+
+    let parser2: ProjectParser = Box::new(|content| {
+        let project: UnvalidatedProjectV2 =
+            toml::from_str(content).context("Failed to parse TOML as UnvalidatedProjectV2")?;
+        Ok(Box::new(project))
+    });
+    map.insert(2, parser2);
```
cbgbt: nit: You can use the `maplit` crate to define this in a more idiomatic way.

twoliter/src/project/migrate/parser.rs
```diff
@@ -0,0 +1,236 @@
...
+});
+
+/// Parses a TOML string into the appropriate project type based on schema version
+pub fn parse_project_toml(content: &str) -> Result<Box<dyn Any>> {
+    let version = detect_schema_version(content).unwrap_or(0);
+
+    if let Some(parser) = SCHEMA_VERSION_PARSERS.get(&version) {
+        parser(content)
+    } else {
+        Err(anyhow!("Unknown schema version: {}", version))
```
cbgbt: nit:

idiomatically, this should be:

```rust
anyhow::bail!(format!("Unknown schema version: {version}"));

File: twoliter/src/project/migrate/migrator.rs [hunk 44]
```diff
@@ -0,0 +1,409 @@
...
+            return Ok(content);
+        }
+
+        let path = self.find_migration_path(source, target).ok_or_else(|| {
+            anyhow!(
+                "Failed to find migration path from v{} to v{}",
+                source,
+                target
+            )
+        })?;
```
cbgbt: nit: the `context` methods are the idiomatic way to do this.

```suggestion
        let path = self.find_migration_path(source, target).with_context(||
            format!("Failed to find migration path from v{source} to v{target}")
        )?;
```

cbgbt (approved):
Well done.

sky1122:
force-pushed to change the commit order

jmt-lab (approved):

twoliter/src/project/migrate/mod.rs
sky1122: oh yes. I was self confused and delete it. Now I change back to default. Thank you for the explainning.

sky1122:
forced pushed(s) to address the feedback and also make the rust clippy happy

jmt-lab (approved):

cbgbt (approved):

PR #554, opened 2025-07-15
update cargo dist

jmt-lab:
**Description of changes:**
* Experiment to update to use latest cargo-dist and the zigbuild instead of cross

**Testing done:**

**Terms of contribution:**


.github/workflows/release.yml
```diff
@@ -1,214 +1,291 @@
...
 #
-# If there's a prerelease-style suffix to the version then the Github Release™
+# If there's a prerelease-style suffix to the version, then the release(s)
 # will be marked as a prerelease.
 on:
+  pull_request:
   push:
     tags:
-      - "*-?v[0-9]+*"
+      - '**[0-9]+.[0-9]+.[0-9]+*'
```
jmt-lab, line 45: I think we'll need to undo this change

PR #555, opened 2025-07-15
cargo: enable profile-hint-mostly-unused for AWS SDKs

cbgbt:
**Description of changes:**
This takes advantage of a newly-introduced nightly compile feature: https://blog.rust-lang.org/inside-rust/2025/07/15/call-for-testing-hint-mostly-unused/

The AWS SDKs are a substantial portion of compile times for twoliter. Enabling this new nightly feature allows cargo codegen specific dependencies "just-in-time" rather than up-front. This allows cargo to refrain from generating code for unused portions of an API surface.

On my machine this reduces release and debug build times by nearly a full minute.

### Release Builds Without Hints
without-hints

### Release Builds With Hints
with-hints

**Testing done:**
* [x] build and publish core-kit
* [x] build and publish aws-k8s-1.32 variants

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #557, opened 2025-07-16
fix: remove features to workaround cargo-dist bug

jmt-lab:
**Description of changes:**
* Removed the features restriction on pubsys and testsys that hit a bug in our fork of cargo-dist when calling build with feature flags

**Bug that was Hit**
We found that when our cargo-dist fork runs cargo build when you provide multiple features instead of adding the appropriate features list it adds a second --features argument leading to the error.

**Terms of contribution:**

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,6 @@ ci = ["github"]
 installers = []
 # Target platforms to build apps for (Rust target-triple syntax)
 targets = ["x86_64-unknown-linux-musl", "aarch64-unknown-linux-musl"]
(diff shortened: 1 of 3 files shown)


cbgbt (approved):
With #555 and the change to use "thin" LTO for `--release`, this seems fine to merge. These features were mostly set up as a mechanism to be able to avoid recompiling the AWS SDKs during development.

cbgbt:
Do you mind documenting the bug that we are circumventing in this PR description?

sumukhballal (approved):

Issue #450, opened 2025-02-04
Bug: Twoliter does not validate that the kit or sdk version = field in the lock matches the source url correctly


jmt-lab:
We had an occurence where a developer accidentally commited a manual edit of a Twoliter.lock file leading to the version field for the sdk entry to mismatch with the resolved source = "" uri. While this will not cause any issues as the actual object pulled when lockfile is used by twoliter will always be the source it adds ambiguity. 

Example human modification that put the lock file into invalid state:

```toml
schema-version = 1
kit = []

[sdk]
name = "bottlerocket-sdk"
version = "0.50.0"
vendor = "bottlerocket"
source = "public.ecr.aws/bottlerocket/bottlerocket-sdk:v0.50.1"
digest = "HEh3Lx3F6P4OEPnFubF++RMpMW2vlfp/Tc/tGjnBRcM="
``` 

We should make sure our validation when we load a lockfile ensures that the version -> source tag since pubsys will always publish kits currently with that tagging scheme.

PR #563, opened 2025-07-24
fix: a check that prevent inconsistent manual change in twoliter.lock

Sparksssj:
**Issue number:450**

Closes # https://github.com/bottlerocket-os/twoliter/issues/450

**Description of changes:**
Add a check to prevent manual change in twoliter.lock that causes inconsistency. Now the build would fail if there's such inconsistency

**Testing done:**
Manual tests done on both sdk and kit packages.

**Terms of contribution:**

### DIFF
--- a/twoliter/src/project/lock/mod.rs
+++ b/twoliter/src/project/lock/mod.rs
@@ -204,20 +204,18 @@ impl Lock {
         }
     }
 
+    /// Validates that the version and source URI of each locked image in the lockfile are consistent.
     fn validate_version_source_consistency(&self) -> Result<()> {
-        Self::validate_locked_image_consistency(&self.sdk, "SDK")?;
+        Self::validate_locked_image_consistency(&self.sdk)?;
 
-        for (index, kit) in self.kit.iter().enumerate() {
-            Self::validate_locked_image_consistency(
-                kit,
-                &format!("kit[{}] ({})", index, kit.name),
-            )?;
+        for kit in &self.kit {
+            Self::validate_locked_image_consistency(kit)?;
         }
 
         Ok(())
     }
 
-    fn validate_locked_image_consistency(image: &LockedImage, context: &str) -> Result<()> {
+    fn validate_locked_image_consistency(image: &LockedImage) -> Result<()> {
         let expected_tag = format!("v{}", image.version);
 
         let actual_tag = image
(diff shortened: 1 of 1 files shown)


twoliter/src/project/lock/mod.rs
```diff
@@ -200,6 +204,43 @@ impl Lock {
         }
     }
 
+    fn validate_version_source_consistency(&self) -> Result<()> {
```
jmt-lab, line 208: This looks good, but have you tested this with Twoliter.override files present just to make sure an autogenerated lockfile with a .override file present doesn't hit this falsely?

jmt-lab (approved):

twoliter/src/project/lock/mod.rs
cbgbt, line 208: Could you also add docstring comments to these functions, telling us what they validate? It would be great to read the function with an understanding of what it is guarding against.

twoliter/src/project/lock/mod.rs
```diff
@@ -200,6 +204,43 @@ impl Lock {
...
+            Self::validate_locked_image_consistency(
+                kit,
+                &format!("kit[{}] ({})", index, kit.name),
+            )?;
+        }
+
+        Ok(())
+    }
+
+    fn validate_locked_image_consistency(image: &LockedImage, context: &str) -> Result<()> {
```
cbgbt: We should find a different way to represent `context` that isn't strings.

Is there an existing enum type we can use to disambiguate SDK images from Kit images? Or perhaps we should just remove the `context` argument and rely on `LockedImage` having a reasonable implementation of `Display` (which it does!)

twoliter/src/project/lock/mod.rs
```diff
@@ -200,6 +204,43 @@ impl Lock {
...
+            .rsplit(':')
+            .next()
+            .context("source URI does not contain a tag (missing ':' separator)")?;
+
+        if actual_tag != expected_tag {
+            bail!(
+                "Version-source mismatch in lockfile for {}: \
+                version field is '{}' but source URI '{}' has tag '{}'. \
+                Expected source to end with ':{expected_tag}'. \
+                This usually happens when the lockfile was manually edited.",
```
sam-berning: nit: would be good to include a hint on the proper way to update the lockfile

twoliter/src/project/lock/mod.rs
Sparksssj: I had a test on that, there's no problem with .override file

qianxjcraig (approved):

PR #564, opened 2025-07-25
fix: krane using lazy_static! caused uncleaned temp files

### LINKED ISSUE #546: Twoliter leaks temp directories containing the `krane` binary
## Problem
Twoliter leaks temporary directories containing the `krane` binary that is shipped with the tool

## Cause
We [use `lazy_static!` to construct a temporary directory](https://github.com/bottlerocket-os/twoliter/blob/d00e650346a50da9fb25a521fc56edb1cd8562e7/tools/krane/src/lib.rs#L12-L20) where we write the `krane` binary.

```rust
lazy_static::lazy_static! {
    pub static ref KRANE: Krane = Krane::seal().unwrap();
}

#[derive(Debug)]
pub struct Krane {
    // Hold the file in memory to keep the fd open
    _tmp_dir: TempDir,
    path: PathBuf,
}
```

This static reference has little guarantee that the inner `Drop` implementation will be called on process exit, meaning that it's very easy for the tempfile to linger indefinitely.

## Proposed Solution

We could refrain from putting `krane` in a tempfile, and always write it to a defined location (like `./tools`). We could alternatively force the caller to tell us where to put it.

### Complications
Overwriting a binary that another process is running is problematic (see [this article about ETXTBSY](https://lwn.net/Articles/866493/).) This makes consistently writing to a known location problematic for concurrent twoliter invocations.

### Resolution
We can avoid concurrent run/modify issues of the binary by:
* Only writing `krane` under `./tools` if there is a checksum mismatch
* Writing `krane` to a tempfile under `./tools` and then `mv`-ing it into place so that we get a new inode rather than replacing an existing inode. This means that running processes with a reference to the old inode will continue to work.

Sparksssj:
**Issue number:546**

Closes #546 

**Description of changes:**
We began by analyzing a resource leak in Twoliter where temporary directories containing the krane binary (
/tmp/.tmp*/krane) were accumulating and not being cleaned up. We discarded lazy_static! approach which brings up the problem, and instead we extract the binary to a desired place (in tools directory) with atomic overwrite when needed.

**Testing done:**

Manual tests done

**Terms of contribution:**

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,6 @@
 [workspace]
 resolver = "2"
 
-default-members = ["twoliter"]
 members = [
     "tools/amispec",
     "tools/bottlerocket-variant",
--- a/tools/amispec/src/lib.rs
+++ b/tools/amispec/src/lib.rs
@@ -167,13 +167,16 @@ impl AmiSpec {
                 vec![
                     aws_sdk_ec2::types::TagSpecification::builder()
                         .resource_type(aws_sdk_ec2::types::ResourceType::Image)
-                        .set_tags(Some(
-                            tags.iter()
+                        .set_tags(Some({
+                            let mut sorted_tags: Vec<_> = tags.iter().collect();
+                            sorted_tags.sort_by_key(|(k, _)| *k);
+                            sorted_tags
+                                .into_iter()
                                 .map(|(k, v)| {
                                     aws_sdk_ec2::types::Tag::builder().key(k).value(v).build()
                                 })
-                                .collect(),
-                        ))
+                                .collect()
+                        }))
                         .build(),
                 ]
             }))
--- a/tools/include-env-compressed/tests/include_build_script.rs
+++ b/tools/include-env-compressed/tests/include_build_script.rs
@@ -4,11 +4,15 @@ const BUILD_SCRIPT_CONTENTS: &str = include_str!("../build.rs");
 
 #[test]
 fn test_include_build_script() {
-    // tests run in debug mode, so this will be uncompressed
     let build_script = include_env_compressed::include_archive_from_env!("MY_BUILD_SCRIPT");
     let included_build_script = std::io::read_to_string(build_script.reader()).unwrap();
     assert_eq!(included_build_script.trim(), BUILD_SCRIPT_CONTENTS.trim());
+
+    // The macro uses cfg(debug_assertions) to decide compression
+    #[cfg(debug_assertions)]
     assert_eq!(build_script.kind(), ArchiveKind::Uncompressed);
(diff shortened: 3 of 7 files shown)


jmt-lab (approved):

tools/krane/src/lib.rs
```diff
@@ -1,60 +1,106 @@
...
-    }
+        {
+            let mut temp_file = fs::File::create(&temp_path)?;
+            let permissions = fs::Permissions::from_mode(0o755);
+            temp_file.set_permissions(permissions)?;
+
+            let mut krane_reader = KRANE_BIN.reader();
+            std::io::copy(&mut krane_reader, &mut temp_file)?;
+
+            use std::io::Write;
```
sam-berning: is there some reason this `use` statement is mixed in with the rest of the code? I tend to prefer to put `use` statements at the beginning of the block that is using it to make it clear where it applies, since it will apply to the whole block

tools/oci-cli-wrapper/src/lib.rs
```diff
@@ -31,12 +31,12 @@ pub struct ImageTool {
 }
 
 impl ImageTool {
-    /// Uses the builtin `krane` provided by the `tools/krane` crate.
     pub fn from_builtin_krane() -> Self {
+        let tools_dir = std::path::Path::new("./build/tools");
```
sam-berning: this will create `build/tools` and install `krane` under whatever directory we run the command in, which isn't quite right - we want it installed under `build/tools` in the Twoliter project's root directory. you can find the right dir from within Twoliter if you have a `Project` by doing
```rust
let toolsdir = project.project_dir().join("build/tools");
```
such as [here](https://github.com/bottlerocket-os/twoliter/blob/develop/twoliter/src/cmd/build.rs#L56C9-L56C66). so maybe we can add an argument to this function that accepts a `tools_dir` - this should work for all of the Twoliter uses.

however, since we also use `ImageTool::from_builtin_krane()` in pubsys ([here](https://github.com/bottlerocket-os/twoliter/blob/develop/tools/pubsys/src/kit/publish_kit/mod.rs#L34)) this might be a bit tricky. maybe for now we can rely on the `TWOLITER_TOOLS_DIR` variable, since that should always be available when `pubsys publish kit` is invoked via twoliter?

sam-berning (changes requested):

tools/oci-cli-wrapper/src/lib.rs
Sparksssj: Yea you are right, updated!

tools/krane/src/lib.rs
Sparksssj: Yes this should be moved up.

Issue #421, opened 2024-12-19
slow start-up when not at AWS Seattle offices (I am in Europe)


mikn:
Hi! As mentioned during the community meeting, it takes me somewhere around 15-30 seconds before I get to the build stage. This is at a high level what happens:

### Combined Application and Network Timeline

| Time     | Type | Event/Message |
|----------|------|--------------|
| 21:32:35 | App  | Command started: `strace -t -f -e trace=network twoliter build variant metal-dev` |
| 21:32:35 | Net  | Socket: Initial connection to ECR (44.216.203.94:443) |
| 21:32:35 | Log  | INFO twoliter::project::lock - Resolving project references to check against lock file |
| 21:32:35 | Log  | INFO twoliter::project::lock::image - Resolving dependency image dependency 'bottlerocket-core-kit-5.0.0@public.ecr.aws/bottlerocket/bottlerocket-core-kit:v5.0.0' |
| 21:32:35 | Net  | Socket: Secondary ECR connection (44.216.203.94:443) |
| 21:32:36 | Net  | Socket: AWS service connection (75.2.101.78:443) |
| 21:32:36 | Net  | Socket: AWS service connection (99.83.145.10:443) |
| 21:32:37 | Net  | Socket: Additional ECR connection (44.216.203.94:443) |
| 21:32:37 | Net  | DNS: Failed NSCD lookup |
| 21:32:41 | Net  | IPv6: Multiple connection attempts failed (ENETUNREACH) |
| 21:32:41 | Log  | INFO twoliter::project::lock::image - Resolving dependency image dependency 'bottlerocket-kernel-kit-1.0.0@public.ecr.aws/bottlerocket/bottlerocket-kernel-kit:v1.0.0' |
| 21:32:42 | Net  | Socket: Additional AWS service connection (99.83.145.10:443) |
| 21:32:43 | Net  | Socket: AWS service connection (75.2.101.78:443) |
| 21:32:44 | Net  | Socket: Additional AWS service connection (99.83.145.10:443) |
| 21:32:45 | Net  | Socket: Additional ECR connection (44.216.203.94:443) |
| 21:32:46 | Net  | Socket: AWS service connection (75.2.101.78:443) |
| 21:32:47 | Net  | Socket: AWS service connection (99.83.145.10:443) |
| 21:32:48 | Net  | Socket: AWS service connection (3.164.226.167:443) |
[...]

PR #565, opened 2025-07-28
twoliter: always fetch kit deps before build tasks

cbgbt:
**Issue number:**

Related to  #421

**Description of changes:**
Currently, when building `bottlerocket-os/bottlerocket`, twoliter resolves dependencies twice on startup. Once when [fetching external kits](https://github.com/bottlerocket-os/bottlerocket/blob/54e53fdc7299cb0379c0e711b4a80ab62b4f47da/Makefile.toml#L405-L416), and then again just prior to the build. This is because the `Makefile.toml` file triggers the fetch to ensure that all dependencies are prepared locally prior to starting the build.

This change removes the requirement to explicitly fetch deps by just making twoliter do it on each build.

**Testing done:**
* Removed the dependency fetching step from `bottlerocket-os/bottlerocket` and performed image builds from a clean workspace.

**Terms of contribution:**

### DIFF
--- a/twoliter/src/cmd/build.rs
+++ b/twoliter/src/cmd/build.rs
@@ -53,7 +53,10 @@ impl BuildKit {
     pub(super) async fn run(&self) -> Result<()> {
         let project = project::load_or_find_project(self.project_path.clone()).await?;
         let project = project.load_lock::<Locked>().await?;
+        project.fetch(self.arch.as_str()).await?;
+
         let toolsdir = project.project_dir().join("build/tools");
+
         install_tools(&toolsdir).await?;
         let makefile_path = toolsdir.join("Makefile.toml");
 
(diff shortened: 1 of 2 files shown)


jmt-lab (approved):

sam-berning:
LGTM overall - but do we also need to remove the `fetch` dependencies from the `Makefile.toml` in this PR?

cbgbt:
> LGTM overall - but do we also need to remove the `fetch` dependencies from the `Makefile.toml` in this PR?

@sam-berning we do not -- the `fetch` task in Twoliter's makefile is still relevant. The task that we are making obsolete is the one in the `bottlerocket-os/bottlerocket` repository itself.

sam-berning (approved):

PR #577, opened 2025-09-05
feat(pubsys): add support for ami tags

### LINKED ISSUE #574: Twoliter pubsys should be able to tag AMIs after publish/copy
Some customers store additional image metadata in EC2 tags. Currently, the only way to update these tags is to replicate the behavior of the pubsys by assuming a role and calling EC2 CreateTags directly via scripts. The proposal here is to upstream this into the pubsys.

Customers could specify tags in Infra.toml per region or globally:

```toml
[aws]
regions = ["us-east-1", "us-west-2"]
tags = { storage-amount = 30}

[aws.regions.east-1.tags]
my-key = "my-value"

[aws.regions.us-west-2.tags]
WestCoast = "best coast!"
kernel-version = 6.1
```

Or they could specify tags when calling `cargo make ami` via the PUBLISH_AMI_TAGS environment variable:

```sh
cargo make -e PUBLISH_AMI_TAGS='[{"Key":"storage-amount","Value":30},{"Key":"WestCoast","Value":"best coast!"},{"Key":"kernel-version","Value":6.1}]' \
ami
```

Would produce EC2 tags like:

*us-east-1 AMI*
```json
[
    {
        "Key": "storage-amount",
        "Value": 30
    },
    {
        "Key": "my-key",
        "Value": "my-value"
    }
]
```

*us-west-2 AMI*
```json
[
  {
    "Key": "storage-amount",
    "Value": 30
  },
  {
    "Key": "WestCoast",
    "Value": "best coast!"
  },
  {
    "Key": "kernel-version",
    "Value": 6.1
  }
]

```

jmt-lab:
**Issue number:** #574 

Closes #574 

**Description of changes:**
Adds new section under aws-specific fields in Infra.toml for tags tags defined under tags.global will apply to all AMIs created, while tags created to aws.tags.region.<region> will only apply to that region

**Testing done:**

- Created a bottlerocket build and then created a Infra.toml with:
```toml
[aws]
regions = ["us-west-2", "us-east-1"]

[aws.tags.global]
tag0 = "foo"

[aws.tags.region.us-west-2]
tag1 = "bar"
```
- Ran pubsys publish ami through cargo make
- [x] us-west-2 ami has global tag0 = "foo"
- [x] us-east-1 ami has global tag0 = "foo"
- [x] us-west-2 ami has region specific tag1 = "bar"
- [x] us-east-1 ami DOES NOT have tag1 = "bar"

**Terms of contribution:**

### DIFF
--- a/tools/pubsys/src/aws/ami/mod.rs
+++ b/tools/pubsys/src/aws/ami/mod.rs
@@ -17,7 +17,7 @@ use crate::Args;
 use aws_sdk_ebs::Client as EbsClient;
 use aws_sdk_ec2::error::{ProvideErrorMetadata, SdkError};
 use aws_sdk_ec2::operation::copy_image::{CopyImageError, CopyImageOutput};
-use aws_sdk_ec2::types::OperationType;
+use aws_sdk_ec2::types::{OperationType, ResourceType, Tag, TagSpecification};
 use aws_sdk_ec2::{config::Region, Client as Ec2Client};
 use aws_sdk_sts::operation::get_caller_identity::{
     GetCallerIdentityError, GetCallerIdentityOutput,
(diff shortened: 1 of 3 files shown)


tools/pubsys/src/aws/ami/mod.rs
```diff
@@ -382,13 +390,27 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
...
             .set_source_image_id(Some(ids_of_image.image_id.clone()))
             .set_source_region(Some(base_region.as_ref().to_string()))
+            .set_tag_specifications(tags.map(|x| {
+                vec![TagSpecification::builder()
+                    .resource_type(ResourceType::Image)
+                    .set_tags(Some(
+                        x.iter()
+                            .map(|(key, value)| Tag::builder().key(key).value(value).build())
+                            .collect(),
+                    ))
```
bcressey, line 411: I'd prefer to do this via `amispec` if there's not a fundamental requirement for per-region tags.

If we do, there's the question of whether to treat this as an upsert - should we also copy any other tags the AMI might have, in addition to setting these new ones?

Seems like `validate-ami` would also need to be updated to do tag validation. There might be other tags on the AMI, but it should at least check that the tags in `Infra.toml` match the live values.

jmt-lab: Yeah, I talked with icf and regional tags aren't needed here so I think moving this to global only into amispec.toml is probably the right move here. I think because the primary tag creation is during publish-ami i think we don't need to worry about pre-existing tags as we are creating the ami ourself so there shouldn't be any. And i agree i think validate-ami should validate the tags

jmt-lab: So turns out, Sean already added tags to amispec.toml :D, we just haven't been using it yet. Will update to just use this.

PR #579, opened 2025-09-15
feat(pubsys): add support for ami tags

### LINKED ISSUE #574: Twoliter pubsys should be able to tag AMIs after publish/copy

jmt-lab:
**Issue number:** #574 

Closes #574 

**Description of changes:**
Enable use of tags in amispec.toml in pubsys

**Testing done:**

- Created a bottlerocket build and then created a amispec.toml with:
```toml
[tags]
tag0 = "foo"
tag1 = "bar"
```
- Ran pubsys publish ami through cargo make
- [x] us-west-2 ami has tag0 = "foo", tag1 = "bar"
- [x] us-east-1 ami has global tag0 = "foo", tag1 = "bar"

**Terms of contribution:**

### DIFF
--- a/tools/pubsys/src/aws/ami/mod.rs
+++ b/tools/pubsys/src/aws/ami/mod.rs
@@ -17,7 +17,7 @@ use crate::Args;
 use aws_sdk_ebs::Client as EbsClient;
 use aws_sdk_ec2::error::{ProvideErrorMetadata, SdkError};
 use aws_sdk_ec2::operation::copy_image::{CopyImageError, CopyImageOutput};
-use aws_sdk_ec2::types::OperationType;
+use aws_sdk_ec2::types::{OperationType, ResourceType, Tag, TagSpecification};
 use aws_sdk_ec2::{config::Region, Client as Ec2Client};
 use aws_sdk_sts::operation::get_caller_identity::{
     GetCallerIdentityError, GetCallerIdentityOutput,
@@ -208,13 +208,19 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
             "Registering '{}' in {}",
             tentative_amispec.name, base_region
         );
-        let new_ids = register_image(ami_args, &base_region, base_ebs_client, &base_ec2_client)
-            .await
-            .context(error::RegisterImageSnafu {
-                name: &tentative_amispec.name,
-                arch: &arch,
-                region: base_region.as_ref(),
-            })?;
+        let new_ids = register_image(
+            ami_args,
+            &base_region,
+            base_ebs_client,
+            &base_ec2_client,
+            tentative_amispec.tags.clone(),
+        )
+        .await
+        .context(error::RegisterImageSnafu {
+            name: &tentative_amispec.name,
+            arch: &arch,
+            region: base_region.as_ref(),
+        })?;
         info!(
             "Registered AMI '{}' in {}: {}",
             tentative_amispec.name, base_region, new_ids.image_id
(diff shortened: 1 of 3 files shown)


jmt-lab:
Continued from https://github.com/bottlerocket-os/twoliter/pull/577 which got accidentally canceled cause of incorrect git command

jmt-lab:
Sorry this isn't ready yet i need to add the validation logic to validate-ami

jmt-lab:
Actually it looks like we do not do any validation of anything amispec.toml related in validate-ami aso unsure if we should change it

tools/pubsys/src/aws/ami/register/mod.rs
```diff
@@ -48,6 +50,16 @@ async fn _register_image(
...
+            vec![TagSpecification::builder()
+                .resource_type(ResourceType::Image)
+                .set_tags(Some(
+                    x.iter()
+                        .map(|(key, value)| Tag::builder().key(key).value(value).build())
+                        .collect(),
+                ))
+                .build()]
+        }))
         .send_with(ec2_client)
```
cbgbt, line 63: This tag specification setting seems like it should be done as part of `as_register_image_call()` rather than separately here at the call site.

cbgbt (commented):
What would it look like to expose tagging of the underlying EBS snapshots as part of this effort?

I can't recall why this isn't already implemented!

tools/pubsys/src/aws/ami/register/mod.rs
cbgbt: Oh, it seems we actually already do that [here](https://github.com/bottlerocket-os/twoliter/blob/f10dec2f4c1b5fc0dc99e5954eb85f21fc014014/tools/amispec/src/lib.rs#L166)

tools/pubsys/src/aws/ami/mod.rs
```diff
@@ -382,13 +388,24 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
...
+            .set_tag_specifications(tentative_amispec.tags.as_ref().map(|x| {
+                vec![TagSpecification::builder()
+                    .resource_type(ResourceType::Image)
+                    .set_tags(Some(
+                        x.iter()
+                            .map(|(key, value)| Tag::builder().key(key).value(value).build())
+                            .collect(),
+                    ))
+                    .build()]
+            }))
```
cbgbt, line 408: @jmt pointed out that despite my other comments being true about the initial registration, we do still need to add code to handle tag copies during the image copy.

I think rather than manually inserting them based on the tentative ami spec (which would miss any additional logic handled during the rendering process) we can call `set_copy_image_tags()`

PR #583, opened 2025-09-23
feat: merge SBOM packages and remove old SBOM's

### LINKED ISSUE #584: Merge SBOM packages
I'd like for the twoliter script to merge all of the SBOM's generated by the `sbomtool generate` command into `image-spdx.json` and `image-cyclonedx.json` files. It should also remove all of the SBOM packages after merging them into the final JSON's. 

Currently, the SBOM packages are stored in `/usr/share/sboms/`, where each package in this directory has associated `spdx` and `cyclonedx` JSON SBOMs. Using the `sbomtool merge` command, these should be merged under a `/usr/share/bottlerocket/sbom/` directory, into `image-spdx.json` and `image-cyclonedx.json` files, respectively. After merging, the SBOMs in `/usr/share/sboms/` should also be removed.

cezar-r:
**Issue number:** #584

Closes #584

**Description of changes:**
Use `sbomtool` to merge SBOM packages into a single file per format

This change consolidates SBOM data from individual package SBOMs into unified, compressed 
files.

Changes:

- Added `metadata-sbom` subpackage to `metadata.spec` that provides a marker file for SBOM 
metadata
- Modified `rpm2img` to install SBOM subpackages into a separate mount point during image 
build
- Iterates through all installed packages and installs their corresponding `-sbom` subpackages
if available
- Uses `sbomtool` merge to combine all package SBOMs into single files for each format (`spdx` 
and `cyclonedx`)
- Compresses merged SBOMs with `lz4` and stores them in `/usr/share/bottlerocket/` on the root 
partition
- Copies compressed SBOMs to the output directory for external access
- Removes individual SBOM JSON files after merging to save space

**Testing done:**
- Built and mounted self-built AMI in build host. Verified the partition did not have a `/usr/share/sboms` directory but had a `usr/share/bottlerocket/sbom/` directory with `image-spdx.json.lz4` and `image-cyclonedx.json.lz4` files.
```
# aarch64
→ ls /mnt/br-ami/aarch64-bottlerocket-linux-gnu/sys-root/usr/share/
audit  bottlerocket  brush  dbus-1  eks  factory  i18n  info  iproute2  keyutils  licenses  locale  logdog.d  man  pci.ids  storewolf  templates  terminfo  updog  whippet  xfsprogs  zoneinfo

→ ll /mnt/br-ami/aarch64-bottlerocket-linux-gnu/sys-root/usr/share/bottlerocket/
total 196K
-rw-r--r--. 1 root root  63K Dec 11 03:29 application-inventory.json
-rw-r--r--. 1 root root 365K Dec 11 03:29 cyclonedx-sbom.json
-rw-r--r--. 1 root root   76 Dec 11 03:29 image-features.env
drwxr-xr-x. 3 root root   46 Dec 11 03:29 kernel-devel/
-rw-r--r--. 1 root root 1.4M Dec 11 03:29 spdx-sbom.json
```

```
# x86_64
→ ls /mnt/br-ami/x86_64-bottlerocket-linux-gnu/sys-root/usr/share
audit         brush   eks      i18n  iproute2  licenses  logdog.d  pci.ids    templates  updog    xfsprogs
bottlerocket  dbus-1  factory  info  keyutils  locale    man       storewolf  terminfo   whippet  zoneinfo

→  ll /mnt/br-ami/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/bottlerocket/
total 328K
-rw-r--r--. 1 root root  65K Dec 11 04:36 application-inventory.json
-rw-r--r--. 1 root root 612K Dec 11 04:36 cyclonedx-sbom.json
-rw-r--r--. 1 root root   76 Dec 11 04:36 image-features.env
drwxr-xr-x. 3 root root   46 Dec 11 04:36 kernel-devel/
-rw-r--r--. 1 root root 2.6M Dec 11 04:36 spdx-sbom.json
```

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/build.Dockerfile
+++ b/twoliter/embedded/build.Dockerfile
@@ -298,7 +298,7 @@ RUN --mount=target=/host \
       EXTERNAL_KIT_REPOS+=("--repofrompath=${REPO_NAME},${REPO_PATH}" --enablerepo "${REPO_NAME}"); \
     done && \
     echo "${EXTERNAL_KIT_REPOS[@]}" && \
-    mkdir -p /local/rpms && \
+    mkdir -p /local/rpms /local/sbom-rpms && \
     download_dir_opt="$(dnf install --help | grep -Fwq downloaddir && echo "--downloaddir /local/rpms" ||:)" && \
     dnf -y \
       --disablerepo '*' \
@@ -312,7 +312,8 @@ RUN --mount=target=/host \
       install \
         --downloadonly \
 	${download_dir_opt} \
-        $(printf "bottlerocket-%s\n" metadata ${PACKAGES}) && \
+        $(printf "bottlerocket-%s\n" metadata metadata-sbom ${PACKAGES}) && \
+    find /local/rpms -mindepth 2 -type f -name '*-sbom-*.rpm' -exec mv -t /local/sbom-rpms {} + && \
     find /local/rpms -mindepth 2 -type f -name '*.rpm' -exec mv -t /local/rpms {} + && \
     find /local/rpms -mindepth 1 ! -name '*.rpm' -delete && \
     rm /output && \
@@ -371,6 +372,7 @@ RUN --mount=target=/host \
     /host/build/tools/pipesys link --fd-socket "${OUTPUT_SOCKET}" --target /output && \
     /host/build/tools/rpm2img \
       --package-dir=/local/rpms \
+      --sbom-package-dir=/local/sbom-rpms \
       --output-dir=/output \
       --external-kits-path="/bypass/build/external-kits" \
       --output-fmt="${IMAGE_FORMAT}" \
--- a/twoliter/embedded/imghelper
+++ b/twoliter/embedded/imghelper
@@ -28,11 +28,18 @@ SBKEYS="${HOME}/sbkeys"
 declare -a SHIM_SIGN_KEY CODE_SIGN_KEY
 
 sanity_checks() {
-  local output_fmt partition_plan ovf_template uefi_secure_boot
+  local output_fmt partition_plan ovf_template uefi_secure_boot sbom_package_dir
   output_fmt="${1:?}"
   partition_plan="${2:?}"
   ovf_template="${3:?}"
   uefi_secure_boot="${4:?}"
+  sbom_package_dir="${5:?}"
+
+  if [[ ! -d "${sbom_package_dir}" ]]; then
+    echo "SBOM package directory does not exist: ${sbom_package_dir}" >&2
+    exit 1
+  fi
+
   case "${output_fmt}" in
   raw | qcow2 | vmdk) ;;
   *)
--- a/twoliter/embedded/metadata.spec
+++ b/twoliter/embedded/metadata.spec
@@ -69,12 +69,22 @@ Provides: %{_cross_os}image-feature(no-encrypted-storage)
 %description
 %{summary}.
 
+%package sbom
+Summary: SBOM metadata for Bottlerocket image
+Provides: %{_cross_os}image-metadata(sbom)
+
+%description sbom
+SBOM metadata subpackage for Bottlerocket image.
+
 %prep
 
 %build
 
 %install
+mkdir -p %{buildroot}%{_cross_datadir}/bottlerocket
 
 %files
+
+%files sbom
 
 %changelog
--- a/twoliter/embedded/rpm2img
+++ b/twoliter/embedded/rpm2img
@@ -30,6 +30,7 @@ for opt in "$@"; do
   --with-uefi-secure-boot=*) UEFI_SECURE_BOOT="${optarg}" ;;
   --with-in-place-updates=*) IN_PLACE_UPDATES="${optarg}" ;;
   --with-encrypted-storage=*) ENCRYPTED_STORAGE="${optarg}" ;;
+  --sbom-package-dir=*) SBOM_PACKAGE_DIR="${optarg}" ;;
   *)
     echo "unexpected arg: ${opt}" >&2
     exit 1
@@ -46,7 +47,7 @@ done
 . "${0%/*}/imghelper"
 
 sanity_checks \
-  "${OUTPUT_FMT}" "${PARTITION_PLAN}" "${OVF_TEMPLATE}" "${UEFI_SECURE_BOOT}"
+  "${OUTPUT_FMT}" "${PARTITION_PLAN}" "${OVF_TEMPLATE}" "${UEFI_SECURE_BOOT}" "${SBOM_PACKAGE_DIR}"
 
 # Store output artifacts in a versioned directory.
 OUTPUT_DIR="${OUTPUT_DIR}/${VERSION_ID}-${BUILD_ID}"
@@ -255,6 +256,41 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
 # Write the the inventory to a file in the local build output directory so that builders
 # can access the inventory without needed to dig into the generated image.
 printf "%s\n" "${INVENTORY_DATA}" >"${OUTPUT_DIR}/application-inventory.json"
+
+# Install SBOM packages
+SBOM_MOUNT=$(mktemp -d)
+
+# TODO: Eventually this needs to trigger an error also, but it might be too soon to break backwards compatibility.
+mapfile -t sbom_rpms < <(find "${SBOM_PACKAGE_DIR}" -maxdepth 1 -name "*.rpm" -type f)
+
+# There will always be a metadata SBOM package
+if [[ ${#sbom_rpms[@]} -gt 1 ]]; then
+  mkdir -p "${SBOM_MOUNT}/var/lib/rpm"
+  rpm -iv --ignorearch --root "${SBOM_MOUNT}" "${SBOM_PACKAGE_DIR}"/*.rpm
+  usr_dirs=(${SBOM_MOUNT}/*-linux-gnu/sys-root/usr)
+  usr_dir="${usr_dirs[0]}"
+  ln -rs "${usr_dir}" "${SBOM_MOUNT}/usr"
+fi
+
+# Merge SBOMs into a single json file
+KIT_SBOMS_DIR="${SBOM_MOUNT}/usr/share/sboms"
+if [ -d "${KIT_SBOMS_DIR}" ]; then
+  IMAGE_SBOM_DIR="${ROOT_MOUNT}/usr/share/bottlerocket"
+  mkdir -p "${IMAGE_SBOM_DIR}"
+  for format in "spdx" "cyclonedx"; do
+    image_sbom="${format}-sbom.json"
+    image_sbom_path="${IMAGE_SBOM_DIR}/${image_sbom}"
+    find "${KIT_SBOMS_DIR}" -name "*-${format}.json" -type f -exec sbomtool merge --output "${image_sbom_path}" {} \+
+    if [[ ! -f "${image_sbom_path}" ]]; then
+      echo "Could not find image SBOM path: ${image_sbom_path}" >&2
+      exit 1
+    fi
+    cp "${image_sbom_path}" "${OUTPUT_DIR}/${image_sbom}"
+    if [[ "${EROFS_ROOT_PARTITION}" == "no" ]]; then
+      lz4 -z -f "${image_sbom_path}" "${image_sbom_path}.lz4"
+    fi
+  done
+fi
 
 # Regenerate module dependencies, if possible.
 KMOD_DIR="${ROOT_MOUNT}/lib/modules"
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 TOP := $(dir $(abspath $(firstword $(MAKEFILE_LIST))))
 
-BOTTLEROCKET_SDK_VERSION ?= v0.60.0
+BOTTLEROCKET_SDK_VERSION ?= v0.66.0
 BOTTLEROCKET_SDK_IMAGE ?= public.ecr.aws/bottlerocket/bottlerocket-sdk:$(BOTTLEROCKET_SDK_VERSION)
 
 .PHONY: design
--- a/tests/projects/local-kit/Twoliter.toml
+++ b/tests/projects/local-kit/Twoliter.toml
@@ -4,7 +4,7 @@ release-version = "1.0.0"
 [sdk]
 name = "bottlerocket-sdk"
 vendor = "bottlerocket"
-version = "0.60.0"
+version = "0.66.0"
 
 [vendor.bottlerocket]
 # This test will fail if Twoliter overrides aren't working
--- a/tests/projects/project1/Twoliter.toml
+++ b/tests/projects/project1/Twoliter.toml
@@ -4,7 +4,7 @@ release-version = "1.0.0"
 [sdk]
 name = "bottlerocket-sdk"
 vendor = "bottlerocket"
-version = "0.60.0"
+version = "0.66.0"
 
 [vendor.bottlerocket]
 registry = "twoliter.alpha"


twoliter/embedded/rpm2img
```diff
@@ -256,6 +256,19 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
 # can access the inventory without needed to dig into the generated image.
 printf "%s\n" "${INVENTORY_DATA}" >"${OUTPUT_DIR}/application-inventory.json"
 
+# Merge SBOM's into a single json file
```
ginglis13: nit for consistency
```suggestion
# Merge SBOMs into a single json file
```

twoliter/embedded/rpm2img
```diff
@@ -256,6 +256,19 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+KIT_SBOMS_DIR="${ROOT_MOUNT}/usr/share/sboms"
+# Only process SBOMs if the directory exists.
```
ginglis13: this comment is self-explanatory by the line below

twoliter/embedded/rpm2img
```diff
@@ -256,6 +256,19 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+if [ -d "${KIT_SBOMS_DIR}" ]; then
+  IMAGE_SBOM_DIR="${ROOT_MOUNT}/usr/share/bottlerocket/sbom"
+  mkdir -p "${IMAGE_SBOM_DIR}"
+  for format in "spdx" "cyclonedx"; do
+    find "${KIT_SBOMS_DIR}" -name "*-${format}.json" -type f -exec sbomtool merge --output "${IMAGE_SBOM_DIR}/image-${format}.json" {} \+
```
ginglis13: IMO this name should be something more meaningful, maybe like `"${VERSION_ID}-${BUILD_ID}-${format}.json"`, but I'd also like to hear others' opinions

Additionally, it would be useful to do something similar to what we do for application inventory where we write the file out to `${OUTPUT_DIR}`, which helps for quicker iteration cycles

cezar-r: Made these changes, but awaiting others' opinions before resolving

ginglis13 (approved):

twoliter/embedded/rpm2img
```diff
@@ -256,6 +256,23 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
 # can access the inventory without needed to dig into the generated image.
 printf "%s\n" "${INVENTORY_DATA}" >"${OUTPUT_DIR}/application-inventory.json"
 
+# Merge SBOMs into a single json file
+KIT_SBOMS_DIR="${ROOT_MOUNT}/usr/share/sboms"
+if [ -d "${KIT_SBOMS_DIR}" ]; then
+  IMAGE_SBOM_DIR="${ROOT_MOUNT}/usr/share/bottlerocket/sbom"
```
bcressey:
```suggestion
  IMAGE_SBOM_DIR="${ROOT_MOUNT}/usr/share/bottlerocket"
```

twoliter/embedded/rpm2img
```diff
@@ -256,6 +256,23 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+  mkdir -p "${IMAGE_SBOM_DIR}"
+  for format in "spdx" "cyclonedx"; do
+    image_sbom="${VERSION_ID}-${BUILD_ID}-${format}.json"
```
bcressey: I don't love putting the version ID and build ID into the filename, since it makes them essentially random. Can we just create `spdx-sbom.json` and `cyclonedx-sbom.json`?

twoliter/embedded/rpm2img
```diff
@@ -256,6 +256,23 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+    image_sbom_path="${IMAGE_SBOM_DIR}/${image_sbom}"
+    find "${KIT_SBOMS_DIR}" -name "*-${format}.json" -type f -exec sbomtool merge --output "${image_sbom_path}" {} \+
+
+    # Write the inventory to a file in the local build output directory
+    cat "${image_sbom_path}" > "${OUTPUT_DIR}/${image_sbom}"
```
bcressey:
```suggestion
    # Write the inventory to a file in the local build output directory
    cp "${image_sbom_path}" "${OUTPUT_DIR}/${image_sbom}"
```

rpkelly (approved):

rpkelly:
nit: this seems less like a chore and more like a feature

twoliter/embedded/rpm2img
```diff
@@ -256,6 +256,50 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
 # can access the inventory without needed to dig into the generated image.
 printf "%s\n" "${INVENTORY_DATA}" >"${OUTPUT_DIR}/application-inventory.json"
 
+# Install SBOM subpackages for each installed RPM (only for EROFS images)
+if [[ "${EROFS_ROOT_PARTITION}" == "yes" ]]; then
```
bcressey: This restriction doesn't make much sense as build system policy - "we will only produce SBOM artifacts if you're using EROFS".

Options:
1. Drop it and do it unconditionally.
2. Add a different image feature flag that's related to SBOM.
3. Compress the SBOM files on disk, if that gives enough space?
4. Emit the build artifacts always, and only add to the image if we're using EROFS?

Not sure which is best, assuming you can't just always include SBOM artifacts.

twoliter/embedded/rpm2img
```diff
@@ -256,6 +256,50 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+  SBOM_RPMS=()
+  while IFS= read -r rpm_name; do
+    sbom_rpm="${rpm_name}-sbom"
+    if rpm -q --root "${ROOT_MOUNT}" "${sbom_rpm}" >/dev/null 2>&1; then
+      SBOM_RPMS+=("${sbom_rpm}")
+    else
```
bcressey: This is to handle the case where someone adds `foo-sbom` to their variant's package list, so we've installed it? That seems pretty weird.

twoliter/embedded/rpm2img
```diff
@@ -256,6 +256,50 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+      # Check if SBOM package file exists in local packages or external kits
+      sbom_file=$(find "${PACKAGE_DIR}" -maxdepth 1 -name "${sbom_rpm}*.rpm" -print -quit)
+      if [[ -z "${sbom_file}" && -n "${EXTERNAL_KITS_PATH}" ]]; then
+        sbom_file=$(find "${EXTERNAL_KITS_PATH}" -name "${sbom_rpm}*.rpm" -print -quit)
+      fi
+      if [[ -n "${sbom_file}" ]]; then
+        if rpm -iv --ignorearch --root "${ROOT_MOUNT}" "${sbom_file}" >/dev/null 2>&1; then
+          SBOM_RPMS+=("${sbom_rpm}")
+	      fi
```
bcressey: Need to fix your editor settings to not insert a mix of tabs/spaces. Please make sure this follows existing whitespace conventions.

twoliter/embedded/rpm2img
```diff
@@ -256,6 +256,50 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+    if rpm -q --root "${ROOT_MOUNT}" "${sbom_rpm}" >/dev/null 2>&1; then
+      SBOM_RPMS+=("${sbom_rpm}")
+    else
+      # Check if SBOM package file exists in local packages or external kits
+      sbom_file=$(find "${PACKAGE_DIR}" -maxdepth 1 -name "${sbom_rpm}*.rpm" -print -quit)
+      if [[ -z "${sbom_file}" && -n "${EXTERNAL_KITS_PATH}" ]]; then
+        sbom_file=$(find "${EXTERNAL_KITS_PATH}" -name "${sbom_rpm}*.rpm" -print -quit)
+      fi
+      if [[ -n "${sbom_file}" ]]; then
+        if rpm -iv --ignorearch --root "${ROOT_MOUNT}" "${sbom_file}" >/dev/null 2>&1; then
```
bcressey: We shouldn't install packages only to turn around and erase them. We can `rpm -iv` them into a separate directory, not the final root.

twoliter/embedded/rpm2img
```diff
@@ -256,6 +256,50 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+  SBOM_RPMS=()
+  while IFS= read -r rpm_name; do
+    sbom_rpm="${rpm_name}-sbom"
+    if rpm -q --root "${ROOT_MOUNT}" "${sbom_rpm}" >/dev/null 2>&1; then
+      SBOM_RPMS+=("${sbom_rpm}")
+    else
+      # Check if SBOM package file exists in local packages or external kits
+      sbom_file=$(find "${PACKAGE_DIR}" -maxdepth 1 -name "${sbom_rpm}*.rpm" -print -quit)
+      if [[ -z "${sbom_file}" && -n "${EXTERNAL_KITS_PATH}" ]]; then
+        sbom_file=$(find "${EXTERNAL_KITS_PATH}" -name "${sbom_rpm}*.rpm" -print -quit)
```
bcressey: This seems pretty hacky and I think we can do better:
1. Add an "sbom" subpackage to `metadata.spec`.
2. Have it `Provides: %{_cross_os}image-metadata(sbom)` (and `Requires: %{name}`).
3. Copy rpmdb from `${ROOT_MOUNT}` into new temp dir `${SBOM_MOUNT}`.
4. Install `metadata-sbom` into `${SBOM_MOUNT}` via `rpm -iv`.
5. For each of the RPM SBOM packages, add `Supplements: (%{_cross_os}image-metadata(sbom) if %{name})`.

That should give you a new temporary directory with only the SBOM files installed, which you can then concatenate together.

twoliter/embedded/rpm2img
cezar-r: Went with option 1 in new revision

twoliter/embedded/rpm2img
cezar-r: `rpm2img` doesn't have access to the DNF repo, so supporting this would require greater changes to the build process. Passing in the package list also has its own implications like making the computation more heavy. This is a lighter touch and has fewer risks with the build process.

twoliter/embedded/rpm2img
cezar-r: Fixed in latest revision

twoliter/embedded/rpm2img
cezar-r: Fixed, sorry about that 😬

twoliter/embedded/rpm2img
cezar-r: Fixed

twoliter/embedded/rpm2img
cezar-r: Circling back on this, I implemented the more robust DNF approach in the most recent revision. This solved a problem that wasn't immediately apparent in the previous version; the current version filters for packages by arch but the previous version generated SBOM's for packages on both arch's.

twoliter/embedded/metadata.spec
```diff
@@ -69,12 +69,25 @@ Provides: %{_cross_os}image-feature(no-encrypted-storage)
...
+%description sbom
+SBOM metadata subpackage for Bottlerocket image.
+
 %prep
 
 %build
 
 %install
+mkdir -p %{buildroot}%{_cross_datadir}/bottlerocket
+touch %{buildroot}%{_cross_datadir}/bottlerocket/.sbom-metadata
```
bcressey: We shouldn't need to create this file unless something needs it, we can just have an empty subpackage.

twoliter/embedded/rpm2img
```diff
@@ -256,6 +257,48 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+  IMAGE_SBOM_DIR="${ROOT_MOUNT}/usr/share/bottlerocket"
+  mkdir -p "${IMAGE_SBOM_DIR}"
+  for format in "spdx" "cyclonedx"; do
+    image_sbom="${format}-sbom.json"
+    image_sbom_path="${IMAGE_SBOM_DIR}/${image_sbom}"
+    find "${KIT_SBOMS_DIR}" -name "*-${format}.json" -type f -exec sbomtool merge --output "${image_sbom_path}" {} \+
+    if [[ -f "${image_sbom_path}" ]]; then
+      lz4 -z -f "${image_sbom_path}" "${image_sbom_path}.lz4"
+      rm "${image_sbom_path}"
+      cp "${image_sbom_path}.lz4" "${OUTPUT_DIR}/${image_sbom}.lz4"
```
bcressey: Let's copy out the uncompressed file unless it's really huge:
```suggestion
      cp "${image_sbom_path}" "${OUTPUT_DIR}/${image_sbom}"
```

twoliter/embedded/rpm2img
```diff
@@ -256,6 +257,48 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+KIT_SBOMS_DIR="${SBOM_MOUNT}/usr/share/sboms"
+if [ -d "${KIT_SBOMS_DIR}" ]; then
+  IMAGE_SBOM_DIR="${ROOT_MOUNT}/usr/share/bottlerocket"
+  mkdir -p "${IMAGE_SBOM_DIR}"
+  for format in "spdx" "cyclonedx"; do
+    image_sbom="${format}-sbom.json"
+    image_sbom_path="${IMAGE_SBOM_DIR}/${image_sbom}"
+    find "${KIT_SBOMS_DIR}" -name "*-${format}.json" -type f -exec sbomtool merge --output "${image_sbom_path}" {} \+
+    if [[ -f "${image_sbom_path}" ]]; then
+      lz4 -z -f "${image_sbom_path}" "${image_sbom_path}.lz4"
```
bcressey, line 290: Let's move this after the "copy out", and condition the compression on whether or not we're using EROFS.

twoliter/embedded/rpm2img
```diff
@@ -256,6 +257,48 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+# Merge SBOMs into a single json file
+KIT_SBOMS_DIR="${SBOM_MOUNT}/usr/share/sboms"
+if [ -d "${KIT_SBOMS_DIR}" ]; then
+  IMAGE_SBOM_DIR="${ROOT_MOUNT}/usr/share/bottlerocket"
+  mkdir -p "${IMAGE_SBOM_DIR}"
+  for format in "spdx" "cyclonedx"; do
+    image_sbom="${format}-sbom.json"
+    image_sbom_path="${IMAGE_SBOM_DIR}/${image_sbom}"
+    find "${KIT_SBOMS_DIR}" -name "*-${format}.json" -type f -exec sbomtool merge --output "${image_sbom_path}" {} \+
+    if [[ -f "${image_sbom_path}" ]]; then
```
bcressey: Ideally this would be a check that the file exists, and an `exit 1` with an error if it doesn't.

twoliter/embedded/rpm2img
```diff
@@ -256,6 +257,48 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
 # can access the inventory without needed to dig into the generated image.
 printf "%s\n" "${INVENTORY_DATA}" >"${OUTPUT_DIR}/application-inventory.json"
 
+# Install SBOM packages if directory provided
+SBOM_MOUNT=$(mktemp -d)
+trap 'rm -rf "${SBOM_MOUNT}"' EXIT
```
bcressey: This is handled by the existing `trap` which removes all of `/tmp`.

twoliter/embedded/rpm2img
```diff
@@ -256,6 +257,48 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+
+if [[ -n "${SBOM_PACKAGE_DIR:-}" ]] && [[ -d "${SBOM_PACKAGE_DIR}" ]]; then
```
bcressey: Move this to sanity checks, or omit if we're guaranteed that these directories exist. Then outdent the body of this branch.

twoliter/embedded/rpm2img
```diff
@@ -256,6 +257,48 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+  mapfile -t sbom_rpms < <(find "${SBOM_PACKAGE_DIR}" -maxdepth 1 -name "*.rpm" -type f)
+
+  if [[ ${#sbom_rpms[@]} -gt 0 ]]; then
+    mkdir -p "${SBOM_MOUNT}/var/lib/rpm"
+    rpm --root "${SBOM_MOUNT}" --initdb
+
+    CROSS_PREFIX=""
+    first_file=$(rpm -qpl "${sbom_rpms[0]}" 2>/dev/null | head -1)
```
bcressey: don't `... | head` - it will cause SIGPIPE errors sometimes

twoliter/embedded/rpm2img
```diff
@@ -256,6 +257,48 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+    if [[ "${first_file}" =~ ^(/[^/]+/sys-root)/ ]]; then
+      CROSS_PREFIX="${BASH_REMATCH[1]}"
+    fi
+
+    if [[ -n "${CROSS_PREFIX}" ]]; then
+      rpm -iv --ignorearch --badreloc --relocate "${CROSS_PREFIX}=/" --root "${SBOM_MOUNT}" "${sbom_rpms[@]}" >/dev/null 2>&1 ||:
+    else
+      rpm -iv --ignorearch --root "${SBOM_MOUNT}" "${sbom_rpms[@]}" >/dev/null 2>&1 ||:
+    fi
```
bcressey: For this logic - you can do this in two parts:
1. Install all rpms via `rpm -iv --ignorearch --root "${SBOM_MOUNT}"`
2. `ln -rs ${SBOM_MOUNT}/*-linux-gnu/sys-root/usr ${SBOM_MOUNT}/usr`

twoliter/embedded/rpm2img
```diff
@@ -256,6 +257,48 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
 # can access the inventory without needed to dig into the generated image.
 printf "%s\n" "${INVENTORY_DATA}" >"${OUTPUT_DIR}/application-inventory.json"
 
+# Install SBOM packages if directory provided
+SBOM_MOUNT=$(mktemp -d)
+trap 'rm -rf "${SBOM_MOUNT}"' EXIT
+
+if [[ -n "${SBOM_PACKAGE_DIR:-}" ]] && [[ -d "${SBOM_PACKAGE_DIR}" ]]; then
+  mapfile -t sbom_rpms < <(find "${SBOM_PACKAGE_DIR}" -maxdepth 1 -name "*.rpm" -type f)
```
bcressey: Eventually this needs to trigger an error also, but it might be too soon to break backwards compatibility.

twoliter/embedded/build.Dockerfile
```diff
@@ -313,8 +314,24 @@ RUN --mount=target=/host \
...
+      "${KIT_REPOS[@]}" \
+      "${EXTERNAL_KIT_REPOS[@]}" \
+      --nogpgcheck \
+      --forcearch "${ARCH}" \
+      --setopt=cachedir=/local/sbom-rpms \
+      install \
+        --downloadonly \
+        --skip-unavailable \
+	${sbom_download_dir_opt} \
+        $(printf "bottlerocket-%s-sbom\n" metadata ${PACKAGES}) ||: && \
```
bcressey: This won't get the `*-sbom` packages for dependencies of `${PACKAGES}`.

cezar-r: Fixed by adding `metadata-sbom` to the initial `dnf` command

twoliter/embedded/metadata.spec
cezar-r: Fixed

twoliter/embedded/rpm2img
cezar-r: Fixed

twoliter/embedded/rpm2img
cezar-r: Added a comment with this message

swagatbora90 (commented):
Please rename the CR from `chore` to `feat`.

twoliter/embedded/metadata.spec
```diff
@@ -69,12 +69,23 @@ Provides: %{_cross_os}image-feature(no-encrypted-storage)
...
 
 %install
+mkdir -p %{buildroot}%{_cross_datadir}/bottlerocket
 
 %files
 
+%files sbom
+
 %changelog
+
```
bcressey: nit:
```suggestion
```

twoliter/embedded/imghelper
```diff
@@ -33,6 +33,12 @@ sanity_checks() {
   partition_plan="${2:?}"
   ovf_template="${3:?}"
   uefi_secure_boot="${4:?}"
+
+  if [[ -n "${SBOM_PACKAGE_DIR:-}" ]] && [[ ! -d "${SBOM_PACKAGE_DIR}" ]]; then
```
bcressey: If you add the check here, you should take these variables as additional `local` arguments in the same way as the existing ones: declare the local then peel off the corresponding arg number using `:?` to trigger an error if it's undefined.

twoliter/embedded/rpm2img
```diff
@@ -256,6 +257,40 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+  mkdir -p "${IMAGE_SBOM_DIR}"
+  for format in "spdx" "cyclonedx"; do
+    image_sbom="${format}-sbom.json"
+    image_sbom_path="${IMAGE_SBOM_DIR}/${image_sbom}"
+    find "${KIT_SBOMS_DIR}" -name "*-${format}.json" -type f -exec sbomtool merge --output "${image_sbom_path}" {} \+
+    if [[ -f "${image_sbom_path}" ]]; then
+      if [[ "${EROFS_ROOT_PARTITION}" == "yes" ]]; then
+	cp "${image_sbom_path}" "${OUTPUT_DIR}/${image_sbom}"
+      else
+	lz4 -z -f "${image_sbom_path}" "${image_sbom_path}.lz4"
```
bcressey: tabs have made an unwelcome return!

twoliter/embedded/rpm2img
```diff
@@ -256,6 +257,40 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+# Install SBOM packages
+SBOM_MOUNT=$(mktemp -d)
+
+# TODO: Eventually this needs to trigger an error also, but it might be too soon to break backwards compatibility.
+mapfile -t sbom_rpms < <(find "${SBOM_PACKAGE_DIR}" -maxdepth 1 -name "*.rpm" -type f)
+if [[ ${#sbom_rpms[@]} -gt 0 ]]; then
+  mkdir -p "${SBOM_MOUNT}/var/lib/rpm"
+  rpm --root "${SBOM_MOUNT}" --initdb
+  rpm -iv --ignorearch --root "${SBOM_MOUNT}" "${SBOM_PACKAGE_DIR}"/*.rpm
+  ln -rs ${SBOM_MOUNT}/*-linux-gnu/sys-root/usr ${SBOM_MOUNT}/usr
```
bcressey:
```suggestion
  usr_dirs=(${SBOM_MOUNT}/*-linux-gnu/sys-root/usr)
  usr_dir="${usr_dirs[0]}"
  ln -rs "${usr_dir}" "${SBOM_MOUNT}/usr"
```

twoliter/embedded/rpm2img
```diff
@@ -256,6 +257,40 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+if [ -d "${KIT_SBOMS_DIR}" ]; then
+  IMAGE_SBOM_DIR="${ROOT_MOUNT}/usr/share/bottlerocket"
+  mkdir -p "${IMAGE_SBOM_DIR}"
+  for format in "spdx" "cyclonedx"; do
+    image_sbom="${format}-sbom.json"
+    image_sbom_path="${IMAGE_SBOM_DIR}/${image_sbom}"
+    find "${KIT_SBOMS_DIR}" -name "*-${format}.json" -type f -exec sbomtool merge --output "${image_sbom_path}" {} \+
+    if [[ -f "${image_sbom_path}" ]]; then
+      if [[ "${EROFS_ROOT_PARTITION}" == "yes" ]]; then
+	cp "${image_sbom_path}" "${OUTPUT_DIR}/${image_sbom}"
```
bcressey: Let's make the copy-out unconditional before the check, and just compress if we're doing not-EROFS. That way we always copy out an uncompressed file.

twoliter/embedded/build.Dockerfile
```diff
@@ -298,8 +298,9 @@ RUN --mount=target=/host \
       EXTERNAL_KIT_REPOS+=("--repofrompath=${REPO_NAME},${REPO_PATH}" --enablerepo "${REPO_NAME}"); \
     done && \
     echo "${EXTERNAL_KIT_REPOS[@]}" && \
-    mkdir -p /local/rpms && \
+    mkdir -p /local/rpms /local/sbom-rpms && \
     download_dir_opt="$(dnf install --help | grep -Fwq downloaddir && echo "--downloaddir /local/rpms" ||:)" && \
+    sbom_download_dir_opt="$(dnf install --help | grep -Fwq downloaddir && echo "--downloaddir /local/sbom-rpms" ||:)" && \
```
bcressey:
```suggestion
```

twoliter/embedded/build.Dockerfile
```diff
@@ -312,7 +313,9 @@ RUN --mount=target=/host \
       install \
         --downloadonly \
 	${download_dir_opt} \
-        $(printf "bottlerocket-%s\n" metadata ${PACKAGES}) && \
+        $(printf "bottlerocket-%s\n" metadata metadata-sbom ${PACKAGES}) && \
+    find /local/rpms -mindepth 2 -type f -name '*-sbom-*.rpm' -exec mv -t /local/sbom-rpms {} + && \
+    find /local/rpms -mindepth 2 -name '*-sbom-*.rpm' -delete && \
```
bcressey: this shouldn't be needed because we move out the matching rpms in the command above:
```suggestion
```

cbgbt: A bit sketchy to pattern match like this I think. What if a kit provides some kind of sbom utility.

Can we download these packages to a separate directory?

cbgbt: Oh, I misread this. We do.

twoliter/embedded/rpm2img
```diff
@@ -256,6 +257,40 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
 printf "%s\n" "${INVENTORY_DATA}" >"${OUTPUT_DIR}/application-inventory.json"
 
+# Install SBOM packages
+SBOM_MOUNT=$(mktemp -d)
+
+# TODO: Eventually this needs to trigger an error also, but it might be too soon to break backwards compatibility.
+mapfile -t sbom_rpms < <(find "${SBOM_PACKAGE_DIR}" -maxdepth 1 -name "*.rpm" -type f)
+if [[ ${#sbom_rpms[@]} -gt 0 ]]; then
+  mkdir -p "${SBOM_MOUNT}/var/lib/rpm"
+  rpm --root "${SBOM_MOUNT}" --initdb
```
bcressey: should be able to omit this also:
```suggestion
```

twoliter/embedded/metadata.spec
```diff
@@ -69,12 +69,22 @@ Provides: %{_cross_os}image-feature(no-encrypted-storage)
...
 
 %install
+mkdir -p %{buildroot}%{_cross_datadir}/bottlerocket
 
 %files
 
-%changelog
+%files sbom
+
+%changelog
```
bcressey: nit: you'll need to configure your editor to add newlines to the end of files

twoliter/embedded/rpm2img
```diff
@@ -256,6 +257,40 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+  mkdir -p "${IMAGE_SBOM_DIR}"
+  for format in "spdx" "cyclonedx"; do
+    image_sbom="${format}-sbom.json"
+    image_sbom_path="${IMAGE_SBOM_DIR}/${image_sbom}"
+    find "${KIT_SBOMS_DIR}" -name "*-${format}.json" -type f -exec sbomtool merge --output "${image_sbom_path}" {} \+
+    if [[ -f "${image_sbom_path}" ]]; then
+      cp "${image_sbom_path}" "${OUTPUT_DIR}/${image_sbom}"
+      if [[ "${EROFS_ROOT_PARTITION}" == "no" ]]; then
+        lz4 -z -f "${image_sbom_path}" "${image_sbom_path}.lz4"
+        cp "${image_sbom_path}.lz4" "${OUTPUT_DIR}/${image_sbom}.lz4"
```
bcressey: Can omit this part - we already have the uncompressed copy:
```suggestion
```

twoliter/embedded/rpm2img
```diff
@@ -256,6 +257,40 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+      fi
+    else
+      exit 1
+    fi
```
bcressey: preferred style is to invert the earlier conditional and early return/exit:
```
if [[ ! -f "${image_sbom_path}" ]]; then
  echo some error message >&2
  exit 1
fi
<outdent branch body above>
```

twoliter/embedded/rpm2img
```diff
@@ -256,6 +257,40 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+
+# Merge SBOMs into a single json file
+KIT_SBOMS_DIR="${SBOM_MOUNT}/usr/share/sboms"
+if [ -d "${KIT_SBOMS_DIR}" ]; then
+  IMAGE_SBOM_DIR="${ROOT_MOUNT}/usr/share/bottlerocket"
+  mkdir -p "${IMAGE_SBOM_DIR}"
+  for format in "spdx" "cyclonedx"; do
+    image_sbom="${format}-sbom.json"
+    image_sbom_path="${IMAGE_SBOM_DIR}/${image_sbom}"
+    find "${KIT_SBOMS_DIR}" -name "*-${format}.json" -type f -exec sbomtool merge --output "${image_sbom_path}" {} \+
```
bcressey, line 283: How long have we had a "good" version of `sbomtool` that works as expected? Do we need to guard this on whether `sbomtool` exists in the SDK?

cezar-r: If `sbomtool` isn't available, this branch won't be reached because `sbomtool` creates the `KIT_SBOMS_DIR="${SBOM_MOUNT}/usr/share/sboms"` dir. 

The `sbomtool` has been in the last 2 SDK versions, having been released in October

twoliter/embedded/rpm2img
```diff
@@ -256,6 +257,42 @@ printf "%s\n" "${INVENTORY_DATA}" >"${ROOT_MOUNT}/usr/share/bottlerocket/applica
...
+    if [[ ! -f "${image_sbom_path}" ]]; then
+      echo "Could not find image SBOM path: ${image_sbom_path}" >&2
+      exit 1
+    else
+      cp "${image_sbom_path}" "${OUTPUT_DIR}/${image_sbom}"
+      if [[ "${EROFS_ROOT_PARTITION}" == "no" ]]; then
+        lz4 -z -f "${image_sbom_path}" "${image_sbom_path}.lz4"
+      fi
+    fi
```
bcressey: This else should be removed and the block should be outdented because we'll exit if we take the first branch.

twoliter/embedded/metadata.spec
```diff
@@ -69,12 +69,22 @@ Provides: %{_cross_os}image-feature(no-encrypted-storage)
 %description
 %{summary}.
 
+%package sbom
+Summary: SBOM metadata for Bottlerocket image
+Provides: %{_cross_os}image-metadata(sbom)
+
+%description sbom
+SBOM metadata subpackage for Bottlerocket image.
```
ginglis13, line 77: nit: 
```suggestion
%{summary}.
```

twoliter/embedded/metadata.spec
```diff
@@ -69,12 +69,22 @@ Provides: %{_cross_os}image-feature(no-encrypted-storage)
...
 %prep
 
 %build
 
 %install
+mkdir -p %{buildroot}%{_cross_datadir}/bottlerocket
 
 %files
 
+%files sbom
```
ginglis13, line 88: Should we declare the installed dir?

```suggestion
%files sbom
%dir %{_cross_datadir}/bottlerocket
```

ginglis13 (approved):

bcressey (approved):

PR #589, opened 2025-11-05
add encrypted storage feature and deprecate older features

bcressey:
**Issue number:**
Related: https://github.com/bottlerocket-os/bottlerocket/issues/4680

**Description of changes:**
Add an image feature for encrypted storage.

My plan is for this to cover both `/local` (using dm-crypt) and `/.bottlerocket/datastore` (using fscrypt). I'll post the related core kit changes soon.

Deprecate two image features that are no longer useful.

`aws-ecs-1` was the last variant where GRUB's `$private` variable wasn't set in the earliest version, and also the last one to use `wicked` instead of `systemd-networkd`. Since `aws-ecs-1` is gone, and `wicked` is gone, there's no need to keep these features around.

**Testing done:**
Confirmed that I could add the encrypted storage feature to a variant and trigger the install of supporting software.

Verified that software that requires the networkd image feature continues to install.

**Terms of contribution:**

### DIFF
--- a/tools/buildsys/src/manifest.rs
+++ b/tools/buildsys/src/manifest.rs
@@ -828,6 +828,7 @@ const EXPERIMENTAL_IMAGE_FEATURES: &[&ImageFeature] = &[];
 
 const DEPRECATED_IMAGE_FEATURES: &[&ImageFeature] = &[
     &ImageFeature::GrubSetPrivateVar,
+    &ImageFeature::SystemdNetworkd,
 ];
 
 impl TryFrom<String> for ImageFeature {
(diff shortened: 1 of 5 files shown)


tools/buildsys/src/manifest.rs
```diff
@@ -821,6 +826,8 @@ pub enum ImageFeature {
 
 const EXPERIMENTAL_IMAGE_FEATURES: [&ImageFeature; 0] = [];
 
+const DEPRECATED_IMAGE_FEATURES: [&ImageFeature; 2] = [];
```
cbgbt: You can avoid having to specify the size of the array by making this a const reference:

```suggestion
const DEPRECATED_IMAGE_FEATURES: &[&ImageFeature] = [];
```

cbgbt (approved):
Only a minor style nit. iirc this code is consistent with what exists today.

lgtm!

cbgbt (approved):

ginglis13 (approved):

bcressey:
⬆️ force push fixes a tiny whitespace nit

arnaldo2792 (approved):

PR #591, opened 2025-11-07
chore: update dependencies to latest

jmt-lab:
**Description of changes:**
* Ran `cargo update`
* Bumped major dependencies and ran build

**Terms of contribution:**


tools/pipesys/src/cmd/mod.rs
```diff
@@ -95,17 +97,24 @@ fn fetch_fd(socket: &str) -> Result<i32> {
         .filter(|fd| **fd >= MIN_FD)
         .with_context(|| format!("did not receive valid file descriptor from socket {socket}"))?;
 
-    let dupfd =
-        duplicate_fd(*fd).with_context(|| format!("failed to duplicate file descriptor {fd}"))?;
+    // SAFETY: The filedescriptor is received over uds socket and as such has to be assumed as
+    // borrowed.
```
cbgbt: I did a bit of reading on this -- the FD received over the UDS is actually duplicated by the kernel and thus we own it.

Likewise the integral return value from `duplicate_fd` is also technically a new Owned FD.

We can probably call `OwnedFd::from_raw_fd()` on the descriptor read over the wire (in an unsafe block with the justification that should probably refer to the `SCM_RIGHTS` section of [this man page](https://man7.org/linux/man-pages/man7/unix.7.html)) and likewise it might be a good idea for `duplicate_fd` to return one, too.

clarify.toml
```diff
@@ -26,6 +26,7 @@ expression = "MIT OR Apache-2.0"
 license-files = [
     { path = "LICENSE-APACHE", hash = 0x24b54f4b },
     { path = "LICENSE-MIT", hash = 0xc97e30bd },
+    { path = "assets/images/LICENSE.md", hash = 0x7c815b17 },
```
cbgbt: This should go into `skip-files` since we do not distribute these assets.

deny.toml
```diff
@@ -62,6 +62,8 @@ wildcards = "deny"
 skip = [
     # several dependencies are using multiple versions of base64
     { name = "base64" },
+    # governor and guppy collid bringing two versions of hashbrown
```
cbgbt:
```suggestion
    # governor and guppy collide bringing two versions of hashbrown
```

bcressey (approved):

cbgbt (approved):

cbgbt:
Hmm, I made a mistake when switching to `OwnedFd` by casting it to a raw_fd on output, causing `Drop` to execute and closing the file. Fix incoming.

cbgbt (approved):

sam-berning (approved):

PR #595, opened 2025-11-10
krane: update ecr-login to v0.11.0

sam-berning:
**Issue number:**

Closes #

**Description of changes:**

Update ecr-login to v0.11.0. Also bump the base version of krane to v0.20.6 and drop the unneeded patches, since all of the dependency versions in those patches have now been taken by upstream krane, see: https://github.com/google/go-containerregistry/blob/v0.20.6/cmd/krane/go.mod.

**Testing done:**

Built twoliter, and ensured that the resulting krane is able to get credentials for EUSC regions using

```
echo 123456789012.dkr.ecr.eusc-de-east-1.amazonaws.eu | krane auth get
```

**Terms of contribution:**


Sparksssj (approved):

jmt-lab (approved):

sam-berning:
^ updated to use go-containerregistry from git archives instead of from the release artifacts, which makes many of the changes to `build.rs` from the previous revision unnecessary, and should fix the build failure.

cbgbt (approved):
lgtm assuming the build succeeds

PR #604, opened 2025-11-26
feat: add check-advisories to check BRSA fields

piyush-jena:
**Description of changes:**
  * Validate CVE ID/GHSA ID in the Rust crate. Moves implementation of this (#467) into the Rust crate.
  * Validate BRSA's package-metadata and make sure the included package is at a higher version.
  * Make sure End of Life BRSA is present if the BRSA package doesn't exist.

How it works?
We implement a `advisory-checker` rust crate that takes 2 arguments - The packages and advisories directory. The binary is called during build by the upper level build.Dockerfile. Since, its already in the sdk context and the rpm macros are loaded, we can use `rpmspec` and `rpm --eval` macros to extract package name and versions. We deserialize the advisory files to extract the expected version and then compare the Epoch, Version and Release to their respective packages to find out if we are in a newer version and if package doesn't exist, we make sure that the last BRSA for the package is an End of Life BRSA.

**Testing done:**
Cropping output for brevity. 

Testing it on `bottlerocket-core-kit`. I removed the recently published EOL BRSAs and changed the epoch for an `aws-iam-authenticator` BRSA. This is the error that's generated.
```
  #10 ERROR: process "/bin/sh -c if [ -d \"/host/advisories/\" ]; then         /host/build/tools/advisory-checker             --packages-dir /host/packages/             --advisories-dir /host/advisories/;     else         echo \"Advisories directory not found, skipping check.\";     fi" did not complete successfully: exit code: 1
  ------
   > [rpmbuild 4/7] RUN --mount=target=/host     if [ -d "/host/advisories/" ]; then         /host/build/tools/advisory-checker             --packages-dir /host/packages/             --advisories-dir /host/advisories/;     else         echo "Advisories directory not found, skipping check.";     fi:
  2.684 Error: 3 Advisory violations found.
  2.684 
  2.684 Advisory violations found:
  2.684 Advisory Product 'libexpat' has no package and latest advisory 'BRSA-ja1ptrrpkcey' is not end-of-life
  2.684 BRSA ID: BRSA-ng7athla9aog
  2.684  package name: aws-iam-authenticator
  2.684  package version: 1:0.7.10
  2.684  min. expected version: 2:0.7.5
  2.684 Advisory Product 'containerd-2.0' has no package and latest advisory 'BRSA-s6xothqqu5vw' is not end-of-life
 ```
 
 No issues when run on latest `bottlerocket`, `bottlerocket-core-kit` and `bottlerocket-kernel-kit`.

**Terms of contribution:**

### DIFF
--- a/tests/integration-tests/src/advisory_checker.rs
+++ b/tests/integration-tests/src/advisory_checker.rs
@@ -1,0 +1,210 @@
+use crate::run_command;
+use std::fs::{create_dir, create_dir_all, remove_dir, write};
+use std::process::Output;
+use tempfile::TempDir;
+
+const ADVISORY_CHECKER_PATH: &str = env!("CARGO_BIN_FILE_ADVISORY_CHECKER");
+const PACKAGES_DIR: &str = "packages";
+const ADVISORIES_DIR: &str = "advisories";
+
+fn create_spec(name: &str, version: &str) -> String {
+    format!(
+        r#"Name: {name}
+Version: {version}
+Release: 1
+Summary: Test package
+License: MIT
+
+%description
+Test package
+"#
+    )
+}
+
+fn create_advisory(id: &str, cve: &str, severity: &str, pkg: &str, version: &str) -> String {
+    format!(
+        r#"[advisory]
+id = "{id}"
+title = "Test Advisory"
+cve = "{cve}"
+severity = "{severity}"
+description = "Test vulnerability"
+
+[[advisory.products]]
+package-name = "{pkg}"
+patched-version = "{version}"
+patched-epoch = "0"
+
+[updateinfo]
+issue-date = 2026-01-01
+version = "1"
+"#
+    )
+}
+
+// This function creates a dummy kit. As input we provide packages as tuples
+// (package name, version) and advisories as a tuple (advisory path, advisory toml).
+// The output is the temporary directory where the kit is created.
+fn create_test_kit(packages: &[(&str, &str)], advisories: &[(&str, &str)]) -> TempDir {
+    let dir = TempDir::new().unwrap();
+
+    let packages_dir = dir.path().join(PACKAGES_DIR);
+    let advisories_dir = dir.path().join(ADVISORIES_DIR);
+    create_dir(&packages_dir).unwrap();
+    create_dir(&advisories_dir).unwrap();
+
+    for (name, version) in packages {
+        let pkg_dir = packages_dir.join(name);
+        create_dir_all(&pkg_dir).unwrap();
+        let content = create_spec(name, version);
+        write(pkg_dir.join("test.spec"), content).unwrap();
+    }
+
+    for (filename, content) in advisories {
+        let path = advisories_dir.join(filename);
+        create_dir_all(path.parent().unwrap()).unwrap();
+        write(&path, content).unwrap();
+    }
+
+    dir
+}
+
+fn run_advisory_checker(kit: &TempDir) -> Output {
+    run_command(
+        ADVISORY_CHECKER_PATH,
+        [
+            "--packages-dir",
+            kit.path().join(PACKAGES_DIR).to_str().unwrap(),
+            "--advisories-dir",
+            kit.path().join(ADVISORIES_DIR).to_str().unwrap(),
+        ],
+        [],
+    )
+}
+
+#[test]
+#[ignore]
+fn test_missing_advisories_dir_succeeds() {
+    // Given a kit with a package but no advisories directory
+    let kit = create_test_kit(&[("testpkg", "1.0.0")], &[]);
+    remove_dir(kit.path().join(ADVISORIES_DIR)).unwrap();
+
+    // When we run the advisory checker
+    let output = run_advisory_checker(&kit);
+
+    // Then it succeeds and reports that advisory checks were skipped
+    assert!(output.status.success());
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    assert!(stdout.contains("skipping"));
+}
+
+#[test]
+#[ignore]
+fn test_empty_advisories_dir_succeeds() {
+    // Given a kit with a package and an empty advisories directory
+    let kit = create_test_kit(&[("testpkg", "1.0.0")], &[]);
+
+    // When we run the advisory checker
+    let output = run_advisory_checker(&kit);
+
+    // Then it succeeds with no violations
+    assert!(output.status.success());
+}
+
+#[test]
+#[ignore]
+fn test_ignores_non_toml_files() {
+    // Given a kit whose advisories directory contains only non-toml files
+    let kit = create_test_kit(&[("testpkg", "1.0.0")], &[("staging/.gitkeep", "")]);
+
+    // When we run the advisory checker
+    let output = run_advisory_checker(&kit);
+
+    // Then it succeeds since non-toml files are ignored
+    assert!(output.status.success());
+}
+
+#[test]
+#[ignore]
+fn test_advisory_violation_fails() {
+    // Given a package at version 1.0.0 and an advisory requiring 2.0.0
+    let advisory = create_advisory("BRSA-test123", "CVE-2024-12345", "high", "testpkg", "2.0.0");
+    let kit = create_test_kit(
+        &[("testpkg", "1.0.0")],
+        &[("v01/BRSA-test.toml", &advisory)],
+    );
+
+    // When we run the advisory checker
+    let output = run_advisory_checker(&kit);
+
+    // Then it fails and reports the advisory violation
+    assert!(!output.status.success());
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    assert!(stdout.contains("Advisory violations found"));
+}
+
+#[test]
+#[ignore]
+fn test_advisory_satisfied_succeeds() {
+    // Given a package at version 3.0.0 and an advisory requiring 2.0.0
+    let advisory = create_advisory(
+        "BRSA-test456",
+        "CVE-2024-67890",
+        "moderate",
+        "testpkg",
+        "2.0.0",
+    );
+    let kit = create_test_kit(
+        &[("testpkg", "3.0.0")],
+        &[("v01/BRSA-test.toml", &advisory)],
+    );
+
+    // When we run the advisory checker
+    let output = run_advisory_checker(&kit);
+
+    // Then it succeeds
+    assert!(output.status.success());
+}
+
+#[test]
+#[ignore]
+fn test_advisory_for_removed_package_fails() {
+    // Given an advisory for "testpkg" but only "otherpkg" exists in the kit
+    let advisory = create_advisory(
+        "BRSA-test789",
+        "CVE-2024-11111",
+        "critical",
+        "testpkg",
+        "2.0.0",
+    );
+    let kit = create_test_kit(
+        &[("otherpkg", "1.0.0")],
+        &[("v01/BRSA-test.toml", &advisory)],
+    );
+
+    // When we run the advisory checker
+    let output = run_advisory_checker(&kit);
+
+    // Then it fails because the advisory's package has no end-of-life marker
+    assert!(!output.status.success());
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    assert!(stdout.contains("not end-of-life"));
+}
+
+#[test]
+#[ignore]
+fn test_parse_advisory_error_invalid_toml() {
+    // Given a kit with an advisory file containing invalid TOML
+    let kit = create_test_kit(
+        &[("testpkg", "1.0.0")],
+        &[("v01/BRSA-invalid.toml", "not valid toml {{{")],
+    );
+
+    // When we run the advisory checker
+    let output = run_advisory_checker(&kit);
+
+    // Then it fails with a parse error
+    assert!(!output.status.success());
+    let stderr = String::from_utf8_lossy(&output.stderr);
+    assert!(stderr.contains("Failed to parse advisory"));
+}
--- a/tests/integration-tests/src/appinventory.rs
+++ b/tests/integration-tests/src/appinventory.rs
@@ -18,7 +18,7 @@ const EXPECTED_INVENTORY_PATH: &str =
 
 // Last Bottlerocket release that doesn't include source packages in application inventory.
 // For older releases, we check that reference inventory is a subset of current.
-const SOURCE_PACKAGE_INVENTORY_ANCHOR_VERSION: &str = "1.55.0";
+const SOURCE_PACKAGE_INVENTORY_ANCHOR_VERSION: &str = "1.56.0";
 
 #[derive(Serialize, Deserialize)]
 pub struct GithubRelease {
--- a/tools/advisory-checker/Cargo.toml
+++ b/tools/advisory-checker/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "advisory-checker"
+version = "0.1.0"
+edition = "2021"
+license = "Apache-2.0 OR MIT"
+publish = false
+exclude = ["README.md"]
+
+[dependencies]
+nutype = { workspace = true, features = ["regex", "serde"] }
+regex.workspace = true
+serde = { workspace = true, features = ["derive"] }
+clap = { workspace = true, features = ["derive"] }
+snafu.workspace = true
+toml.workspace = true
+rpm.workspace = true
+
+[dev-dependencies]
+test-case.workspace = true
+
+[lints]
+workspace = true
--- a/tools/advisory-checker/src/error.rs
+++ b/tools/advisory-checker/src/error.rs
@@ -1,0 +1,44 @@
+use snafu::Snafu;
+use std::path::PathBuf;
+use std::process::Output;
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub), module)]
+pub enum Error {
+    #[snafu(display("Failed to execute command: {}", source))]
+    ExecutionFailure { source: std::io::Error },
+
+    #[snafu(display("Command '{}' failed: {}", bin_path, String::from_utf8_lossy(&output.stderr)))]
+    CommandFailure { bin_path: String, output: Output },
+
+    #[snafu(display("Failed to read '{}'", path.display()))]
+    ReadFile {
+        path: PathBuf,
+        source: std::io::Error,
+    },
+
+    #[snafu(display("Failed to parse advisory '{}'", path.display()))]
+    ParseAdvisory {
+        path: PathBuf,
+        source: toml::de::Error,
+    },
+
+    #[snafu(display(
+        "Incorrect output format for rpmspec command. Expected name|epoch|version. Got '{spec_output}'"
+    ))]
+    RpmSpecFormat { spec_output: String },
+
+    #[snafu(display("Failed to read directory '{}'", path.display()))]
+    ReadDir {
+        path: PathBuf,
+        source: std::io::Error,
+    },
+
+    #[snafu(display("Failed to read directory entry"))]
+    ReadDirEntry { source: std::io::Error },
+
+    #[snafu(display("{} Advisory violations found.", violations.len()))]
+    AdvisoryViolations { violations: Vec<String> },
+}
--- a/tools/advisory-checker/src/main.rs
+++ b/tools/advisory-checker/src/main.rs
@@ -1,0 +1,359 @@
+//! Validates rpmspec files against security advisories to ensure packages meet
+//! minimum version requirements for known vulnerabilities.
+
+mod error;
+mod models;
+
+use clap::Parser;
+use error::{error::*, Result};
+use models::{Advisory, Args, Epoch, PackageName, PackageVersionManifest, EV};
+use rpm::rpm_evr_compare;
+use snafu::{ensure, ResultExt};
+use std::collections::HashMap;
+use std::ffi::OsStr;
+use std::fs;
+use std::path::Path;
+use std::process::Command;
+
+// This is the last time when we removed a package from the kits without an
+// EOL advisory. We make sure that any advisory published post this date
+// either has a corresponding package or an EOL advisory.
+const LAST_UNTRACKED_PACKAGE_REMOVAL_DATE: &str = "2025-07-26";
+
+fn command<I, S>(bin_path: &str, args: I) -> Result<String>
+where
+    I: IntoIterator<Item = S>,
+    S: AsRef<OsStr>,
+{
+    let mut cmd = Command::new(bin_path);
+    cmd.args(args);
+    let output = cmd.output().context(ExecutionFailureSnafu)?;
+
+    ensure!(
+        output.status.success(),
+        CommandFailureSnafu { bin_path, output }
+    );
+
+    Ok(String::from_utf8_lossy(&output.stdout).to_string())
+}
+
+fn get_rpm_metadata(spec_path: &Path) -> Result<PackageVersionManifest> {
+    let pkg_prefix = command("rpm", ["--eval", "%{_cross_os}"])?
+        .trim()
+        .to_string();
+    let spec_path_str = spec_path.to_string_lossy();
+
+    let rpm_metadata = command(
+        "rpmspec",
+        [
+            "-q",
+            "--qf",
+            "%{Name}|%{Epoch}|%{Version}\n",
+            &spec_path_str,
+        ],
+    )?;
+
+    parse_rpm_metadata(pkg_prefix, rpm_metadata)
+}
+
+fn parse_rpm_metadata(pkg_prefix: String, rpm_metadata: String) -> Result<PackageVersionManifest> {
+    let mut manifest = PackageVersionManifest::new();
+    for metadata in rpm_metadata.lines() {
+        let parts: Vec<&str> = metadata.split('|').collect();
+        ensure!(
+            parts.len() == 3,
+            RpmSpecFormatSnafu {
+                spec_output: metadata
+            }
+        );
+
+        let name = parts[0].strip_prefix(&pkg_prefix).unwrap_or(parts[0]);
+        let epoch_str = if parts[1] == "(none)" { "0" } else { parts[1] };
+        let epoch = Epoch::try_new(epoch_str).expect("epoch should be valid integer");
+        let version = parts[2];
+
+        manifest.insert(PackageName::new(name), EV::new(&epoch, version));
+    }
+
+    Ok(manifest)
+}
+
+fn find_spec_file(package_dir: &Path) -> Option<std::path::PathBuf> {
+    fs::read_dir(package_dir).ok()?.find_map(|entry| {
+        let path = entry.ok()?.path();
+        if path.extension() == Some(OsStr::new("spec")) {
+            Some(path)
+        } else {
+            None
+        }
+    })
+}
+
+fn build_all_package_metadata(packages_dir: &Path) -> Result<PackageVersionManifest> {
+    let mut manifest = PackageVersionManifest::new();
+    if !packages_dir.exists() {
+        return Ok(manifest);
+    }
+
+    for entry in fs::read_dir(packages_dir).context(ReadDirSnafu { path: packages_dir })? {
+        let entry = entry.context(ReadDirEntrySnafu)?;
+        if entry.path().is_dir() {
+            // Inside the package directory, we extract metadata from the spec file
+            if let Some(spec_path) = find_spec_file(&entry.path()) {
+                let pkg_manifest = get_rpm_metadata(&spec_path)?;
+                manifest.merge(pkg_manifest);
+            }
+        }
+    }
+
+    Ok(manifest)
+}
+
+fn collect_advisories(advisories_dir: &Path) -> Result<HashMap<PackageName, Vec<Advisory>>> {
+    let mut advisories_by_product: HashMap<PackageName, Vec<Advisory>> = HashMap::new();
+
+    if !advisories_dir.exists() {
+        return Ok(advisories_by_product);
+    }
+
+    for entry in fs::read_dir(advisories_dir).context(ReadDirSnafu {
+        path: advisories_dir,
+    })? {
+        let version_dir = entry.context(ReadDirEntrySnafu)?;
+        if version_dir.path().is_dir() {
+            // Inside the version directory, we read the advisories and store them
+            for brsa in fs::read_dir(version_dir.path()).context(ReadDirSnafu {
+                path: advisories_dir,
+            })? {
+                let brsa = brsa.context(ReadDirEntrySnafu)?;
+                if brsa.path().extension() == Some(OsStr::new("toml")) {
+                    let content = fs::read_to_string(brsa.path())
+                        .context(ReadFileSnafu { path: brsa.path() })?;
+                    let advisory: Advisory = toml::from_str(&content)
+                        .context(ParseAdvisorySnafu { path: brsa.path() })?;
+                    for product in &advisory.advisory_info.products {
+                        advisories_by_product
+                            .entry(product.package_name.clone())
+                            .or_default()
+                            .push(advisory.clone());
+                    }
+                }
+            }
+        }
+    }
+
+    Ok(advisories_by_product)
+}
+
+fn validate_advisories(advisories_dir: &Path, packages_dir: &Path) -> Result<()> {
+    // We go through all the packages in package directory and create a map with
+    // package as key and Epoch:Version as value. We create a separate map from
+    // the advisories with package as key and an array of advisories as value.
+    let pkg_metadata = build_all_package_metadata(packages_dir)?;
+    let advisories_by_product = collect_advisories(advisories_dir)?;
+
+    let mut violations = Vec::new();
+
+    // For each advisory, we check if the corresponding package exists and the
+    // package is patched or if its an end of life advisory.
+    for (package_name, mut advisories) in advisories_by_product {
+        advisories.sort_by(|a, b| a.update_info.cmp(&b.update_info));
+
+        if let Some(spec_ev) = pkg_metadata.get(&package_name) {
+            // Package exists - check version compliance
+            for advisory in &advisories {
+                for product in &advisory.advisory_info.products {
+                    if product.package_name != package_name {
+                        continue;
+                    }
+                    let advisory_ev = EV::new(&product.patched_epoch, &product.patched_version);
+                    if rpm_evr_compare(&spec_ev.to_string(), &advisory_ev.to_string()).is_lt() {
+                        violations.push(format!(
+                            "BRSA ID: {}\n package name: {}\n package version: {spec_ev}\n min. expected version: {advisory_ev}",
+                            advisory.advisory_info.id, package_name
+                        ));
+                    }
+                }
+            }
+        } else {
+            // Package missing - check end-of-life
+            if let Some(latest) = advisories.last() {
+                let issue_date = latest.update_info.issue_date.to_string();
+
+                // If an advisory is created after LAST_UNTRACKED_PACKAGE_REMOVAL_DATE we make sure that
+                // either a package exists in the kit for that advisory or there is an end of life
+                // advisory. If this test fails, it indicates that there is a typo in the package name.
+                if issue_date.as_str() > LAST_UNTRACKED_PACKAGE_REMOVAL_DATE
+                    && !latest.advisory_info.end_of_life
+                {
+                    violations.push(format!(
+                        "Advisory Product '{}' has no package and latest advisory '{}' is not end-of-life",
+                        package_name, latest.advisory_info.id
+                    ));
+                }
+            }
+        }
+    }
+
+    if !violations.is_empty() {
+        println!("Advisory violations found:\n{}", violations.join("\n"));
+        return Err(error::Error::AdvisoryViolations { violations });
+    }
+
+    println!("All advisories validated. No errors were found.");
+    Ok(())
+}
+
+#[snafu::report]
+fn main() -> Result<()> {
+    let args = Args::parse();
+    if !args.advisories_dir.exists() || !args.advisories_dir.is_dir() {
+        println!(
+            "Advisories directory '{}' not found, skipping advisory check",
+            args.advisories_dir.display()
+        );
+        return Ok(());
+    }
+    validate_advisories(&args.advisories_dir, &args.packages_dir)?;
+    println!(
+        "Advisory validation passed for {}",
+        args.advisories_dir.display()
+    );
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use models::CveId;
+    use test_case::test_case;
+
+    fn make_advisory_toml(id_field: &str, id_value: &str) -> String {
+        format!(
+            r#"
+[advisory]
+id = "TEST-001"
+title = "Test Advisory"
+severity = "high"
+description = "Test description"
+{id_field} = "{id_value}"
+[[advisory.products]]
+package-name = "test"
+patched-version = "1.0"
+
+[updateinfo]
+issue-date = 2024-01-01
+version = "1"
+"#
+        )
+    }
+
+    #[test_case("CVE-2024-1234"; "basic 4 digit")]
+    #[test_case("CVE-2024-12345"; "5 digit")]
+    #[test_case("CVE-2025-472688"; "6 digit")]
+    #[test_case("CVE-1999-0001"; "old year")]
+    fn valid_cve(cve: &str) {
+        let toml = make_advisory_toml("cve", cve);
+        assert!(
+            toml::from_str::<Advisory>(&toml).is_ok(),
+            "Failed to parse valid CVE: {cve}"
+        );
+    }
+
+    #[test_case("CVE-2024-001"; "too few digits")]
+    #[test_case("CVE-2024-INVALID"; "non numeric")]
+    #[test_case("CVE-24-1234"; "year too short")]
+    #[test_case("CVE-2024-0"; "single digit")]
+    #[test_case("CVE–2024–1234"; "en dash")]
+    #[test_case("CVE—2024—1234"; "em dash")]
+    #[test_case("GHSA-xxxx-xxxx-xxxx"; "wrong format")]
+    fn invalid_cve(cve: &str) {
+        let toml = make_advisory_toml("cve", cve);
+        assert!(
+            toml::from_str::<Advisory>(&toml).is_err(),
+            "Should reject invalid CVE: {cve}"
+        );
+    }
+
+    #[test_case("GHSA-23fg-6c23-wxrv"; "standard")]
+    #[test_case("GHSA-2222-3333-4444"; "all numeric")]
+    #[test_case("GHSA-cfgh-jmpq-rvwx"; "all alpha")]
+    fn valid_ghsa(ghsa: &str) {
+        let toml = make_advisory_toml("ghsa", ghsa);
+        assert!(
+            toml::from_str::<Advisory>(&toml).is_ok(),
+            "Failed to parse valid GHSA: {ghsa}"
+        );
+    }
+
+    #[test_case("GHSA-123-456-789"; "too short")]
+    #[test_case("GHSA-12345-67890-12345"; "too long")]
+    #[test_case("CVE-2024-1234"; "wrong format")]
+    fn invalid_ghsa(ghsa: &str) {
+        let toml = make_advisory_toml("ghsa", ghsa);
+        assert!(
+            toml::from_str::<Advisory>(&toml).is_err(),
+            "Should reject invalid GHSA: {ghsa}"
+        );
+    }
+
+    #[test]
+    fn complete_advisory() {
+        let toml = make_advisory_toml("cve", "CVE-2025-12345");
+        let advisory: Advisory = toml::from_str(&toml).expect("Failed to parse complete advisory");
+        assert_eq!(advisory.advisory_info.id, "TEST-001");
+        assert_eq!(
+            advisory.advisory_info.cve,
+            CveId::try_new("CVE-2025-12345").ok()
+        );
+        assert_eq!(advisory.advisory_info.products.len(), 1);
+        assert_eq!(
+            advisory.advisory_info.products[0].package_name,
+            PackageName::new("test")
+        );
+        assert_eq!(
+            advisory.advisory_info.products[0].patched_epoch,
+            Epoch::try_new("0").unwrap()
+        );
+    }
+
+    #[test_case("testpkg|0|1.0.0\n"; "Epoch present")]
+    #[test_case("testpkg|(none)|1.0.0\n"; "Epoch not present")]
+    #[test_case("testpkg|(none)|1.0.0\ntestpkg-bin|(none)|1.0.0\ntestpkg-lib|(none)|1.0.0\n"; "Multiple sub-packages")]
+    fn parse_rpm_metadata_basic(rpm_metadata: &str) {
+        let output = rpm_metadata.to_string();
+        let result = parse_rpm_metadata("".to_string(), output).unwrap();
+        let evr = result.get(&PackageName::new("testpkg")).unwrap();
+        assert_eq!(evr.to_string(), "0:1.0.0");
+    }
+
+    #[test_case("testpkg\n"; "invalid output 1")]
+    #[test_case("testpkg|(none)\n"; "invalid output 2")]
+    fn parse_invalid_rpm_metadata(rpm_metadata: &str) {
+        let result = parse_rpm_metadata("".to_string(), rpm_metadata.to_string());
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn parse_rpm_metadata_with_prefix() {
+        let output = "bottlerocket-testpkg|1|3.0.0\n".to_string();
+        let result = parse_rpm_metadata("bottlerocket-".to_string(), output).unwrap();
+        let evr = result.get(&PackageName::new("testpkg")).unwrap();
+        assert_eq!(evr.to_string(), "1:3.0.0");
+    }
+
+    #[test]
+    fn parse_rpm_metadata_multiple_packages() {
+        let output = "pkg1|0|1.0\npkg2|1|2.0\n".to_string();
+        let result = parse_rpm_metadata("".to_string(), output).unwrap();
+
+        assert_eq!(
+            result.get(&PackageName::new("pkg1")).unwrap().to_string(),
+            "0:1.0"
+        );
+        assert_eq!(
+            result.get(&PackageName::new("pkg2")).unwrap().to_string(),
+            "1:2.0"
+        );
+    }
+}
--- a/tools/advisory-checker/src/models.rs
+++ b/tools/advisory-checker/src/models.rs
@@ -1,0 +1,182 @@
+//! Advisory data models and deserialization logic.
+
+use clap::Parser;
+use nutype::nutype;
+use regex::Regex;
+use serde::Deserialize;
+use std::collections::HashMap;
+use std::fmt::Display;
+use std::path::PathBuf;
+use std::sync::LazyLock;
+
+const DEFAULT_EPOCH: &str = "0";
+
+static CVE_ID_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r#"^CVE-[0-9]{4}-[0-9]{4,19}$"#).expect("Failed to compile CVE_ID_REGEX")
+});
+
+static GHSA_ID_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r"^GHSA(-[23456789cfghjmpqrvwx]{4}){3}$").expect("Failed to compile GHSA_ID_REGEX")
+});
+
+fn is_integer(s: &str) -> bool {
+    s.parse::<i64>().is_ok()
+}
+
+#[nutype(
+    validate(predicate = is_integer),
+    default = DEFAULT_EPOCH,
+    derive(Debug, Clone, Deserialize, PartialEq, Display, AsRef, Default)
+)]
+pub struct Epoch(String);
+
+/// Short for Epoch, Version. We don't use Release from EVR versioning scheme because every
+/// package update aims at upgrading the version and BRSAs do not include Release field.
+#[derive(Debug, Clone)]
+pub struct EV {
+    pub epoch: Epoch,
+    pub version: String,
+}
+
+impl EV {
+    pub fn new(epoch: &Epoch, version: impl Into<String>) -> Self {
+        Self {
+            epoch: epoch.clone(),
+            version: version.into(),
+        }
+    }
+}
+
+impl Display for EV {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}:{}", self.epoch, self.version)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Deserialize)]
+pub struct PackageName(String);
+
+impl Display for PackageName {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl PackageName {
+    pub fn new(name: impl Into<String>) -> Self {
+        Self(name.into())
+    }
+}
+
+#[derive(Debug)]
+pub struct PackageVersionManifest {
+    packages: HashMap<PackageName, EV>,
+}
+
+impl PackageVersionManifest {
+    pub fn new() -> Self {
+        Self {
+            packages: HashMap::new(),
+        }
+    }
+
+    pub fn insert(&mut self, name: PackageName, evr: EV) {
+        self.packages.insert(name, evr);
+    }
+
+    pub fn get(&self, name: &PackageName) -> Option<&EV> {
+        self.packages.get(name)
+    }
+
+    pub fn merge(&mut self, other: PackageVersionManifest) {
+        self.packages.extend(other.packages);
+    }
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct Advisory {
+    #[serde(rename = "advisory")]
+    pub advisory_info: AdvisoryInfo,
+    #[serde(rename = "updateinfo")]
+    pub update_info: UpdateInfo,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[expect(dead_code)]
+pub struct AdvisoryInfo {
+    pub id: String,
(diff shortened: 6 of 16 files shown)

### CODE REVIEW

File: tools/advisory-parser/src/models.rs [hunk 1]
```diff
@@ -0,0 +1,182 @@
...
+    #[allow(dead_code)]
+    /// CVE identifier (e.g., "CVE-2024-1234")
+    pub cve: Option<String>,
+
+    #[allow(dead_code)]
+    /// GitHub Security Advisory identifier (e.g., "GHSA-xxxx-xxxx-xxxx")
+    pub ghsa: Option<String>,
+
+    #[allow(dead_code)]
+    /// Severity level (e.g., "high", "medium", "low")
```
KCSesh: nit: Medium works, but usually we do: 
```suggestion
    /// Severity level (e.g., "high", "moderate", "low")
```

jmt-lab (approved):
LGTM

twoliter/embedded/rpm2kit
```diff
@@ -4,20 +4,73 @@
...
+
+    echo "Checking advisory: ${advisory_file}"
+
+    while IFS='|' read -r pkg_name patched_version patched_epoch; do
+      local rpm_found=false
+
+      while IFS= read -r rpm_file; do
+        [[ -f "${rpm_file}" ]] || continue
+
+        rpm_info=$(rpm -qp --queryformat "%{NAME} %{VERSION} %{RELEASE} %{EPOCH}" "${rpm_file}" 2>/dev/null)
```
vigh-m: Could you explore using `rpmspec` to query the version info directly from the specfile. Then we don't have to wait for the rpms to build to run this check

twoliter/embedded/rpm2kit
```diff
@@ -4,20 +4,73 @@
...
+        else
+          echo "  Advisory OK: ${pkg_name}"
+        fi
+      done < <(find "${packages_dir}" -name "bottlerocket-${pkg_name}-*.rpm" -type f 2>/dev/null)
+
+      if [[ "${rpm_found}" == false ]]; then
+        echo "  WARNING: RPM not found for package: ${pkg_name}"
+      fi
+    done < <("${0%/*}/advisory-parser" "${advisory_file}")
+  done < <(find "${advisories_dir}" -type f -name "*.toml")
```
vigh-m: In line with the above comment, if `rpmspec` approach works, we can find the specfile directly using the package name and reduce one find operation and loop

tools/advisory-parser/src/models.rs
```diff
@@ -0,0 +1,182 @@
...
+                        "id" => id = Some(map.next_value()?),
+                        "title" => title = Some(map.next_value()?),
+                        "cve" => {
+                            let value: String = map.next_value()?;
+                            // Reject non-ASCII characters (e.g., en dash U+2013, em dash U+2014)
+                            if !value.is_ascii() {
+                                return Err(de::Error::custom("CVE contains non-ASCII characters"));
+                            }
+                            // CVE format: CVE-YYYY-NNNNN where YYYY is year, NNNNN is 4+ digits
+                            let re = Regex::new(r#"^CVE-\d{4}-(0\d{3}|[1-9]\d{3,})$"#).unwrap();
```
ytsssun: Can you make this regex pattern a static variable outside of the scope here - so that it can only be compiled once. 

You can use `lazy_static` for it.

ginglis13: I'd recommend using the regex directly from https://www.cve.org/Resources/Media/Archives/OldWebsite/cve/identifiers/tech-guidance.html, it's slightly different:


```suggestion
                            // The CVE ID regex from https://www.cve.org/Resources/Media/Archives/OldWebsite/cve/identifiers/tech-guidance.html,
                            let re = Regex::new(r#"^CVE-\d{4}-\d{4,}$"#).unwrap();
```

Also, consider moving this to a `lazy_static` compilation of the regular expression rather than compiling per deserialization.

ytsssun: Can you also double check on the regex pattern. I found https://github.com/CVEProject/cve-schema/blob/main/schema/CVE_Record_Format.json#L55 from the official schema.

So maybe 
```suggestion
                            let re = Regex::new(r#"^CVE-[0-9]{4}-[0-9]{4,19}$"#).unwrap();
```

ytsssun: Another thing is `unwrap` is discouraged as it causes panic with no proper error message. Can you use `unwrap_or` or `expect()` or other relevant function to properly handle the error when regex pattern fails to compile.

tools/advisory-parser/src/models.rs
```diff
@@ -0,0 +1,182 @@
...
+                            let value: String = map.next_value()?;
+                            // Reject non-ASCII characters  (e.g., en dash U+2013, em dash U+2014)
+                            if !value.is_ascii() {
+                                return Err(de::Error::custom(
+                                    "GHSA contains non-ASCII characters",
+                                ));
+                            }
+                            // GHSA format: GHSA-xxxx-xxxx-xxxx using base32 chars (excluding 0,1,a,e,i,l,o,s,u)
+                            let re =
+                                Regex::new(r#"^GHSA(-[23456789cfghjmpqrvwx]{4}){3}$"#).unwrap();
```
ytsssun: Same comment, try to compile the regex pattern as a static variable. Also double check the pattern. And replace the usage of `unwrap`.

tools/advisory-parser/src/models.rs
```diff
@@ -0,0 +1,182 @@
...
+    /// Unique identifier for the Bottlerocket Security Advisory
+    pub id: String,
+
+    #[allow(dead_code)]
+    /// Title describing the security issue
+    pub title: String,
+
+    #[allow(dead_code)]
+    /// CVE identifier (e.g., "CVE-2024-1234")
+    pub cve: Option<String>,
```
ginglis13: Why not use field level custom deserializers? This would work:

```
#[serde(deserialize_with = "deserialize_cve_id")]
#[serde(default)]
pub cve: Option<String>,

// ...

pub const CVE_ID_REGEX: &str = r"^CVE-[0-9]{4}-[0-9]{4,}$";

// ...

fn deserialize_cve_id<'de, D>(deserializer: D) -> std::result::Result<Option<String>, D::Error>
where
    D: Deserializer<'de>,
{
    let option: Option<String> = Option::deserialize(deserializer)?;
    if let Some(cve_string) = option {
        let re = Regex::new(CVE_ID_REGEX).unwrap();
        if !re.is_match(&cve_string) {
            return Err(serde::de::Error::custom(AdvisoryInvalidCveIdSnafu {
                input: cve_string,
            }));
        }
        Ok(Some(cve_string))
    } else {
        Ok(None)
    }
}
```

you could do similar for the ghsa field

tools/advisory-parser/src/models.rs
```diff
@@ -0,0 +1,182 @@
...
+                "products",
+            ],
+            AdvisoryInfoVisitor,
+        )
+    }
+}
+
+/// Package metadata affected by the advisory.
+#[derive(Deserialize)]
+pub struct Product {
```
ginglis13: You can do this instead of renaming each individual field

```suggestion
#[derive(Deserialize)]
#[serde(rename_all = "kebab-case")]
pub struct Product {
```

twoliter/embedded/rpm2kit
```diff
@@ -4,20 +4,73 @@
...
+        [[ -f "${rpm_file}" ]] || continue
+
+        rpm_info=$(rpm -qp --queryformat "%{NAME} %{VERSION} %{RELEASE} %{EPOCH}" "${rpm_file}" 2>/dev/null)
+        read -r rpm_name rpm_version rpm_release rpm_epoch <<< "${rpm_info}"
+        [[ "${rpm_name}" != "bottlerocket-${pkg_name}" ]] && continue
+
+        rpm_found=true
+        [[ "${rpm_epoch}" == "(none)" ]] && rpm_epoch="0"
+
+        if [[ $(rpm --eval "%{lua: print(rpm.vercmp('${rpm_epoch}:${rpm_version}-${rpm_release}', '${patched_epoch}:${patched_version}'))}" 2>/dev/null) -lt 0 ]]; then
```
ginglis13: Will we run into issues without the patched_release information available? To be safe, shouldn't we compare just epoch and version here?
```suggestion
        if [[ $(rpm --eval "%{lua: print(rpm.vercmp('${rpm_epoch}:${rpm_version}', '${patched_epoch}:${patched_version}'))}" 2>/dev/null) -lt 0 ]]; then
```

twoliter/embedded/rpm2kit
```diff
@@ -4,20 +4,73 @@
...
+          advisory_id=$(basename "${advisory_file}" .toml)
+          echo "  ADVISORY VIOLATION:"
+          echo "    Advisory: ${advisory_file}"
+          echo "    CVE: ${advisory_id}"
+          echo "    RPM: ${rpm_name}"
+          echo "    Found: epoch=${rpm_epoch}, version=${rpm_version}, release=${rpm_release}"
+          echo "    Required: epoch=${patched_epoch}, version=${patched_version}"
+          exit_code=1
+        else
+          echo "  Advisory OK: ${pkg_name}"
```
ginglis13: I'd suggest also outputting the advisory file here too

tools/advisory-parser/src/main.rs
```diff
@@ -0,0 +1,256 @@
...
+mod models;
+
+use models::Advisory;
+use snafu::{ensure, ResultExt};
+use std::path::PathBuf;
+use std::{env, fs, process};
+
+fn run() -> Result<()> {
+    let args: Vec<String> = env::args().collect();
+    ensure!(args.len() == 2, error::UsageSnafu { program: &args[0] });
```
ytsssun: nit: it might be good to document why we want to strictly check for arguments length to be 2, mentioning what is the expected usage for this command would help explaining.

tools/advisory-parser/src/models.rs
piyush-jena: added lazy static on the top and took care of these comments together.

twoliter/embedded/rpm2kit
```diff
@@ -4,20 +4,73 @@
...
+
+    while IFS='|' read -r pkg_name patched_version patched_epoch; do
+      local rpm_found=false
+
+      while IFS= read -r rpm_file; do
+        [[ -f "${rpm_file}" ]] || continue
+
+        rpm_info=$(rpm -qp --queryformat "%{NAME} %{VERSION} %{RELEASE} %{EPOCH}" "${rpm_file}" 2>/dev/null)
+        read -r rpm_name rpm_version rpm_release rpm_epoch <<< "${rpm_info}"
+        [[ "${rpm_name}" != "bottlerocket-${pkg_name}" ]] && continue
```
cbgbt: I want to say that we made some changes to ensure that vendor information gets plumbed through twoliter (@sky1122) Should this "bottlerocket" instead be dynamic based on the kit vendor?

twoliter/embedded/Makefile.toml
```diff
@@ -741,6 +678,27 @@ fi
 '''
 ]
 
+[tasks.check-advisories]
```
cbgbt: Is this intended to be run automatically, as part of build-kit?

tools/advisory-parser/src/models.rs
piyush-jena: I have kept the original deserializer implementation to enforce the requirement of cve or ghsa in the BRSA. I think we should verify this through code in the kits. Let me know if you think differently.

twoliter/embedded/rpm2kit
piyush-jena: I didn't put the file there because the previous line in the log we will have the file name.
```
echo "Checking advisory: ${advisory_file}"
```

The example above shows the expected output. Let me know if you still think we should change the output format.

twoliter/embedded/rpm2kit
piyush-jena: Didn't run into any issue but that might be because I got lucky. Updated this.

tools/advisory-parser/src/models.rs
```diff
@@ -0,0 +1,198 @@
...
+//! This module defines the structure of security advisory TOML files and implements
+//! custom deserialization with validation for CVE and GHSA identifiers.
+
+use lazy_static::lazy_static;
+use regex::Regex;
+use serde::de::{self, MapAccess, Visitor};
+use serde::{Deserialize, Deserializer};
+use std::fmt;
+
+lazy_static! {
```
cbgbt: This is OK but the rust community has moved to [`LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html) since they added it to the std library.

tools/advisory-parser/src/models.rs
cbgbt: I think the best way to handle this is with the `nutype` crate with the serde feature enabled. The Rust mantra is "parse, don't validate"

tools/advisory-parser/src/models.rs
```diff
@@ -0,0 +1,198 @@
...
+#[serde(rename_all = "lowercase")]
+pub enum Severity {
+    Low,
+    #[serde(alias = "medium")]
+    Moderate,
+    High,
+    Critical,
+}
+
+impl<'de> Deserialize<'de> for AdvisoryInfo {
```
cbgbt: Yeah, as mentioned, please use `nutype` for this kind of thing.

In the future, you can get agents to respect this by prompting with something like:
* Use `nutype` to model all domain types
* Avoid "primitive obsession" - domain types must be strongly modeled in the type system

tools/advisory-parser/src/main.rs
```diff
@@ -0,0 +1,258 @@
...
+    fn test_valid_cve_formats() {
+        let valid_cves = vec![
+            "CVE-2024-0001",
+            "CVE-2024-1234",
+            "CVE-2024-12345",
+            "CVE-2025-472688",
+            "CVE-1999-0001",
+        ];
+
+        for cve in valid_cves {
```
cbgbt: Use the `test-case` crate for this.

when you iterate over test data like this in Rust, you don't get unique test cases in the `cargo test` output. But people tend to prefer unique test output per data trial.

`test-case` macro solves for this.

tools/advisory-parser/src/models.rs
```diff
@@ -0,0 +1,198 @@
...
+use serde::{Deserialize, Deserializer};
+use std::fmt;
+
+lazy_static! {
+    /// The CVE ID regex from https://github.com/CVEProject/cve-schema/blob/main/schema/CVE_Record_Format.json#L55
+    static ref CVE_ID_REGEX: Regex = Regex::new(r#"^CVE-[0-9]{4}-[0-9]{4,19}$"#)
+        .expect("Failed to compile CVE_ID_REGEX");
+
+    /// The GHSA ID regex from https://github.com/github/advisory-database?tab=readme-ov-file#ghsa-ids
+    static ref GHSA_ID_REGEX: Regex = Regex::new(r"^GHSA-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{4}$")
```
cbgbt: This isn't actually the regex on the page you linked.

`/GHSA(-[23456789cfghjmpqrvwx]{4}){3}/`

tools/advisory-parser/src/main.rs
```diff
@@ -0,0 +1,258 @@
...
+                "Failed to parse valid GHSA: {}",
+                ghsa
+            );
+        }
+    }
+
+    #[test]
+    fn test_invalid_ghsa_formats() {
+        let invalid_ghsas = vec![
+            "GHSA-xxxx-yyyy-zzzz",    // Invalid characters
```
cbgbt: This seemingly matches the regex that you gave? it allows x,y,z.

twoliter/embedded/rpm2kit
piyush-jena: Updated. This is much faster and can be run without build being present.

twoliter/embedded/rpm2kit
piyush-jena: I used a different approach. First built a dictionary with key package_name and value as epoch:version by running rpmspec on all spec files. Then went through all advisories and searched for the package in the dictionary and compared values.

Its not necessary that the package directory name and the rpm package name is the same.

piyush-jena:
Resolved all comments.

cbgbt (commented):
> I had to fix the advisories in the earlier versions because they didn't have patched-epoch field. We can plan on either defaulting to 0 here or add those fields in the advisories.

Zero as a default should be OK. RPM treats unspecified as 0.

I noticed that `bottlerocket-core-kit/advisories/2.1.0/BRSA-1j0o73qa.toml` has an odd patched version:

```
patched-version = "kernel-5.15.160-104.158.amzn2"
```

I suppose this just shows up as a warning? What should happen here?

piyush-jena:
> > I had to fix the advisories in the earlier versions because they didn't have patched-epoch field. We can plan on either defaulting to 0 here or add those fields in the advisories.
> 
> Zero as a default should be OK. RPM treats unspecified as 0.
> 
> I noticed that `bottlerocket-core-kit/advisories/2.1.0/BRSA-1j0o73qa.toml` has an odd patched version:
> 
> ```
> patched-version = "kernel-5.15.160-104.158.amzn2"
> ```
> 
> I suppose this just shows up as a warning? What should happen here?

@cbgbt 

1. Yeah I used default so that I don't have to fix the old advisories.

2. Because the corresponding package is not present, we just emit a warning like this 
```
Checking advisory: advisories/2.1.0/BRSA-1j0o73qa.toml
  WARNING: Package kernel-5.15 could not be found.
```

twoliter/embedded/Makefile.toml
piyush-jena: I created this as a separate task. I didn't add dependency because the rpms are not necessary and we can run this prior to the build. Imagine a PR that adds an advisory, we can just run a check-advisories check and then merge. The plan is to run it as a CI check in kit repos. To invoke we can just run `make twoliter check-advisories`.

tools/advisory-parser/src/models.rs
piyush-jena: Updated.

tools/advisory-parser/src/models.rs
piyush-jena: Updated. Have to say, pretty clean.

tools/advisory-parser/src/main.rs
piyush-jena: Updated.

twoliter/embedded/Makefile.toml
piyush-jena: I realize you were reviewing an older implementation. Let me describe the new implementation in the description.

twoliter/embedded/rpm2kit
piyush-jena: Running it in bottlerocket-sdk context now so this issue is fixed.

twoliter/embedded/Makefile.toml
```diff
@@ -741,6 +678,83 @@ fi
...
+    while IFS="|" read -r pkg_name pkg_epoch pkg_version; do
+        [ -z "$pkg_name" ] && continue
+        [ "${pkg_epoch}" = "(none)" ] && pkg_epoch="0"
+        pkg_metadata["${pkg_name}"]="${pkg_epoch}:${pkg_version}"
+    done < <(rpmspec -q --qf "%{Name}|%{Epoch}|%{Version}\n" "${spec_file}" 2>/dev/null)
+done < <(find . -name "*.spec" -type f ! -path "*/target/*")
+
+exit_code=0
+
+while IFS= read -r advisory_file; do
```
bcressey: I'm not a fan of this structure. Claude likes to generate these `while IFS= ... ; do ...` constructs but they are pretty unreadable / unmaintainable in my experience. Especially nested with the different redirects.

From an architecture perspective, this would make more sense at package build time. You'll already be in a `docker build` context there with the environment set up for building (macros, etc).

I would suggest pivoting `advisory-parser` into a tool that:
1. Finds advisories that match a given package
2. Takes the single spec to operate on as an argument.
3. Performs the `rpmspec` and `rpm --eval` calls itself as needed.

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,164 @@
...
+        CommandFailureSnafu { bin_path, output }
+    );
+
+    Ok(String::from_utf8_lossy(&output.stdout).to_string())
+}
+
+fn validate_advisories(spec_path: &Path, advisories_dir: &Path) -> Result<()> {
+    let pkg_metadata = get_rpm_metadata(spec_path)?;
+
+    if !pkg_metadata.is_empty() {
```
cbgbt: Should it be an error if we can't extract RPM metadata?

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,164 @@
...
+        output: result.trim(),
+    })
+}
+
+fn visit_advisories(
+    dir: &Path,
+    pkg_metadata: &HashMap<String, String>,
+    violations: &mut Vec<String>,
+) -> Result<()> {
+    for entry in fs::read_dir(dir).context(ReadDirSnafu { path: dir })? {
```
cbgbt: I'd suggest sorting to make this deterministic.

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,164 @@
...
+            visit_advisories(&path, pkg_metadata, violations)?;
+        } else {
+            check_advisory(&path, pkg_metadata, violations)?;
+        }
+    }
+
+    Ok(())
+}
+
+fn check_advisory(
```
cbgbt: Can we get an integration test with edge cases for this? We have other integration tests in twoliter that build kits. This logic seems correct, but it's pretty hard to know without tests.

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,164 @@
...
+    println!("Advisory check passed for {}", spec_path.display());
+    Ok(())
+}
+
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("Error: {e}");
+        process::exit(1);
+    }
+}
```
cbgbt: Snafu now has `#[snafu::report]` macro, so you can get rid of `run()` and just move all the logic to main and add that macro on top:

```rust
#[snafu::report]
fn main() {
    let args: Vec<String> = env::args().collect();
    // ... 
}
```

tools/advisory-checker/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "advisory-checker"
+version = "0.1.0"
+edition = "2021"
+license = "Apache-2.0 OR MIT"
+publish = false
+exclude = ["README.md"]
+
+[dependencies]
+nutype = { version = "0.6", features = ["regex", "serde"] }
```
cbgbt: This should be a workspace dependency

cbgbt (commented):
Do we still need changes to `Makefile.toml` to ensure that these get run during a kit build? Maybe I missed it, but I'm not seeing how this gets invoked.

piyush-jena:
@cbgbt 

> Do we still need changes to Makefile.toml to ensure that these get run during a kit build? Maybe I missed it, but I'm not seeing how this gets invoked.

This runs as part of the `make twoliter build-package` so we don't need to make any changes. A bad advisory would show up as build failure with error messages like shown above in the Description.

twoliter/embedded/build.Dockerfile
```diff
@@ -108,6 +108,10 @@ RUN --mount=target=/host \
       -exec cp {} ./rpmbuild/SOURCES/ \; && \
     rm /bypass
 
+RUN --mount=source=advisories,target=/home/builder/advisories \
+    --mount=target=/host \
+    /host/build/tools/advisory-checker rpmbuild/SPECS/${PACKAGE}.spec /home/builder/advisories/
```
bcressey: Better to run this before the package build.

tools/advisory-checker/src/error.rs
```diff
@@ -0,0 +1,54 @@
...
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub), module)]
+pub enum Error {
+    #[snafu(display("Usage: {} <spec-file> <advisories-directory>", program))]
+    Usage { program: String },
+
+    #[snafu(display("The provided Advisories directory '{}' doesn't exist", path))]
```
bcressey: nit:
```suggestion
    #[snafu(display("The provided advisories directory '{}' doesn't exist", path))]
```

tools/advisory-checker/src/error.rs
```diff
@@ -0,0 +1,54 @@
...
+    AdvisoriesDirMissing { path: String },
+
+    #[snafu(display("Failed to execute command"))]
+    ExecutionFailure {
+        command: Command,
+        source: std::io::Error,
+    },
```
bcressey: We should include the command in the error message.

tools/advisory-checker/src/error.rs
```diff
@@ -0,0 +1,54 @@
...
+    ReadDirEntry { source: std::io::Error },
+
+    #[snafu(display("Failed to parse vercmp output '{}'", output))]
+    ParseVercmp {
+        output: String,
+        source: std::num::ParseIntError,
+    },
+
+    #[snafu(display("Advisory violations found:\n{}", violations.join("\n")))]
+    AdvisoryViolations { violations: Vec<String> },
```
bcressey: I'd prefer to handle the advisory printing in the caller (or the caller's caller) - feels weird to format output inside an error message like this.

tools/advisory-checker/src/error.rs
```diff
@@ -0,0 +1,54 @@
...
+use std::path::PathBuf;
+use std::process::{Command, Output};
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub), module)]
+pub enum Error {
+    #[snafu(display("Usage: {} <spec-file> <advisories-directory>", program))]
+    Usage { program: String },
```
bcressey: Let's use a real arg parser like `clap` for consistency with other tools

tools/advisory-checker/src/models.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// The CVE ID regex from https://github.com/CVEProject/cve-schema/blob/main/schema/CVE_Record_Format.json#L55.
+static CVE_ID_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r#"^CVE-[0-9]{4}-[0-9]{4,19}$"#).expect("Failed to compile CVE_ID_REGEX")
+});
+
+/// The GHSA ID regex from https://github.com/github/advisory-database?tab=readme-ov-file#ghsa-ids.
+static GHSA_ID_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r"^GHSA(-[23456789cfghjmpqrvwx]{4}){3}$").expect("Failed to compile GHSA_ID_REGEX")
+});
```
bcressey: Do we have existing advisories that reference a GHSA?

Are you sure this is correct and a GHSA go in a separate field, and not something like this?
```
cve = "GHSA-...."
```

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,164 @@
+/*!
+advisory-checker validates rpmspec files against security advisories to ensure
+packages meet minimum version requirements for known vulnerabilities.
+*/
+
+mod error;
+mod models;
+
+use error::error::*;
+use error::Result;
```
bcressey: nit: can merge these

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,164 @@
...
+            "--qf",
+            "%{Name}|%{Epoch}|%{Version}\n",
+            &spec_path_str,
+        ],
+    )?;
+
+    let mut metadata = HashMap::new();
+    for line in output.lines() {
+        let parts: Vec<&str> = line.split('|').collect();
+        if parts.len() == 3 {
```
bcressey: Can we error out if we don't find all three parts?

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,164 @@
...
+    Ok(())
+}
+
+fn check_advisory(
+    advisory_path: &Path,
+    pkg_metadata: &HashMap<String, String>,
+    violations: &mut Vec<String>,
+) -> Result<()> {
+    // BRSA files are the only toml files in the advisory directory. We want to ignore the .gitkeep files.
+    if advisory_path.extension() == Some(OsStr::new("toml")) {
```
bcressey: Let's invert this check and early return so we can outdent the rest of the function.

tools/advisory-checker/src/models.rs
```diff
@@ -0,0 +1,72 @@
...
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct Advisory {
+    pub advisory: AdvisoryInfo,
```
bcressey:
```suggestion
    pub advisory_info: AdvisoryInfo,
```

tools/advisory-checker/src/models.rs
piyush-jena: I chose to use the official Regex here which I found from - https://github.com/github/advisory-database?tab=readme-ov-file#ghsa-ids.

twoliter/embedded/build.Dockerfile
piyush-jena: I placed it before the `rpmbuild` but after we put the sources and macros in the right place. Should I move this before that and add the line for copying macros into this?

tools/advisory-checker/src/models.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub id: String,
+    pub title: String,
+    pub cve: Option<CveId>,
+    pub ghsa: Option<GhsaId>,
+    pub severity: Severity,
+    pub description: String,
+    pub products: Vec<Product>,
+}
+
+#[nutype(
```
ginglis13, line 119: nice! TIL about `nutype`

twoliter/embedded/build.Dockerfile
```diff
@@ -108,6 +108,10 @@ RUN --mount=target=/host \
       -exec cp {} ./rpmbuild/SOURCES/ \; && \
     rm /bypass
 
+RUN --mount=source=advisories,target=/home/builder/advisories \
+    --mount=target=/host \
+    /host/build/tools/advisory-checker --spec-file rpmbuild/SPECS/${PACKAGE}.spec --advisories-dir /home/builder/advisories/
+
```
ginglis13: I don't think this will this work for custom kits that may not include an `advisories` directory. Probably safer to mark as not required in the mount args and check for existence of the advisories dir:

```suggestion
RUN --mount=source=advisories,target=/home/builder/advisories,required=false \
    --mount=target=/host \
    if [ -d /home/builder/advisories ]; then \
        /host/build/tools/advisory-checker --spec-file rpmbuild/SPECS/${PACKAGE}.spec --advisories-dir /home/builder/advisories/ \
    fi
```

Alternatively, could just set `required=false` in the mount args and in your advisory-checker binary, throw a warning if the dir doesn't exist and move on.

I'd also advocate for unit/integration testing for this case.

tools/advisory-checker/src/main.rs
piyush-jena: There are 2 sources of error.
1. rpmspec command fails - this means its a malformed spec file and build would fail later.
2. output doesn't have 3 parts: name, epoch, version - I couldn't get it to fail on a valid spec file. If epoch is not present then the value is `(none)`. Name and version are mandatory fields and would just cause command failure above.

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,174 @@
...
+    let mut metadata = HashMap::new();
+    for line in output.lines() {
+        let parts: Vec<&str> = line.split('|').collect();
+        ensure!(parts.len() == 3, RpmSpecFormatSnafu { spec_output: line });
+
+        let name = parts[0].strip_prefix(&pkg_prefix).unwrap_or(parts[0]);
+        let epoch = if parts[1] == "(none)" { "0" } else { parts[1] };
+        let version = parts[2];
+        metadata.insert(name.to_string(), format!("{epoch}:{version}"));
+    }
```
cbgbt: Let's extract this to a function and unit test it.

tools/advisory-checker/src/main.rs
cbgbt: I just mean, if `pkg_metadata` is empty, then something is probably terribly wrong, right? From a defense in depth perspective, we should raise an error here if it is empty.

tools/advisory-checker/src/main.rs
piyush-jena: I added integration tests for this.

twoliter/embedded/build.Dockerfile
piyush-jena: The docker version we have doesn't support `required`. I handled it by mounting the host (the entire kit) directory instead of just advisories and checking the presence of advisories directory in the binary itself.

tools/advisory-checker/src/main.rs
piyush-jena: Updated

piyush-jena:
rebased ^

tests/integration-tests/src/advisory_checker.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+    std::fs::write(&path, content).unwrap();
+    path
+}
+
+#[test]
+#[ignore]
+fn test_missing_advisories_dir_succeeds() {
+    let temp_dir = TempDir::new().unwrap();
+    let spec_file = create_spec_file(&temp_dir, "testpkg", "1.0.0");
```
cbgbt: Seems unnecessary in this test.

tests/integration-tests/src/advisory_checker.rs
```diff
@@ -0,0 +1,282 @@
...
+            "--advisories-dir",
+            nonexistent_dir.to_str().unwrap(),
+        ],
+        [],
+    );
+
+    assert!(output.status.success());
+    let stdout = String::from_utf8_lossy(&output.stdout);
+
+    let output_pattern = r"Advisories directory .* not found, skipping";
```
cbgbt: I'm glad that this behavior is documented but it has me wondering if we need to consider how to protect against cases where we accidentally point the checker at the wrong directory. Should the script fail if the directory doesn't exist? That seems like maybe a choice that would reduce the chance of making a mistake.

tests/integration-tests/src/advisory_checker.rs
```diff
@@ -0,0 +1,282 @@
...
+            spec_file.to_str().unwrap(),
+            "--advisories-dir",
+            advisories_dir.to_str().unwrap(),
+        ],
+        [],
+    );
+
+    assert!(!output.status.success());
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    assert!(stdout.contains("Advisory violations found"));
```
cbgbt, line 143: Can you document the violation in the test with a comment?

tests/integration-tests/src/advisory_checker.rs
```diff
@@ -0,0 +1,282 @@
...
+        ],
+        [],
+    );
+
+    assert!(output.status.success());
+}
+
+#[test]
+#[ignore]
+fn test_advisory_for_different_package_ignored() {
```
cbgbt: Isn't it a problem that this advisory was written for a package that we seemingly do not ship?

piyush-jena: For instance, in core-kit we have advisories for old kernels. We don't want them to cause any issues.

tests/integration-tests/src/advisory_checker.rs
piyush-jena: That was the behaviour I implemented initially but switched to this because of this comment - https://github.com/bottlerocket-os/twoliter/pull/604#discussion_r2628313786. There may be kit users who get rid of the advisories directory and we don't want to break the build for them.

tests/integration-tests/src/advisory_checker.rs
cbgbt: I misread the test. Please disregard this.

tests/integration-tests/src/advisory_checker.rs
cbgbt: Ah, of course.

tests/integration-tests/src/advisory_checker.rs
cbgbt: I wonder if the "right way" is to make it so that we don't invoke the tool when there are no advisories present, but the invocation of the tool implies the presence of advisories.

tools/advisory-checker/src/models.rs
```diff
@@ -0,0 +1,73 @@
...
+
+#[derive(Debug, Deserialize)]
+pub struct Advisory {
+    #[serde(rename = "advisory")]
+    pub advisory_info: AdvisoryInfo,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[allow(dead_code)]
```
cbgbt: Can you elaborate on why this is `allow`? Seems like this should never be dead_code. Is it because of individual fields or something?

In general, we should use `expect` instead of `allow`. Let's also apply this to the smallest possible field and comment on why it is used. It seems like a bug if this struct is never used.

tools/advisory-checker/src/models.rs
```diff
@@ -0,0 +1,73 @@
...
+pub struct AdvisoryInfo {
+    pub id: String,
+    pub title: String,
+    pub cve: Option<CveId>,
+    pub ghsa: Option<GhsaId>,
```
cbgbt, line 113: I think we lost the invariant that we must have at least one of these two fields.

Can we add a regression test to catch this?

I think the canonical way to resolve this is to use the `serde(try_from)` macro option, where the fallible `TryFrom` impl would detect composite invariants.

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,292 @@
...
+            &spec_path_str,
+        ],
+    )?;
+
+    parse_rpm_metadata(pkg_prefix, output)
+}
+
+fn parse_rpm_metadata(pkg_prefix: String, output: String) -> Result<HashMap<String, String>> {
+    let mut metadata = HashMap::new();
+    for line in output.lines() {
```
cbgbt: Can we add a test that captures testing a package with multiple subpackages? I think we only have the one spec template right now.

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,292 @@
...
+            "--qf",
+            "%{Name}|%{Epoch}|%{Version}\n",
+            &spec_path_str,
+        ],
+    )?;
+
+    parse_rpm_metadata(pkg_prefix, output)
+}
+
+fn parse_rpm_metadata(pkg_prefix: String, output: String) -> Result<HashMap<String, String>> {
```
cbgbt: This function does some kind of string processing, but it's actually hard to tell just by looking at the function signature what. At the very least, it needs a docstring.

The "Rust way" would lean more heavily on using a composite type instead of a HashMap though -- You could alter the return type to be `Result<HashMap<PackageName, EVR>>` where `PackageName` and `EVR` were newtypes or type aliases. Alternatively you could return `Result<PackageVersionManifest>` or some kind of composite type that documented what was being processed.

With the existing code "`parse_rpm_metadata`" could mean any kind of information from the RPM, and the function signature doesn't really tell you what. The only way to know what the code is doing is to read all of it.

twoliter/embedded/build.Dockerfile
```diff
@@ -108,6 +108,9 @@ RUN --mount=target=/host \
       -exec cp {} ./rpmbuild/SOURCES/ \; && \
     rm /bypass
 
+RUN --mount=target=/host \
+    /host/build/tools/advisory-checker --spec-file rpmbuild/SPECS/${PACKAGE}.spec --advisories-dir /host/advisories/
+
```
cbgbt: To make it so that the advisory checker itself can fail when its inputs are missing, we could make the invocation here conditional.

tools/advisory-checker/src/models.rs
piyush-jena: This is intentional. While we don't do this in core-kit, we publish empty advisories like [BRSA-ayfi57tij8hr.toml](https://github.com/bottlerocket-os/bottlerocket-kernel-kit/blob/develop/advisories/4.5.0/BRSA-ayfi57tij8hr.toml) in kernel-kit to get people to update.

tools/advisory-checker/src/models.rs
piyush-jena: It was to fix this warning I got.
```
warning: fields `title`, `cve`, `ghsa`, `severity`, and `description` are never read
```
We check CVE, GHSA and Severity while deserializing. Title and Description are unnecessary but I kept it if we want to add additional restrictions.

`#[expect(dead_code)]` fixes the warning.

twoliter/embedded/build.Dockerfile
piyush-jena: Updated and test output
```
[fedora@ip-yyy bottlerocket-core-kit]$ sudo rm -rf advisories/
[fedora@ip-xxx bottlerocket-core-kit]$ PACKAGE="iputils" make twoliter build-package
Found Twoliter v0.15.1 installed.
Skipping installation.
[2026-01-08T00:04:56Z INFO  twoliter::project::lock] Resolving project references to check against lock file
[2026-01-08T00:04:56Z INFO  twoliter::project::lock::image] Resolving dependency image dependency 'bottlerocket-sdk-0.66.0@public.ecr.aws/bottlerocket/bottlerocket-sdk:v0.66.0'.
[2026-01-08T00:04:56Z INFO  twoliter::project::lock] Extracting kit dependencies. dependencies=[]
[2026-01-08T00:04:56Z INFO  twoliter::project::lock] Finished fetching kit dependencies.
[cargo-make] INFO - cargo make 0.37.24
[cargo-make] INFO - 
[cargo-make] INFO - Build File: /home/fedora/Repositories/bottlerocket-core-kit/build/tools/Makefile.toml
[cargo-make] INFO - Task: build-package
[cargo-make] INFO - Running Task: check-cargo-version
[cargo-make] INFO - Running Task: setup
[cargo-make] INFO - Running Task: setup-build
[cargo-make] INFO - Running Task: fetch-sdk
[cargo-make] INFO - Running Task: fetch-sources
[cargo-make] INFO - Running Task: fetch-vendored
[cargo-make] INFO - Running Task: publish-setup
00:05:05 [INFO] Found infra config at path: /home/fedora/Repositories/bottlerocket-core-kit/Infra.toml
00:05:05 [INFO] No repo config in '/home/fedora/Repositories/bottlerocket-core-kit/Infra.toml' - using local roles/keys
[cargo-make] INFO - Running Task: cargo-metadata
[cargo-make] INFO - Running Task: validate-kits
[cargo-make] INFO - Running Task: build-package
   Compiling iputils v0.1.0 (/home/fedora/Repositories/bottlerocket-core-kit/packages/iputils)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 41.98s
[cargo-make] INFO - Build Done in 51.08 seconds.
```

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,299 @@
...
+    ensure!(
+        output.status.success(),
+        CommandFailureSnafu { bin_path, output }
+    );
+
+    Ok(String::from_utf8_lossy(&output.stdout).to_string())
+}
+
+fn rpm_vercmp(ver1: &str, ver2: &str) -> Result<i32> {
+    let lua_expr = format!("%{{lua: print(rpm.vercmp(\"{ver1}\", \"{ver2}\"))}}");
```
KCSesh: It would be good to sanitize the version input on the lua execution.

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,299 @@
...
+    })?;
+
+    for product in advisory.advisory_info.products {
+        let Some(spec_ev) = pkg_metadata.get(&product.package_name) else {
+            continue;
+        };
+
+        let advisory_ev = EV::new(&product.patched_epoch, &product.patched_version);
+
+        if rpm_vercmp(&spec_ev.to_string(), &advisory_ev.to_string())? < 0 {
```
KCSesh: Might be good to wrap another SNAFU here for Error context that relates the specific advisory to the error.

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,299 @@
...
+        path: advisory_path,
+    })?;
+
+    for product in advisory.advisory_info.products {
+        let Some(spec_ev) = pkg_metadata.get(&product.package_name) else {
+            continue;
+        };
+
+        let advisory_ev = EV::new(&product.patched_epoch, &product.patched_version);
+
```
KCSesh: Probably better to just use a rust crate for rpm comparisons. I imagine this is an artifact of moving from bash to rust?

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,299 @@
...
+        path: advisory_path,
+    })?;
+
+    let advisory: Advisory = toml::from_str(&content).context(ParseAdvisorySnafu {
+        path: advisory_path,
+    })?;
+
+    for product in advisory.advisory_info.products {
+        let Some(spec_ev) = pkg_metadata.get(&product.package_name) else {
+            continue;
```
KCSesh: Should this be at minimum a log? This means a package_name is incorrect, right? 

Could it be a typo? Should we error?

tools/advisory-checker/src/models.rs
```diff
@@ -0,0 +1,127 @@
...
+    #[serde(alias = "medium")]
+    Moderate,
+    High,
+    Critical,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct Product {
+    pub package_name: String,
```
KCSesh: Should we make use of: `PackageName` here?

tools/advisory-checker/src/models.rs
```diff
@@ -0,0 +1,127 @@
...
+#[derive(Debug, Clone)]
+pub struct EV {
+    pub epoch: String,
+    pub version: String,
+}
+
+impl EV {
+    pub fn new(epoch: impl Into<String>, version: impl Into<String>) -> Self {
+        Self {
+            epoch: epoch.into(),
```
KCSesh: Might be good to do some verification on if this is a clean Strng to Int. 

Can add logic here, or consider using the `nutype`

tools/advisory-checker/src/main.rs
piyush-jena: It just means that the package for which this particular advisory is created is not in the list of the packages we are shipping in the core-kit. For example: the current advisory is for a kernel-5.10 but core-kit no longer has it in the packages, so we skip this.

tools/advisory-checker/src/main.rs
piyush-jena: Updated to `rpmvercmp(&spec_ev.to_string(), &advisory_ev.to_string()).context(RpmVerCmpSnafu)?.is_lt()`. The error we check here is something that breaks the function. I am handling the advisory violation gracefully so that we are able to collect all the violations in 1 go instead of erroring out the first time we find a violation.

twoliter/embedded/build.Dockerfile
```diff
@@ -108,6 +108,15 @@ RUN --mount=target=/host \
       -exec cp {} ./rpmbuild/SOURCES/ \; && \
     rm /bypass
 
+RUN --mount=target=/host \
```
KCSesh: We can probably move this up before the previous : `RUN --mount=target=/host` if I understand correctly, to fail quicker if there is an advisory problem. 

No need to setup the `builddep` and then to have it fail

tools/advisory-checker/src/error.rs
```diff
@@ -0,0 +1,53 @@
...
+use std::path::PathBuf;
+use std::process::Output;
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub), module)]
+pub enum Error {
+    #[snafu(display("Usage: {} <spec-file> <advisories-directory>", program))]
+    Usage { program: String },
```
KCSesh: Seems not used

tools/advisory-checker/src/error.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    #[snafu(display("The provided advisories path '{}' doesn't exist or is not a directory", path.display()))]
+    AdvisoriesDir { path: PathBuf },
```
KCSesh: Seems unused.

tests/integration-tests/src/appinventory.rs
```diff
@@ -18,7 +18,7 @@ const EXPECTED_INVENTORY_PATH: &str =
 
 // Last Bottlerocket release that doesn't include source packages in application inventory.
 // For older releases, we check that reference inventory is a subset of current.
-const SOURCE_PACKAGE_INVENTORY_ANCHOR_VERSION: &str = "1.52.0";
+const SOURCE_PACKAGE_INVENTORY_ANCHOR_VERSION: &str = "1.53.0";
```
KCSesh: This got delayed to 1.53.0 right?

twoliter/embedded/build.Dockerfile
piyush-jena: Its intentionally placed here. `.rpmmacros` need to be copied to home directory in the sdk context 
to resolve the macros present in the spec files. For example: `%{_cross_os}`.

tests/integration-tests/src/appinventory.rs
piyush-jena: I will wait till before merge to update this. We haven't done a twoliter release after that PR was merged.

twoliter/embedded/build.Dockerfile
piyush-jena: Actually can be moved a step higher and it makes it faster.

piyush-jena:
Fixed a bad rebase ^

tests/integration-tests/src/appinventory.rs
piyush-jena: addressed in #637

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,334 @@
...
+use models::{Advisory, Args, Epoch, PackageName, PackageVersionManifest, EV};
+use rpm::rpm_evr_compare;
+use snafu::{ensure, ResultExt};
+use std::collections::HashMap;
+use std::ffi::OsStr;
+use std::fs;
+use std::path::Path;
+use std::process::Command;
+
+const ADVISORY_CHECKER_ANCHOR_DATE: &str = "2025-06-11";
```
KCSesh: What does this represent? Can you add a comment to describe it

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,334 @@
...
+                    if rpm_evr_compare(&spec_ev.to_string(), &advisory_ev.to_string()).is_lt() {
+                        violations.push(format!(
+                            "BRSA ID: {}\n package name: {}\n package version: {spec_ev}\n min. expected version: {advisory_ev}",
+                            advisory.advisory_info.id, package_name
+                        ));
+                    }
+                }
+            }
+        } else {
+            // Package missing - check end-of-life
```
KCSesh, line 179: Trying to think through the scenarios here, where we may remove a package without an Advisory.
Like if we move a package from one kit to another? 

Would we still create an EOL for that?

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,334 @@
...
+            }
+        }
+        Ok(())
+    }
+
+    scan(dir, &mut advisories_by_product)?;
+    Ok(advisories_by_product)
+}
+
+fn validate_advisories(advisories_dir: &Path, packages_dir: &Path) -> Result<()> {
```
KCSesh, line 148: nit: Might be nice to have some debug logs in this so we can understand this.

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,334 @@
...
+                            advisory.advisory_info.id, package_name
+                        ));
+                    }
+                }
+            }
+        } else {
+            // Package missing - check end-of-life
+            if let Some(latest) = advisories.last() {
+                let issue_date = latest.update_info.issue_date.to_string();
+                if issue_date.as_str() > ADVISORY_CHECKER_ANCHOR_DATE
```
KCSesh: Not sure I follow this

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,334 @@
+//! Validates rpmspec files against security advisories to ensure packages meet
```
KCSesh, line 1: Please make sure to test on Kernel-kit as well.

tools/advisory-checker/src/main.rs
piyush-jena: Updated to make it clearer.

tools/advisory-checker/src/main.rs
piyush-jena: I added some comments to make it easier to understand. Adding more log lines won't help a lot because they won't show up unless there's an error but we have error logging for that. But let me know if this will help in other ways.

tools/advisory-checker/src/main.rs
piyush-jena: Updated the variable name and added comments to make it clearer.

tools/advisory-checker/src/main.rs
piyush-jena: Tested kernel-kit and bob and confirmed it works as expected.

piyush-jena:
Updates to resolve comments, fix integration tests from recent changes, and added comments. ^

piyush-jena:
cleaned up integration tests ^

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,359 @@
...
+use std::fs;
+use std::path::Path;
+use std::process::Command;
+
+// This is the last time when we removed a package from the kits without an
+// EOL advisory. We make sure that any advisory published post this date
+// either has a corresponding package or an EOL advisory.
+const LAST_UNTRACKED_PACKAGE_REMOVAL_DATE: &str = "2025-07-26";
+
+fn command<I, S>(bin_path: &str, args: I) -> Result<String>
```
ginglis13, line 23: looks like the `S` type parameter is unused

tools/advisory-checker/src/main.rs
```diff
@@ -0,0 +1,359 @@
...
+        ensure!(
+            parts.len() == 3,
+            RpmSpecFormatSnafu {
+                spec_output: metadata
+            }
+        );
+
+        let name = parts[0].strip_prefix(&pkg_prefix).unwrap_or(parts[0]);
+        let epoch_str = if parts[1] == "(none)" { "0" } else { parts[1] };
+        let epoch = Epoch::try_new(epoch_str).expect("epoch should be valid integer");
```
ginglis13, line 72: it would be rare to hit this case, but should we return a proper error here instead of panic?

tools/advisory-checker/src/models.rs
```diff
@@ -0,0 +1,180 @@
...
+fn is_integer(s: &str) -> bool {
+    s.parse::<i64>().is_ok()
+}
+
+#[nutype(
+    validate(predicate = is_integer),
+    default = DEFAULT_EPOCH,
+    derive(Debug, Clone, Deserialize, PartialEq, Display, AsRef, Default)
+)]
+pub struct Epoch(String);
```
ginglis13, line 31: Maybe

```suggestion
pub struct Epoch(u64);
```

?

tools/advisory-checker/src/error.rs
```diff
@@ -0,0 +1,44 @@
...
+    #[snafu(display("Failed to read directory '{}'", path.display()))]
+    ReadDir {
+        path: PathBuf,
+        source: std::io::Error,
+    },
+
+    #[snafu(display("Failed to read directory entry"))]
+    ReadDirEntry { source: std::io::Error },
+
+    #[snafu(display("{} Advisory violations found.", violations.len()))]
```
ginglis13, line 42: When an advisory is not valid, users will see e.g.

```
1 Advisory violations found.
```

but shouldn't we also report the violations themselves?

tools/advisory-checker/src/models.rs
```diff
@@ -0,0 +1,180 @@
...
+
+#[derive(Debug, Clone)]
+pub struct EV {
```
ginglis13, line 36: maybe `EpochVersion` or add a docstring here

tools/advisory-checker/src/error.rs
piyush-jena: This is the kind of output we will get:
```
#10 ERROR: process "/bin/sh -c if [ -d \"/host/advisories/\" ]; then         /host/build/tools/advisory-checker             --packages-dir /host/packages/             --advisories-dir /host/advisories/;     else         echo \"Advisories directory not found, skipping check.\";     fi" did not complete successfully: exit code: 1
  ------
   > [rpmbuild 4/7] RUN --mount=target=/host     if [ -d "/host/advisories/" ]; then         /host/build/tools/advisory-checker             --packages-dir /host/packages/             --advisories-dir /host/advisories/;     else         echo "Advisories directory not found, skipping check.";     fi:
  2.684 Error: 3 Advisory violations found.
  2.684 
  2.684 Advisory violations found:
  2.684 Advisory Product 'libexpat' has no package and latest advisory 'BRSA-ja1ptrrpkcey' is not end-of-life
  2.684 BRSA ID: BRSA-ng7athla9aog
  2.684  package name: aws-iam-authenticator
  2.684  package version: 1:0.7.10
  2.684  min. expected version: 2:0.7.5
  2.684 Advisory Product 'containerd-2.0' has no package and latest advisory 'BRSA-s6xothqqu5vw' is not end-of-life
 ```

File: tools/advisory-checker/src/main.rs (hunk 61 above)
Comment (piyush-jena) [reply tools/advisory-checker/src/main.rs:23]: It is used here - `I: IntoIterator<Item=S>` in 
```
fn command<I, S>(bin_path: &str, args: I) -> Result<String>
where
    I: IntoIterator<Item = S>,
    S: AsRef<OsStr>,
```

File: tools/advisory-checker/src/models.rs (hunk 63 above)
Comment (piyush-jena) [reply tools/advisory-checker/src/models.rs:31]: The intention for using Epoch(String)  is because `patched-epoch` in BRSAs is a String. We do validate that its an integer. 
```
#[nutype(
    validate(predicate = is_integer),
    default = DEFAULT_EPOCH,
    derive(Debug, Clone, Deserialize, PartialEq, Display, AsRef, Default)
)]
pub struct Epoch(String);
```

ginglis13 (approved):

tools/advisory-checker/src/models.rs
piyush-jena: Added comments to describe that it is short for Epoch, Version and why we didn't use Release.

KCSesh (approved):

PR #607, opened 2025-12-04
feat: add --quiet flag to reduce build output

cbgbt:
Add a --quiet/-q flag that suppresses informational messages while preserving warnings and errors.

Changes:
- Add --quiet flag to CLI arguments
- Thread quiet parameter through build and make commands
- Conditionally add --quiet to cargo-make invocations
- Set TWOLITER_QUIET env var for embedded tools
- Adjust pubsys/testsys log levels based on quiet mode
- Update init_logger to respect quiet flag

In quiet mode:
- Suppresses cargo-make INFO messages
- Suppresses twoliter INFO logs
- Sets embedded tool log levels to warn
- Still shows all warnings and errors
- Completely silent on successful builds

Normal mode behavior is unchanged.

**Testing done:**
* [x] Ensure error messages still look right under various failure modes in `--quiet` mode
See [this gist](https://gist.github.com/cbgbt/b77a5883427ef00c3b51b9d907fd92a2) for a before/after comparison.

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/Makefile.toml
+++ b/twoliter/embedded/Makefile.toml
@@ -52,7 +52,7 @@ PUBLISH_INFRA_CONFIG_PATH = "${BUILDSYS_ROOT_DIR}/Infra.toml"
 # Default repo to read from PUBLISH_INFRA_CONFIG_PATH
 PUBLISH_REPO = "default"
 
-PUBLISH_LOG_LEVEL = "info"
+PUBLISH_LOG_LEVEL = { script = ["if [ \"${TWOLITER_QUIET}\" = \"1\" ]; then echo warn; else echo info; fi"], condition = { env_not_set = ["PUBLISH_LOG_LEVEL"] } }
 
 # This can be overridden with -e to change the path to the file containing SSM
 # parameter templates.  This file determines the parameter names and values
@@ -142,7 +142,7 @@ TESTSYS_STARTING_COMMIT = { script = ["git describe --tag ${TESTSYS_STARTING_VER
 TESTSYS_TESTS_DIR = "${BUILDSYS_ROOT_DIR}/tests"
 TESTSYS_TEST_CONFIG_PATH = "${BUILDSYS_ROOT_DIR}/Test.toml"
 
-TESTSYS_LOG_LEVEL = "info"
+TESTSYS_LOG_LEVEL = { script = ["if [ \"${TWOLITER_QUIET}\" = \"1\" ]; then echo warn; else echo info; fi"], condition = { env_not_set = ["TESTSYS_LOG_LEVEL"] } }
 
 [env.development]
 # Certain variables are defined here to allow us to override a component value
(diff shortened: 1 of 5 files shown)


swagatbora90:
Why not just allow users to set the log level directly?

cbgbt:
@swagatbora90 I think the log level can already be controlled. What I really want here is to suppress or dramatically reduce the amount of "informational" stdout printing, which isn't all routed through a logging module.

In my mind, the stdout printing is meant for user consumption and the logging is more of a diagnostic output.

twoliter/embedded/Makefile.toml
```diff
@@ -142,7 +142,7 @@ TESTSYS_STARTING_COMMIT = { script = ["git describe --tag ${TESTSYS_STARTING_VER
 TESTSYS_TESTS_DIR = "${BUILDSYS_ROOT_DIR}/tests"
 TESTSYS_TEST_CONFIG_PATH = "${BUILDSYS_ROOT_DIR}/Test.toml"
 
-TESTSYS_LOG_LEVEL = "info"
+TESTSYS_LOG_LEVEL = { script = ["if [ -n \"${TWOLITER_QUIET}\" ]; then echo warn; else echo info; fi"] }
```
yeazelm: Doesn't this "throw away" the value of `TESTSYS_LOG_LEVEL`? Ideally we would check if its set, use that, then check `TWOLITER_QUIET`, then default to info?

cbgbt: Very nice catch!

twoliter/embedded/Makefile.toml
```diff
@@ -52,7 +52,7 @@ PUBLISH_INFRA_CONFIG_PATH = "${BUILDSYS_ROOT_DIR}/Infra.toml"
 # Default repo to read from PUBLISH_INFRA_CONFIG_PATH
 PUBLISH_REPO = "default"
 
-PUBLISH_LOG_LEVEL = "info"
+PUBLISH_LOG_LEVEL = { script = ["if [ -n \"${TWOLITER_QUIET}\" ]; then echo warn; else echo info; fi"], condition = { env_not_set = ["PUBLISH_LOG_LEVEL"] } }
```
ytsssun: `-n` checks if `$TWOLITER_QUIET` exists. But the implementation sets "1" or "0" for it. Should we check the value here directly instead?

twoliter/src/cmd/make.rs
```diff
@@ -62,8 +63,10 @@ impl Make {
             .env("CARGO_HOME", self.cargo_home.display().to_string())
             .env("TWOLITER_TOOLS_DIR", toolsdir.display().to_string())
             .env("BUILDSYS_VERSION_IMAGE", project.release_version())
+            .env("TWOLITER_QUIET", if global_opts.quiet { "1" } else { "0" })
             .makefile(makefile_path)
             .project_dir(project.project_dir())
+            .quiet(global_opts.quiet)
```
ytsssun: Q - looking at the code, it seems that the `CargoMake.quiet` field is not really used? Only the env var was used.

cbgbt:
^ addresses feedback from @ytsssun. Nice catches!

yeazelm (approved):

ytsssun (approved):

PR #613, opened 2025-12-09
tools: include AWS SDK code and message in errors

ginglis13:
**Description of changes:**

Wrap all AWS SDK errors in DisplayErrorContext to provide the full call chain. Some of the existing errors, such as the ssm module's GetParametersSnafu, will provide the source error as "ServiceError" with no additional details or unwrapping of the error chain. See https://docs.aws.amazon.com/sdk-for-rust/latest/dg/error-handling.html#displayErrorContext

**Testing done:**

`make build`

Runtime test:

Here's a test with the new change:

```
./target/release/pubsys --infra-config-path /tmp/test-infra.toml validate-ssm --expected-parameters-path /tmp/expected-params.json
19:22:29 [INFO] Parsing Infra.toml file
19:22:29 [INFO] Found infra config at path: /tmp/test-infra.toml
19:22:29 [INFO] Parsing expected parameters file
19:22:29 [INFO] Parsed expected parameters file
19:22:29 [INFO] Retrieving SSM parameters
19:22:29 [INFO] Retrieving SSM parameters in us-west-2
19:22:29 [ERROR] Failed to retrieve images in region us-west-2: Failed to fetch SSM parameters by path  in us-west-2: ValidationException: 1 validation error detected: Value '' at 'path' failed to satisfy constraint: Member must have length greater than or equal to 1
19:22:29 [INFO] Validating SSM parameters
+-----------+---------+-----------+---------+------------+-------------+
| String    | correct | incorrect | missing | unexpected | unreachable |
+-----------+---------+-----------+---------+------------+-------------+
| us-west-2 | 0       | 0         | 0       | 0          | 1           |
+-----------+---------+-----------+---------+------------+-------------+
```

That same test with 0.14.0 pubsys:

```
 ./target/release/pubsys --infra-config-path /tmp/test-infra.toml validate-ssm --expected-parameters-path /tmp/expected-params.json
19:23:38 [INFO] Parsing Infra.toml file
19:23:38 [INFO] Found infra config at path: /tmp/test-infra.toml
19:23:38 [INFO] Parsing expected parameters file
19:23:38 [INFO] Parsed expected parameters file
19:23:38 [INFO] Retrieving SSM parameters
19:23:38 [INFO] Retrieving SSM parameters in us-west-2
19:23:39 [ERROR] Failed to retrieve images in region us-west-2: Failed to fetch SSM parameters by path  in us-west-2: service error
19:23:39 [INFO] Validating SSM parameters
+-----------+---------+-----------+---------+------------+-------------+
| String    | correct | incorrect | missing | unexpected | unreachable |
+-----------+---------+-----------+---------+------------+-------------+
| us-west-2 | 0       | 0         | 0       | 0          | 1           |
+-----------+---------+-----------+---------+------------+-------------+
```

where I have as these test input files:

```
# /tmp/expected-params.json
{
  "us-west-2": {
    "/aws/service/bottlerocket/test-param": "expected-value"
  }
}
```

and

```
# /tmp/test-infra.toml
[aws]
regions = ["us-west-2"]
```

**Terms of contribution:**

### DIFF
--- a/tools/error-utils/src/lib.rs
+++ b/tools/error-utils/src/lib.rs
@@ -1,0 +1,114 @@
+#![allow(clippy::disallowed_types)]
+
+use aws_smithy_types::error::display::DisplayErrorContext;
+use aws_smithy_types::error::metadata::ProvideErrorMetadata;
+use std::error::Error as StdError;
+use std::fmt;
+
+/// Shadow the AWS SDK error type.
+pub type SdkError<E, R = ::aws_smithy_runtime_api::client::orchestrator::HttpResponse> =
+    ::aws_smithy_runtime_api::client::result::SdkError<E, R>;
+
+/// Wrapper around AWS SDK errors that ensures consistent formatting
+pub struct AwsSdkError<E, R = ::aws_smithy_runtime_api::client::orchestrator::HttpResponse>(
+    pub SdkError<E, R>,
+)
+where
+    E: ProvideErrorMetadata + StdError + 'static;
+
+impl<E> fmt::Display for AwsSdkError<E>
+where
+    E: ProvideErrorMetadata + StdError + 'static,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", format_aws_sdk_error(&self.0))
+    }
+}
+
+impl<E> fmt::Debug for AwsSdkError<E>
+where
+    E: ProvideErrorMetadata + StdError + 'static,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Display::fmt(self, f)
+    }
+}
+
+impl<E> StdError for AwsSdkError<E>
+where
+    E: ProvideErrorMetadata + StdError + 'static,
+{
+    fn source(&self) -> Option<&(dyn StdError + 'static)> {
+        Some(&self.0)
+    }
+}
+
+impl<E> From<SdkError<E>> for AwsSdkError<E>
+where
+    E: ProvideErrorMetadata + StdError + 'static,
+{
+    fn from(error: SdkError<E>) -> Self {
+        AwsSdkError(error)
+    }
+}
+
+impl<E> AwsSdkError<E>
+where
+    E: ProvideErrorMetadata + StdError + 'static,
+{
+    /// Get the service error from the SDK error, if available
+    pub fn as_service_error(&self) -> Option<&E> {
+        self.0.as_service_error()
+    }
+}
+
+/// Helper function to format AWS SDK errors as "code: message" or fall back to full context
+///
+/// # Examples
+///
+/// With code only: "AccessDenied"
+/// With message only: "Request timeout occurred"
+/// With neither: Falls back to DisplayErrorContext for full error details
+pub fn format_aws_sdk_error<E>(sdk_error: &SdkError<E>) -> String
+where
+    E: ProvideErrorMetadata + StdError + 'static,
+{
+    match (sdk_error.code(), sdk_error.message()) {
+        (Some(code), Some(message)) => format!("{code}: {message}"),
+        (Some(code), None) => code.to_string(),
+        (None, Some(message)) => message.to_string(),
+        (None, None) => format!("{}", DisplayErrorContext(sdk_error)),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use aws_sdk_ssm::operation::get_parameters::GetParametersError;
+    use aws_smithy_types::error::metadata::ErrorMetadata;
+
+    #[test]
+    fn test_error_metadata_extraction() {
+        // Test that our logic works with error metadata
+        let error_metadata = ErrorMetadata::builder()
+            .code("InvalidParameter")
+            .message("Parameter is invalid")
+            .build();
+
+        let service_error = GetParametersError::generic(error_metadata);
+
+        // Verify the error has the expected code and message
+        assert_eq!(service_error.code(), Some("InvalidParameter"));
+        assert_eq!(service_error.message(), Some("Parameter is invalid"));
+    }
+
+    #[test]
+    fn test_aws_sdk_error_wrapper_display() {
+        let sdk_error: SdkError<GetParametersError> = SdkError::timeout_error("request timeout");
+        let wrapper = AwsSdkError(sdk_error);
+
+        // The wrapper should use our format_aws_sdk_error function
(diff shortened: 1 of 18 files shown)


qianxjcraig (approved):

ginglis13:
^ force push take a less verbose approach: provide either the code + message from the SDK error, fall back to the entire DisplayErrorContext() wrapped error.

sam-berning (approved):

tools/pubsys/src/error_utils.rs
```diff
@@ -0,0 +1,55 @@
+use aws_sdk_ssm::error::{DisplayErrorContext, ProvideErrorMetadata, SdkError};
+use std::error::Error as StdError;
+
+/// Helper function to format AWS SDK errors as "code: message" or fall back to full context
```
cbgbt: Do you think you could give an example in this docstring? I don't know what to expect.

cbgbt (commented):
Nice fix! I wonder if there's some way for us to categorically resolve this. How can we never mess it up?

One idea I have is to set up a clippy lint forbidding us from using `SdkError` directly, and to instead force us to use a new type which `impl`s `Display` using this method. I've used this to make clippy scan for "forbidden functions" -- you can actually see it in Twoliter's clippy config.

Mind trying that out and seeing if you can get it to work?

ginglis13:
^ force push fixes the unit test added

ginglis13:
@cbgbt I have some changes prepared that do the following to implement the clippy lint approach to enforce a new `AwsSdkError` wrapper of the `SdkError`:


1. Create a `tools/error-utils` crate with `AwsSdkError<E>` wrapper that implements Display, Debug, Error, and From<SdkError<E>> and includes the `format_aws_sdk_error` helper

2. Add clippy lint configuration to forbid direct `SdkError` usage

3. Migrate all AWS SDK error handling across `pubsys` and `testsys`:
   - Replace `SdkError` types with `AwsSdkError` wrapper
   - Convert snafu .context() calls to .map_err() -> this might be the part I'm least excited about. Instead of using Snafu's .context(), we have to directly map_err to the new wrapper. If we wanted to retain the .context() syntax, we'd have to directly reference the source AWS SDK error in the snafu, which would not appease the new lint. It's a tradeoff - we lose the ergonomic .context() method for AWS SDK errors, but we gain lint compliance and the wrapper error with the consistent error message formatting.
   - Use the wrapper's Display implementation directly in display strings

It's a fair amount of increase in the scope of this PR - working to get these changes segmented into logical commits, but this is the general approach

ginglis13:
^ force push adds 4 commits, based on comment in https://github.com/bottlerocket-os/twoliter/pull/613#issuecomment-3634857885

1. Add a new wrapper type for AWS SDK errors with enhanced error message formatting
2.  Use this type in pubsys
3. Use this type in testsys
4. Add a new clippy lint for enforcing that we use the error_utils type rather than AWS SDK error types

tools/pubsys/src/aws/ssm/mod.rs
```diff
@@ -254,7 +254,7 @@ pub(crate) async fn run(args: &Args, ssm_args: &SsmArgs) -> Result<()> {
 
 fn is_aws_service_access_denied_error(result: &ssm::Result<HashMap<SsmKey, String>>) -> bool {
     if let Err(ssm::error::Error::GetParameters { source, .. }) = result {
-        if let Some(err) = source.as_service_error() {
+        if let Some(err) = source.0.as_service_error() {
```
cbgbt: nit: perhaps we should re-expose this function within our wrapper?

tools/pubsys/Cargo.toml
```diff
@@ -12,6 +12,7 @@ async-stream.workspace = true
 async-trait.workspace = true
 aws-config.workspace = true
 aws-credential-types.workspace = true
+error-utils = { workspace = true }
```
cbgbt: nit: breaks from existing style

tools/testsys/Cargo.toml
```diff
@@ -12,6 +12,7 @@ publish = false
 [dependencies]
 async-trait.workspace = true
 aws-config.workspace = true
+error-utils = { workspace = true }
```
cbgbt: nit: breaks from existing style

cbgbt (approved):
Very nice!

ginglis13:
^ force push fixes some import style inconsistencies and exposes `as_service_error()` on the `AwsSdkError` type

cbgbt (approved):

tools/error-utils/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    #[test]
+    fn test_aws_sdk_error_wrapper_display() {
+        let sdk_error: SdkError<GetParametersError> = SdkError::timeout_error("request timeout");
+        let wrapper = AwsSdkError(sdk_error);
+
+        // The wrapper should use our format_aws_sdk_error function
+        let result = wrapper.to_string();
+        assert!(result.len() > 0);
+        assert!(result.contains("timeout") || result.contains("Timeout"));
```
sam-berning: nit: I'm a bit confused by the intent of this assert, is it this loose to allow for SSM changing the timeout error implementation? I'd think we could assert an exact match for the message.

tools/pubsys/src/aws/ami/mod.rs
```diff
@@ -450,7 +456,7 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
                 error!(
                     "Copy to {} failed: {}",
                     region,
-                    e.into_service_error().code().unwrap_or("unknown")
+                    e.0.into_service_error().code().unwrap_or("unknown")
```
sam-berning: nit: why not implement `into_service_error` on AwsSdkError? I don't love using the positional field here, I feel like we should make that private.

sam-berning (approved):
Nice work!

tools/pubsys/src/aws/ami/mod.rs
ginglis13: Adjusting this to 

```
e.as_service_error().and_then(|err| err.code()).unwrap_or("unknown")
```

tools/error-utils/src/lib.rs
ginglis13: I'll adjust to the exact message - I had it as a looser assertion thinking that message is subject to change, but the message has a pretty predictable structure so we can use the specific wording.

ginglis13:
^ force push uses `as_service_error` wrapper in place of direct tuple access across all changes

sam-berning (approved):

PR #617, opened 2025-12-16
rpm2img: add support for aliases for packages

piyush-jena:
**Description of changes:**
This PR allows creation of aliases for packages whose name is changed so as to support multiple versions of the same package. For example: bottlerocket-core-kit supports `containerd-1.7`, `containerd-2.0` and `containerd-2.1`. With this change, we allow creation of aliases so that we get additional entries in the application-inventory.json which can ensure that compliance solutions are able to properly track the old packages with new names.

**Testing done:**
These changes may or may not be valid for the actual repo but attempts to test corner cases. 
I added the following Obsoletes in `containerd-2.0`
```
Obsoletes: %{_cross_os}%{gorepo}
Obsoletes: %{_cross_os}%{gorepo}-1.7
Obsoletes: %{_cross_os}%{gorepo} <= 1.7.15
```

This the output I got after making the above change for containerd-2.0, building the package and mounting the rpm in a bottlerocket-sdk container. 
```
bash-5.2$ rpm -qp --queryformat '[%{OBSOLETES}\n]' /containerd-2.0/bottlerocket-containerd-2.0-2.0.7-1.1764962845.9b3a8b3a.br1.x86_64.rpm
bottlerocket-containerd
bottlerocket-containerd
bottlerocket-containerd-1.7
```
I used the above to write the logic. I added Obsoletes for `systemd-257` and `containerd-2.1` (to bin sub-package) as well and built a aws-k8s-1.34 ami.

Additional entries in application-inventory.json
```
   {
      "Name": "containerd",
      "Publisher": "bottlerocket-core-kit",
      "Version": "2.1.6",
      "Release": "1.1767894444.d2ba7912.br1",
      "Epoch": "0",
      "InstalledTime": "2026-01-13T20:50:12Z",
      "ApplicationType": "Unspecified",
      "Architecture": "x86_64",
      "Url": "https://github.com/containerd/containerd",
      "Summary": "Alias for bottlerocket-containerd-2.1-bin"
    },
    {
      "Name": "containerd-1.7",
      "Publisher": "bottlerocket-core-kit",
      "Version": "2.1.6",
      "Release": "1.1767894444.d2ba7912.br1",
      "Epoch": "0",
      "InstalledTime": "2026-01-13T20:50:12Z",
      "ApplicationType": "Unspecified",
      "Architecture": "x86_64",
      "Url": "https://github.com/containerd/containerd",
      "Summary": "Alias for bottlerocket-containerd-2.1-bin"
    },
    {
      "Name": "containerd-2.1",
      "Publisher": "bottlerocket-core-kit",
      "Version": "2.1.6",
      "Release": "1.1767894444.d2ba7912.br1",
      "Epoch": "0",
      "InstalledTime": "2026-01-13T20:50:12Z",
      "ApplicationType": "Unspecified",
      "Architecture": "x86_64",
      "Url": "https://github.com/containerd/containerd",
      "Summary": "An industry-standard container runtime"
    },
    {
      "Name": "systemd",
      "Publisher": "bottlerocket-core-kit",
      "Version": "257.9",
      "Release": "1.1767894444.d2ba7912.br1",
      "Epoch": "0",
      "InstalledTime": "2026-01-13T20:50:12Z",
      "ApplicationType": "Unspecified",
      "Architecture": "x86_64",
      "Url": "https://www.freedesktop.org/wiki/Software/systemd",
      "Summary": "Alias for bottlerocket-systemd-257"
    },
    {
      "Name": "systemd-252",
      "Publisher": "bottlerocket-core-kit",
      "Version": "257.9",
      "Release": "1.1767894444.d2ba7912.br1",
      "Epoch": "0",
      "InstalledTime": "2026-01-13T20:50:12Z",
      "ApplicationType": "Unspecified",
      "Architecture": "x86_64",
      "Url": "https://www.freedesktop.org/wiki/Software/systemd",
      "Summary": "Alias for bottlerocket-systemd-257"
    },
    {
      "Name": "systemd-257",
      "Publisher": "bottlerocket-core-kit",
      "Version": "257.9",
      "Release": "1.1767894444.d2ba7912.br1",
      "Epoch": "0",
      "InstalledTime": "2026-01-13T20:50:12Z",
      "ApplicationType": "Unspecified",
      "Architecture": "x86_64",
      "Url": "https://www.freedesktop.org/wiki/Software/systemd",
      "Summary": "System and Service Manager"
    },
```

containerd-2.1
```
%package bin
Summary: An industry-standard container runtime's binaries
Provides: %{name}(binaries)
Requires: (%{_cross_os}image-feature(no-fips) and %{name})
Conflicts: (%{_cross_os}image-feature(fips) or %{name}-fips-bin)
Obsoletes: %{_cross_os}%{gorepo}
Obsoletes: %{_cross_os}%{gorepo}-1.7
Obsoletes: %{_cross_os}%{gorepo} <= 1.7.15
```
systemd-257
```
Obsoletes: %{_cross_os}systemd
Obsoletes: %{_cross_os}systemd-252
```

**Terms of contribution:**

### DIFF
--- a/tests/integration-tests/src/appinventory.rs
+++ b/tests/integration-tests/src/appinventory.rs
@@ -76,6 +76,8 @@ async fn create_bob(version: &str) -> TempDir {
     cmd!(
         git,
         "clone",
+        "--depth",
+        "1",
         "-b",
         version,
         "https://github.com/bottlerocket-os/bottlerocket.git",
@@ -232,4 +234,76 @@ fn build_and_fetch_application_inventory(
         "twoliter did not generate application-inventory.json"
     );
     std::fs::read_to_string(&aif).expect("failed to read application-inventory.json file")
+}
+const SINGLE_OBSOLETE_PATCH: &str = include_str!("patches/single-obsolete.patch");
+const MULTIPLE_OBSOLETES_PATCH: &str = include_str!("patches/multiple-obsoletes.patch");
+
+fn apply_patch(bob_path: &Path, patch: &str) {
+    let git = which("git").expect("failed to find git");
+    cmd!(git, "-C", bob_path, "apply", "-")
+        .stdin_bytes(patch.as_bytes())
+        .run()
+        .expect("failed to apply patch");
+}
+
+#[tokio::test]
+async fn test_application_inventory_single_obsolete() {
+    // We clone v1.52.0 bob to ensure the patches are applied reliably
+    let bob_version = "v1.52.0";
+    let bob_src = create_bob(bob_version).await;
+
+    // To test aliases, we add a single obsoletes in settings-defaults package
+    apply_patch(bob_src.path(), SINGLE_OBSOLETE_PATCH);
+
+    // Build and get the application inventory file contents with this twoliter
+    let aif = build_and_fetch_application_inventory(TWOLITER_PATH, bob_src.path());
+    let inventory: InventoryView =
+        serde_json::from_str(&aif).expect("failed to deserialize inventory");
+
+    // In this case an extra entry is added in the application-inventory
+    assert!(
+        inventory
+            .content
+            .iter()
+            .any(|c| c.name == "bottlerocket-test-obsolete-pkg" && c.publisher == "Bottlerocket"),
+        "expected test-obsolete-pkg in inventory from Obsoletes"
+    );
+}
+
+#[tokio::test]
+async fn test_application_inventory_multiple_obsoletes() {
+    // We clone v1.52.0 bob to ensure the patches are applied reliably
+    let bob_version = "v1.52.0";
+    let bob_src = create_bob(bob_version).await;
+    apply_patch(bob_src.path(), MULTIPLE_OBSOLETES_PATCH);
+
+    // To test aliases, we add 2 obsoletes in settings-defaults package
+    let aif = build_and_fetch_application_inventory(TWOLITER_PATH, bob_src.path());
+    let inventory: InventoryView =
+        serde_json::from_str(&aif).expect("failed to deserialize inventory");
+
+    // In this case 2 extra entries are added in the application-inventory
+    let settings_defaults_publisher = inventory
+        .content
+        .iter()
+        .find(|c| c.name == "bottlerocket-settings-defaults")
+        .map(|c| c.publisher.clone())
+        .expect("expected bottlerocket-settings-defaults in inventory");
+
+    assert!(
+        inventory
+            .content
+            .iter()
+            .any(|c| c.name == "bottlerocket-test-obsolete-one"
+                && c.publisher == settings_defaults_publisher),
+        "expected test-obsolete-one in inventory from Obsoletes"
+    );
+    assert!(
+        inventory
+            .content
+            .iter()
+            .any(|c| c.name == "bottlerocket-test-obsolete-two"
+                && c.publisher == settings_defaults_publisher),
+        "expected test-obsolete-two in inventory from Obsoletes"
+    );
 }
--- a/twoliter/embedded/rpm2img
+++ b/twoliter/embedded/rpm2img
@@ -166,6 +166,8 @@ fi
 INSTALL_TIME="$(date -u +%Y-%m-%dT%H:%M:%SZ)"
 rpm -iv --ignorearch --root "${ROOT_MOUNT}" "${PACKAGE_DIR}"/*.rpm
 
+PKG_PREFIX="bottlerocket-"
+
 # Inventory installed packages.
 INVENTORY_QUERY="\{\"Name\":\"%{NAME}\"\
 ,\"SourcePackage\":\"%{SOURCERPM}\"\
@@ -177,7 +179,8 @@ INVENTORY_QUERY="\{\"Name\":\"%{NAME}\"\
 ,\"ApplicationType\":\"%{GROUP}\"\
 ,\"Architecture\":\"%{ARCH}\"\
 ,\"Url\":\"%{URL}\"\
-,\"Summary\":\"%{Summary}\"\}\n"
+,\"Summary\":\"%{Summary}\"\
+,\"Alias\":\"[%{Obsoletes},]\"\}\n"
 
 # shellcheck disable=SC2312 # Array is validated elsewhere.
 mapfile -t installed_rpms <<<"$(rpm -qa --root "${ROOT_MOUNT}" \
@@ -186,14 +189,18 @@ mapfile -t installed_rpms <<<"$(rpm -qa --root "${ROOT_MOUNT}" \
 # Wrap installed_rpms mapfile into json.
 INVENTORY_DATA="$(jq --raw-output . <<<"${installed_rpms[@]}")"
 # Sort by package name and add 'Content' as top-level.
-INVENTORY_DATA="$(jq --slurp 'sort_by(.Name)' <<<"${INVENTORY_DATA}" | jq '{"Content": .}')"
+# We convert comma separated Aliases to a json array.
+INVENTORY_DATA="$(jq --slurp '
+  map(.Alias |= (if . and . != "" then (split(",") | map(select(. != ""))) else [] end))
+  | sort_by(.Name)
+  | {Content: .}' <<<"${INVENTORY_DATA}")"
 
 INVENTORY_BEFORE_PARENTS="$(jq '.Content | length' <<<"${INVENTORY_DATA}")"
 INVENTORY_DATA="$(jq '
   .Content |= (
     . + map(
       (.Version) as $v | (.Release) as $r | (.SourcePackage | sub("-" + $v + "-" + $r + "\\.src\\.rpm$";"")) as $p |
-      . + {Name: $p, Summary: ($p + " (source package)")}
+      . + {Name: $p, Summary: ("Source Package for " + $p)}
     )
     | unique_by(.Name)
     | map(del(.SourcePackage))
@@ -223,12 +230,12 @@ for kit in "${kits[@]}"; do
   kit_path="${EXTERNAL_KITS_PATH}/${kit_vendor}/${kit}/${ARCH}"
 
   # Query the kit repo of RPMs for all package names.
-  kit_inventory_query="%{NAME}\n"
+  kit_inventory_query="%{NAME}\n%{Obsoletes}"
   # shellcheck disable=SC2312 # Array is validated elsewhere.
   mapfile -t KIT_PKGS <<<"$(dnf -q --repofrompath \
     kit,file://"${kit_path}" \
     --repo=kit repoquery \
-    --queryformat "${kit_inventory_query}" | grep -v "^$")"
+    --queryformat "${kit_inventory_query}" | grep -v "^$" | awk '{print $1}')"
   # Convert the bash array of kit packages to a JSON array.
   kit_list="$(\
     jq \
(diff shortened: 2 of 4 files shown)


twoliter/embedded/rpm2img
```diff
@@ -176,7 +175,8 @@ INVENTORY_QUERY="\{\"Name\":\"%{NAME}\"\
 ,\"ApplicationType\":\"%{GROUP}\"\
 ,\"Architecture\":\"%{ARCH}\"\
 ,\"Url\":\"%{URL}\"\
-,\"Summary\":\"%{Summary}\"\}\n"
+,\"Summary\":\"%{Summary}\"\
+,\"Alias\":\"[%{PROVIDES},]\"\}\n"
```
bcressey: `Obsoletes` is the canonical field for this sort of behavior in RPM, when one package replaces another that may have been previously installed. We don't have a current use for it in Bottlerocket since there's no package manager at runtime.

With `containerd-1.7` as an example:
```
diff --git a/packages/containerd-1.7/containerd-1.7.spec b/packages/containerd-1.7/containerd-1.7.spec
index 2aa494802..4d572889b 100644
--- a/packages/containerd-1.7/containerd-1.7.spec
+++ b/packages/containerd-1.7/containerd-1.7.spec
@@ -53,6 +53,7 @@ Requires: %{name}(binaries)

 Provides: %{_cross_os}%{gorepo} = %{package_priority_epoch}:
 Conflicts: %{_cross_os}%{gorepo}
+Obsoletes: %{_cross_os}%{gorepo}
```

twoliter/embedded/rpm2img
```diff
@@ -237,10 +237,22 @@ for kit in "${kits[@]}"; do
...
+    ]
+  ) |
+  .Content |= map(del(.Alias))
+' <<<"${INVENTORY_DATA}")"
+
 # Verify we successfully inventoried some RPMs.
 INVENTORY_COUNT="$(jq '.Content | length' <<<"${INVENTORY_DATA}")"
 RPM_COUNT="$(find "${PACKAGE_DIR}" -maxdepth 1 -type f -name '*.rpm' | wc -l)"
-if [[ "${INVENTORY_COUNT}" -ne "${RPM_COUNT}" ]]; then
+if [[ "$((INVENTORY_COUNT - ALIAS_COUNT))" -ne "${RPM_COUNT}" ]]; then
```
bcressey: nit: might be more readable to add the RPM and alias counts together and compare with inventory count.

twoliter/embedded/rpm2img
```diff
@@ -176,7 +175,8 @@ INVENTORY_QUERY="\{\"Name\":\"%{NAME}\"\
 ,\"ApplicationType\":\"%{GROUP}\"\
 ,\"Architecture\":\"%{ARCH}\"\
 ,\"Url\":\"%{URL}\"\
-,\"Summary\":\"%{Summary}\"\}\n"
+,\"Summary\":\"%{Summary}\"\
+,\"Alias\":\"%{Obsoletes}\"\}\n"
```
ginglis13: Can you share output of this query? As I understand it, `Obsoletes` could contain multiple packages - which if so, you'll need to adjust your `jq` to account for that.

piyush-jena:
Updated the logic to accomodate the above comments. Testing corner cases are detailed in the description above.

twoliter/embedded/rpm2img
```diff
@@ -187,6 +187,30 @@ INVENTORY_DATA="$(jq --raw-output . <<<"${installed_rpms[@]}")"
...
+# Add alias entries from Obsoletes field.
+# Alias is formatted as "[pkg1,pkg2,]" - we split and create an entry for each.
+# We strip the bottlerocket- prefix because these Alias packages are not available in KIT_PKGS variable.
+ORIG_COUNT=$(jq '.Content | length' <<<"${INVENTORY_DATA}")
+
+INVENTORY_DATA=$(jq '
+  .Content |= (
+    . + [
+      .[] |
+      select(.Alias != null and .Alias != "(none)" and .Alias != "[]") |
```
cbgbt: How does Alias end up being `"(none)"`?

twoliter/embedded/rpm2img
```diff
@@ -187,6 +187,30 @@ INVENTORY_DATA="$(jq --raw-output . <<<"${installed_rpms[@]}")"
 # Sort by package name and add 'Content' as top-level.
 INVENTORY_DATA="$(jq --slurp 'sort_by(.Name)' <<<"${INVENTORY_DATA}" | jq '{"Content": .}')"
 
+# Add alias entries from Obsoletes field.
+# Alias is formatted as "[pkg1,pkg2,]" - we split and create an entry for each.
+# We strip the bottlerocket- prefix because these Alias packages are not available in KIT_PKGS variable.
+ORIG_COUNT=$(jq '.Content | length' <<<"${INVENTORY_DATA}")
+
+INVENTORY_DATA=$(jq '
```
cbgbt: @cezar-r had a mechanism he was using to run integration tests against some of these scripts in https://github.com/bottlerocket-os/twoliter/pull/623

Could we possibly find a way to function-ize this and call it with a set of sample inputs from a test?

twoliter/embedded/rpm2img
piyush-jena:
```
bash-5.2$ rpm -qp --queryformat '%{OBSOLETES}\n' glibc/bottlerocket-glibc-2.42-1.1764962845.9b3a8b3a.br1.x86_64.rpm
(none)
bash-5.2$ rpm -qp --queryformat '%{OBSOLETES}\n' containerd-2.0/bottlerocket-containerd-2.0-2.0.7-1.1764962845.9b3a8b3a.br1.x86_64.rpm 
bottlerocket-containerd
```

If no Obsoletes are present and we didn't use square brackets in the query. If it were present, then only first would have been output. 

```
bash-5.2$ rpm -qp --queryformat '[%{OBSOLETES},]\n' containerd-2.0/bottlerocket-containerd-2.0-2.0.7-1.1764962845.9b3a8b3a.
br1.x86_64.rpm 
bottlerocket-containerd,bottlerocket-containerd,bottlerocket-containerd-1.7,
```
This is why we use `[%{OBSOLETES},]`. It will be an impossible value in this case. This would output `pkg1,pkg2,pkg3,...` and so on. Let me test if we can get rid of the ltrim and rtrim of square brackets. It seems like they are not in the output.

piyush-jena: I tested it and got rid of (none) and got rid of the removal of square brackets (which will never exist).

tests/integration-tests/src/appinventory.rs
```diff
@@ -205,3 +205,66 @@ fn build_and_fetch_application_inventory(
...
+}
+
+#[tokio::test]
+async fn test_application_inventory_single_obsolete() {
+    // We clone the latest release bob
+    let bob_version = find_latest_version("bottlerocket").await;
+    let bob_src = create_bob(bob_version.as_str()).await;
+
+    // To test aliases, we add a single obsoletes in settings-defaults package
+    apply_patch(bob_src.path(), SINGLE_OBSOLETE_PATCH);
```
cbgbt: I'd prefer we test this by creating a local workspace fixture if we can -- otherwise we'll need to regenerate these patches based on upstream changes.

tests/integration-tests/src/appinventory.rs
```diff
@@ -205,3 +205,66 @@ fn build_and_fetch_application_inventory(
...
+    let bob_version = find_latest_version("bottlerocket").await;
+    let bob_src = create_bob(bob_version.as_str()).await;
+    apply_patch(bob_src.path(), MULTIPLE_OBSOLETES_PATCH);
+
+    // To test aliases, we add 2 obsoletes in settings-defaults package
+    let aif = build_and_fetch_application_inventory(TWOLITER_PATH, bob_src.path());
+    let inventory: InventoryView =
+        serde_json::from_str(&aif).expect("failed to deserialize inventory");
+
+    // In this case 2 extra entries are added in the application-inventory
```
cbgbt, line 285: I think we need to assert that besides these two fields being present the contents are equivalent, no?

tests/integration-tests/src/appinventory.rs
piyush-jena: I couldn't create one that could be used to build an image. I have locked the version of bob to v1.52.0.

tests/integration-tests/src/appinventory.rs
piyush-jena: I didn't add it since it would require an extra bob build and `test_twoliter_application_inventory` does 2 builds and verifies the content of the application_inventory.

cbgbt (approved):

twoliter/embedded/rpm2img
```diff
@@ -239,8 +264,8 @@ done
 # Verify we successfully inventoried some RPMs.
 INVENTORY_COUNT="$(jq '.Content | length' <<<"${INVENTORY_DATA}")"
 RPM_COUNT="$(find "${PACKAGE_DIR}" -maxdepth 1 -type f -name '*.rpm' | wc -l)"
-if [[ "${INVENTORY_COUNT}" -ne "${RPM_COUNT}" ]]; then
-  echo "Inventory of RPMs does not match what was expected: '${INVENTORY_COUNT}/${RPM_COUNT}'" >&2
+if [[ "${INVENTORY_COUNT}" -ne "$((RPM_COUNT + ALIAS_COUNT))" ]]; then
+  echo "Inventory of RPMs does not match what was expected: '${INVENTORY_COUNT}/$((RPM_COUNT + ALIAS_COUNT))'" >&2
```
koooosh: nit: Label variables in the output error message: 
```
echo "Inventory count mismatch: ${INVENTORY_COUNT} inventory entries vs ${RPM_COUNT} RPMs + ${ALIAS_COUNT} aliases (expected $((RPM_COUNT + ALIAS_COUNT)))" >&2
```

Right now, it just prints numbers like: `6/7` which is unclear

koooosh (approved):

KCSesh (approved):

twoliter/embedded/rpm2img
```diff
@@ -204,6 +205,30 @@ INVENTORY_DATA="$(jq '
...
+
+INVENTORY_DATA=$(jq '
+  .Content |= (
+    . + [
+      .[] |
+      select(.Alias != null and .Alias != "") |
+      . as $orig |
+      # We trim the trailing comma from Alias to arrive at pkg1,pkg2
+      (.Alias | rtrimstr(",") | split(",")[] | select(. != "")) |
+      sub("^bottlerocket-";"") |
```
bcressey: This doesn't seem to match the prefix handling logic later, and therefore seems either incorrect or a likely source of future bugs.

The comment implies that this is the best we can do, but I don't agree with that - we could maintain a map of packages that are in `KIT_PKGS` with a list of corresponding obsoletes.

twoliter/embedded/rpm2img
```diff
@@ -204,6 +205,30 @@ INVENTORY_DATA="$(jq '
...
+ORIG_COUNT=$(jq '.Content | length' <<<"${INVENTORY_DATA}")
+
+INVENTORY_DATA=$(jq '
+  .Content |= (
+    . + [
+      .[] |
+      select(.Alias != null and .Alias != "") |
+      . as $orig |
+      # We trim the trailing comma from Alias to arrive at pkg1,pkg2
+      (.Alias | rtrimstr(",") | split(",")[] | select(. != "")) |
```
bcressey: What happens if we don't strip the trailing comma? Seems like the last element of the split would be filtered out by the final select.

piyush-jena: True. Got rid of it. It wasn't necessary.

twoliter/embedded/rpm2img
piyush-jena: I updated the inventory query used by dnf to get the Aliases in the `KIT_PKGS` list. I am using the first word only so it gets rid of " <= 1.7.15" if in case we have an Obsoletes entry of the form `Obsoletes: %{_cross_os}%{gorepo} <= 1.7.15`.  I was able to verify that the entries have other fields correct too.

KCSesh:
Might be nice to actually update the Summary to mention what it is an alias to, so that the application json, doesn't require a spec sheet lookup

```
   {
      "Name": "containerd",
      "Publisher": "bottlerocket-core-kit",
      "Version": "2.1.6",
      "Release": "1.1767894444.d2ba7912.br1",
      "Epoch": "0",
      "InstalledTime": "2026-01-13T10:37:48Z",
      "ApplicationType": "Unspecified",
      "Architecture": "x86_64",
      "Url": "https://github.com/containerd/containerd",
      "Summary": "Alias for containerd-2.1"
    },
    {
      "Name": "containerd-1.7",
      "Publisher": "bottlerocket-core-kit",
      "Version": "2.1.6",
      "Release": "1.1767894444.d2ba7912.br1",
      "Epoch": "0",
      "InstalledTime": "2026-01-13T10:37:48Z",
      "ApplicationType": "Unspecified",
      "Architecture": "x86_64",
      "Url": "https://github.com/containerd/containerd",
      "Summary": "Alias for containerd-2.1"
      },
       {
      "Name": "containerd-2.1",
      "Publisher": "bottlerocket-core-kit",
      "Version": "2.1.6",
      "Release": "1.1767894444.d2ba7912.br1",
      "Epoch": "0",
      "InstalledTime": "2026-01-13T10:37:48Z",
      "ApplicationType": "Unspecified",
      "Architecture": "x86_64",
      "Url": "https://github.com/containerd/containerd",
      "Summary": "An industry-standard container runtime's binaries"
      }
  
```

twoliter/embedded/rpm2img
```diff
@@ -222,13 +245,14 @@ for kit in "${kits[@]}"; do
   # Set the path inside the build container to the kit RPMs and repo.
   kit_path="${EXTERNAL_KITS_PATH}/${kit_vendor}/${kit}/${ARCH}"
 
-  # Query the kit repo of RPMs for all package names.
-  kit_inventory_query="%{NAME}\n"
+  # Query the kit repo of RPMs for all package names and the Aliases (Obsoletes)
+  # The Aliases if exists already have a new-line character at the end
+  kit_inventory_query="%{NAME}\n%{Obsoletes}"
```
bcressey: There's an undocumented assumption here that all package names are unique across all kits, and therefore that the first kit repo where a given package name appears is its actual source.

That feels problematic for Obsoletes where we might have three packages (containerd-1.7, containerd-2.0, containerd-2.1) that all obsolete the same original package (containerd).

Downstream, non-kit builds might also want to use aliases - e.g. `firecracker-containerd-2.1` that has an obsoletes for `containerd-2.1` or `containerd`.

This logic would split those up so the obsoletes from core kit would be used instead, so you'd end up with:
1. package firecracker-containerd, publisher Bottlerocket (package not found in kit)
2. alias containerd-2.1, publisher Bottlerocket (obsoletes not found in kit)
3. alias containerd, publisher bottlerocket-core-kit (obsoletes found in kit, but kit package not installed)

PR #625, opened 2025-12-19
add pcrsys to predict PCR values

bcressey:
**Issue number:**
Related: https://github.com/bottlerocket-os/bottlerocket/issues/4680

**Description of changes:**
Add the `pcrsys` tool to predict PCR values at build time.

**Testing done:**
#### Notes
- PCR 5 has multiple valid values due to GPT partition table variations
- PCR 11 has multiple valid values representing different boot stages

#### x86_64-aws-ecs-3-fips
- **Status**: PASS
- **Disk prediction**: PASS
- **AMI prediction**: PASS
- **AMI ID**: ami-06dde408a20fb8183
- **Mismatched PCRs**: None

| PCR | Predicted | Actual | Match |
|-----|-----------|--------|-------|
| 0 | 737f767a12f54e70eecbc868... | 737f767a12f54e70eecbc868... | ✅ |
| 2 | 3d458cfe55cc03ea1f443f15... | 3d458cfe55cc03ea1f443f15... | ✅ |
| 3 | 3d458cfe55cc03ea1f443f15... | 3d458cfe55cc03ea1f443f15... | ✅ |
| 5 | (72 values) | 0caef3f6e685e782816fe0ce... | ✅ (in set) |
| 6 | 3d458cfe55cc03ea1f443f15... | 3d458cfe55cc03ea1f443f15... | ✅ |
| 7 | 0215b629b95be153a9a15d7c... | 0215b629b95be153a9a15d7c... | ✅ |
| 10 | 000000000000000000000000... | 000000000000000000000000... | ✅ |
| 11 | (7 values) | 857516a0408cb6ef303e5e32... | ✅ (in set) |
| 12 | 000000000000000000000000... | 000000000000000000000000... | ✅ |
| 13 | 000000000000000000000000... | 000000000000000000000000... | ✅ |
| 14 | b1d778bd45c6af726c0e5d78... | b1d778bd45c6af726c0e5d78... | ✅ |
| 15 | 000000000000000000000000... | 000000000000000000000000... | ✅ |

#### aarch64-aws-ecs-3-fips
- **Status**: PASS
- **Disk prediction**: PASS
- **AMI prediction**: PASS
- **AMI ID**: ami-03c94418f6a89ed0f
- **Mismatched PCRs**: None

| PCR | Predicted | Actual | Match |
|-----|-----------|--------|-------|
| 0 | 737f767a12f54e70eecbc868... | 737f767a12f54e70eecbc868... | ✅ |
| 2 | 3d458cfe55cc03ea1f443f15... | 3d458cfe55cc03ea1f443f15... | ✅ |                                                                                                 | 3 | 3d458cfe55cc03ea1f443f15... | 3d458cfe55cc03ea1f443f15... | ✅ |
| 5 | (72 values) | 21fb8687e39e0220baa8d1e1... | ✅ (in set) |
| 6 | 3d458cfe55cc03ea1f443f15... | 3d458cfe55cc03ea1f443f15... | ✅ |
| 7 | 0215b629b95be153a9a15d7c... | 0215b629b95be153a9a15d7c... | ✅ |
| 10 | 000000000000000000000000... | 000000000000000000000000... | ✅ |
| 11 | (7 values) | 857516a0408cb6ef303e5e32... | ✅ (in set) |
| 12 | 000000000000000000000000... | 000000000000000000000000... | ✅ |
| 13 | 000000000000000000000000... | 000000000000000000000000... | ✅ |
| 14 | b1d778bd45c6af726c0e5d78... | b1d778bd45c6af726c0e5d78... | ✅ |
| 15 | 000000000000000000000000... | 000000000000000000000000... | ✅ |

#### x86_64-vmware-dev
- **Status**: PASS
- **Disk prediction**: PASS
- **AMI prediction**: N/A (VMware variant)
- **Mismatched PCRs**: None

| PCR | Predicted | Actual | Match |
|-----|-----------|--------|-------|
| 1 | 3d458cfe55cc03ea1f443f15... | 3d458cfe55cc03ea1f443f15... | ✅ |
| 2 | 3d458cfe55cc03ea1f443f15... | 3d458cfe55cc03ea1f443f15... | ✅ |
| 3 | 3d458cfe55cc03ea1f443f15... | 3d458cfe55cc03ea1f443f15... | ✅ |
| 4 | 7f26fb1af9759301841435d7... | 7f26fb1af9759301841435d7... | ✅ |
| 6 | 3d458cfe55cc03ea1f443f15... | 3d458cfe55cc03ea1f443f15... | ✅ |
| 7 | 8a36f6d1613525fe65a7a916... | 8a36f6d1613525fe65a7a916... | ✅ |
| 9 | 331c858ffa0ef3ed08ea94c8... | 331c858ffa0ef3ed08ea94c8... | ✅ |
| 10 | 000000000000000000000000... | 000000000000000000000000... | ✅ |
| 11 | (7 values) | 857516a0408cb6ef303e5e32... | ✅ (in set) |
| 12 | 000000000000000000000000... | 000000000000000000000000... | ✅ |
| 13 | 000000000000000000000000... | 000000000000000000000000... | ✅ |
| 14 | b1d778bd45c6af726c0e5d78... | b1d778bd45c6af726c0e5d78... | ✅ |
| 15 | 000000000000000000000000... | 000000000000000000000000... | ✅ |

**Terms of contribution:**


tools/pcrsys/src/efi.rs
```diff
@@ -0,0 +1,213 @@
...
+pub const EFI_GLOBAL_VARIABLE_GUID: Uuid =
+    Uuid::from_bytes(hex!("8be4df61 93ca 11d2 aa0d 00e098032b8c"));
+
+/// GUID for image security database variables (db, dbx).
+pub const EFI_IMAGE_SECURITY_DATABASE_GUID: Uuid =
+    Uuid::from_bytes(hex!("d719b2cb 3d3a 4596 a3bc dad00e67656f"));
+
+/// Collection of EFI variables from efi-vars.json.
+#[derive(Debug, Deserialize)]
+pub struct EfiVars {
```
cbgbt: Should we disallow unknown fields here?

tools/pcrsys/src/efi.rs
```diff
@@ -0,0 +1,213 @@
...
+}
+
+impl EfiVars {
+    /// Find a variable by name.
+    pub fn get(&self, name: &str) -> Option<&EfiVar> {
+        self.variables.iter().find(|v| v.name == name)
+    }
+
+    /// Validate that all required variables are present with correct GUIDs.
+    pub fn validate(&self) -> Result<()> {
```
cbgbt: Seems like state that we should punt to the type system. Can we do this on parse?

`#[serde(try_from = )]` is good for stuff like this if you don't want to pull in a more opinionated macro from crates.io.

tools/pcrsys/src/aws/uefi.rs
```diff
@@ -0,0 +1,241 @@
...
+    let decompressed = decompress_with_dict(&binary[16..])?;
+    parse_variables(&decompressed)
+}
+
+/// Decompress zlib data using the AWS preset dictionary.
+///
+/// The dictionary can only be set after zlib signals it needs one.
+fn decompress_with_dict(compressed: &[u8]) -> Result<Vec<u8>> {
+    let mut decomp = Decompress::new(true);
+    let mut output = vec![0u8; 256 * 1024];
```
cbgbt: Mind commenting what constrains the output to actually fit into this buffer? Is it that the efivars are a known structure size?

EDIT: I think I get it now - we need to decompress to an offset so `capacity()` isn't right, but we `truncate()` later to right-size. Can we comment on how we know the output is constrained by this range?

cbgbt (commented):
This is cool. Making my way through, will incrementally post comments. Just got the point where you actually start predicting PCRs.

tools/pcrsys/src/aws/zlib_dict.rs
```diff
@@ -0,0 +1,233 @@
...
+0EEBBQAAAAADDhMAAAAAAADCAQAAAAAACAEBAwoUAFNHweC++dI=";
+
+pub static AWS_ZLIB_DICT: LazyLock<Vec<u8>> = LazyLock::new(|| {
+    STANDARD
+        .decode(AWS_ZLIB_DICT_B64)
+        .expect("invalid base64 in AWS_ZLIB_DICT_B64")
+});
+
+#[cfg(test)]
+mod tests {
```
cbgbt, line 226: A test that base64 decodes and compares to a checksum of the source bytes could be a good one.

tools/pcrsys/src/aws/uefi.rs
```diff
@@ -0,0 +1,241 @@
...
+
+    if let Err(e) = &result {
+        if e.needs_dictionary().is_some() {
+            decomp
+                .set_dictionary(&AWS_ZLIB_DICT)
+                .whatever_context("set dictionary")?;
+            let in_off =
+                usize::try_from(decomp.total_in()).whatever_context("total_in overflow")?;
+            let out_off =
+                usize::try_from(decomp.total_out()).whatever_context("total_out overflow")?;
```
cbgbt: Would it be wise to check that we didn't silently truncate data by comparing `total_out()` to our buffer size?

twoliter/embedded/img2img
```diff
@@ -309,6 +309,15 @@ boot_usage=$(get_ext4_usage "${BOOT_IMAGE}" "BOOT-A")
 # Combine usage data into final JSON
 echo "${root_usage}" "${boot_usage}" | jq -s 'add | {disk_usage: .}' > "${OUTPUT_DIR}/artifact-metadata.json"
 
+# Predict PCR values after all signing is complete.
+if [[ "${UEFI_SECURE_BOOT}" == "yes" ]]; then
```
cbgbt, line 313: So basically, we unconditionally compute these. A question I am writing down (even if I just come back and answer myself is) under the circumstances where we have secure-boot variables, can we unconditionally generate useful pcr predictions?

I'm writing this comment before having read all the code, but I think the answer is partially "depends on if pcrsys supports the underlying platform."


ETA: Seemingly yes, and we have code for all of the platforms currently supported in bottlerocket-os/bottlerocket

tools/pcrsys/README.md
```diff
@@ -0,0 +1,86 @@
+# pcrsys
+
+A TPM Platform Configuration Register (PCR) prediction tool for Bottlerocket.
+
+## Overview
+
+`pcrsys` predicts TPM PCR values (SHA-256) for Bottlerocket images without requiring a running system. It analyzes disk images and EFI variables to compute what the PCR values would be after boot, enabling pre-registration of expected measurements for attestation policies.
```
cbgbt, line 7: This is extremely cool. Nice work!

tools/pcrsys/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Usage
+
+### From a local disk image
+
+```bash
+pcrsys disk --image /path/to/disk.img --efi-vars /path/to/efi-vars.json [--platform aws|vmware|metal]
+```
+
+### From an AWS AMI
```
cbgbt, line 17: I imagine this being useful and like it, but just making sure my understanding of this PR is complete - we aren't using this just yet, right?

I think a workflow I see being useful is basically unpacking an AMI to repack later. I imagine that was the incentive?

tools/pcrsys/src/predict.rs
```diff
@@ -0,0 +1,29 @@
+//! Core PCR prediction types and functions.
+
+use serde::Serialize;
+
+/// Collection of PCR predictions for JSON output.
+#[derive(Serialize)]
+pub struct PcrPredictions {
+    /// List of PCR records.
+    pub records: Vec<PcrRecord>,
+}
```
cbgbt, line 97: Is multiple records for the same index valid? If it's not, should we change this model to make such runtime usage impossible?

---

This is almost certainly not a good idea and perhaps unhinged, but I wonder if you could (ab)use `bon` (or something like it) to statically check that each PCR is only set once.

tools/pcrsys/src/pcrs/pcr1.rs
```diff
@@ -0,0 +1,38 @@
...
+use crate::platform::Platform;
+use crate::predict::{extend_pcr_separator, PcrRecord, PCR_INIT_VAL};
+
+/// Predict PCR 1 value.
+pub fn predict(platform: Platform) -> Option<PcrRecord> {
+    match platform {
+        Platform::Vmware => Some(PcrRecord::new(1, extend_pcr_separator(&PCR_INIT_VAL))),
+        Platform::Aws | Platform::Metal => None,
+    }
+}
```
cbgbt: The platform-specific nature of each prediction has me wondering if it would be easier to maintain this by moving the different pcr prediction implementations into some kind of Platform trait + a factory method on CLI start that parses the input into the right one.

tools/pcrsys/src/gpt.rs
```diff
@@ -16,6 +17,66 @@ pub const fn uuid_to_guid(uuid: [u8; 16]) -> [u8; 16] {
...
+#[derive(Debug, Clone, Copy)]
+pub struct GptPrio(u64);
+
+impl GptPrio {
+    /// Get priority value (bits 48-51).
+    pub fn priority(self) -> u64 {
+        self.0.get_bits(48..52)
+    }
+
+    /// Set priority value (bits 48-51). Panics if `priority > 15`.
```
cbgbt: nit: I somewhat prefer using `nutype` to validate such preconditions.

ETA: this would make it slightly easier to determine what went wrong if we validate the preconditions (e.g. error with more data, or even a panic when the faulty data is constructed rather than used.) It does complicate the API though. I'm open to a suggestion that the juice isn't worth the squeeze here.

tools/pcrsys/src/pcrs/pcr5.rs
```diff
@@ -0,0 +1,205 @@
...
+use crate::predict::{
+    extend_pcr, extend_pcr_separator, extend_pcr_string, PcrRecord, PCR_INIT_VAL,
+};
+use crate::Result;
+use sha2::{Digest, Sha256};
+
+/// Priority bit combinations: (priority, tries_left, successful)
+const COMBINATIONS: [(u8, u8, bool); 6] = [
+    (0, 0, false), // inactive
+    (0, 1, false), // inactive with tries
```
cbgbt, line 26: Note for posterity: Kiro suggested this may be invalid but after checking the `signpost` implementation it is decidedly valid.

tools/pcrsys/src/pcrs/pcr7.rs
```diff
@@ -0,0 +1,323 @@
...
+        return esl.to_vec();
+    }
+
+    let mut result = esl.to_vec();
+    let header_size = u32::from_le_bytes(esl[20..24].try_into().unwrap_or([0; 4])) as usize;
+    let sig_size = u32::from_le_bytes(esl[24..28].try_into().unwrap_or([0; 4])) as usize;
+
+    if sig_size == 0 {
+        return result;
+    }
```
cbgbt: Defense-in-depth question: Should we actually fail if the ESL is malformed? The value is derived from our EFI variables, which are extracted from our image.

If the data is malformed either:
* There's a serious problem with the EFI variables already
* There's a bug in our extraction code

I think failing on either case is probably better than passing, but then potentially having an attestion problem later.

tools/pcrsys/src/pcrs/pcr7.rs
```diff
@@ -0,0 +1,323 @@
...
+    // 5. dbx
+    let dbx_var_data =
+        generate_efi_variable_data(&EFI_IMAGE_SECURITY_DATABASE_GUID, "dbx", dbx_data);
+    pcr = extend_pcr(&pcr, &Sha256::digest(&dbx_var_data).into());
+
+    // 6. Separator
+    pcr = extend_pcr_separator(&pcr);
+
+    // 7. db authority - certificate used to verify shim
+    let db_sig = extract_first_sig_from_esl(db_data)?;
```
cbgbt, line 127: I spent some time noodling on this so leaving another posterity comment - what if there are more certs in the db? Is the signing cert always the first?

I think technically this code works because it assumes our sbkeys infrastructure which adds one cert to the db, the signing cert. That's probably fine.

```rust
// assumes single cert in db per Bottlerocket sbkeys profile
```

might be a good comment, but I wouldn't block on it.

cbgbt (commented):
Overall I found this implementation to be mechanical and straightforward to follow. The test coverage is great. Nice work!

I tend to lean more heavily on the type system and left some comments that suggest such changes, but I wouldn't block on those. Feel free to take em or leave em.

bcressey:
⬆️ force push to address feedback

Re-ran tests:
| Variant | Disk vs Runtime | AMI vs Disk | PCRs Checked |
|---------|:-:|:-:|---|
| x86_64-aws-ecs-3-fips | ✅ | ✅ | 0,2,3,5,6,7,10-15 |
| aarch64-aws-ecs-3-fips | ✅ | ✅ | 0,2,3,5,6,7,10-15 |
| x86_64-vmware-dev | ✅ | N/A | 1-4,6,7,9-15 |

Notes:
- PCRs 4 and 9 skipped for AWS variants (A/B boot partitions)
- PCR 5 matched 1 of 72 predicted GPT variations on both AWS variants
- PCR 11 matched boot phase 5/7 on all variants

tools/pcrsys/src/aws/uefi.rs
bcressey: I've replaced this with a loop that resizes the vector after the decompressor signals that it was too small. That avoids any hard-coded buffer sizes.

tools/pcrsys/src/aws/uefi.rs
bcressey: This should also be addressed by the new resize loop.

twoliter/embedded/img2img
bcressey: Yup, my goal was for this to "just work" for any variant with Secure Boot support. Without Secure Boot enforced, the predictions aren't meaningful.

tools/pcrsys/README.md
bcressey: > I imagine this being useful and like it, but just making sure my understanding of this PR is complete - we aren't using this just yet, right?

What I had in mind was for using `pcrsys` as a standalone tool to work with a public Bottlerocket AMI, even if we hadn't published `pcr-predictions.json` somewhere, or you didn't know where to find it, or you just wanted to compare results.

One wrinkle is that EC2 doesn't allow DescribeImageAttribute for AMIs that you don't own, including public AMIs, so you'd first need to copy it to an AMI of your own.

tools/pcrsys/src/predict.rs
bcressey: I've restructured the output to use a map of PCR indexes to values, and adjusted the interface for the prediction modules so they can provide both the key and the value.

tools/pcrsys/src/pcrs/pcr1.rs
bcressey: I tried this a few ways after refactoring the predictor interface, since the `PcrContext` struct made this more workable by streamlining the parameters.

Ultimately I felt that the logic for PCRs was sufficiently similar across platforms that I didn't see a good way to divide it into per-platform functions without either duplication or indirection into generic implementations.

tools/pcrsys/src/gpt.rs
bcressey: Added a `Nibble` type via `nutype`.

tools/pcrsys/src/pcrs/pcr7.rs
bcressey: Added an `EslHeader` type with more restrictive parsing.

bcressey:
⬆️ force push rebase onto develop to try to address the `Cargo.lock` issue

bcressey:
⬆️ force push - update `Cargo.lock` to try to fix CI.

bcressey:
⬆️ force push - finally fixed CI.

cbgbt (approved):
lgtm, nice work! Thanks also for fixing the CI failure.

mgsharm (approved):

tools/pcrsys/src/gpt.rs
```diff
@@ -0,0 +1,564 @@
...
+        let info = PartitionInfo {
+            number: 1,
+            start_lba: 2048,
+            end_lba: 4095,
+        };
+        assert_eq!(info.size_bytes(), (4095 - 2048 + 1) * 512);
+    }
+
+    /// Create a minimal valid GPT for testing.
+    fn mock_gpt() -> Vec<u8> {
```
arnaldo2792: Awesome! 🎉

tools/pcrsys/src/aws/uefi.rs
```diff
@@ -0,0 +1,298 @@
...
+///
+/// Decodes the base64 data, verifies the "AMZNUEFI" magic header,
+/// decompresses using zlib with the AWS preset dictionary, and
+/// extracts the PK, KEK, db, and dbx variables.
+pub fn parse_aws_uefi_data(b64_data: &str) -> Result<EfiVars> {
+    let binary = STANDARD
+        .decode(b64_data.trim())
+        .whatever_context("failed to decode base64")?;
+
+    ensure_whatever!(binary.len() >= UEFI_HEADER_SIZE, "UEFI data too short");
```
jmt-lab, line 25: nit(not-blocker): It would be nice to have proper error variants for these edge cases, I realize its a bunch of them but it can be nice. This can be a future improvement though

tools/pcrsys/src/pcrs/mod.rs
```diff
@@ -0,0 +1,17 @@
+//! PCR prediction modules for each TPM Platform Configuration Register.
```
jmt-lab, line 1: nit(not-blocker): It would be nice if potentially we can find a way to properly trait implement these numerical prediction methods, but i'm not sure really myself the right way to go about that after looking it seems pretty straight forward this way too.

jmt-lab (approved):

tools/pcrsys/src/aws/uefi.rs
bcressey: One of the things I like about snafu/whatever is it provides a way to level up into proper error variants if we need to, without paying for that complexity up front.

For now, since this isn't a library and there's a wide range of possible errors, collecting them into an enum felt like overkill. Happy to tackle it in the future if the usage pattern changes.

PR #637, opened 2026-02-06
chore: bump source package anchor version

### LINKED ISSUE #636: App Inventory integration test fails on crate dependency bumps
The `test_twoliter_application_inventory` integration test fails on any PR (including [Dependabot crate dependency bumps](https://github.com/bottlerocket-os/twoliter/pull/632) when HEAD `twoliter` generates different inventory output than the last released twoliter.

This currently happens because commit [eab2e7a](https://github.com/bottlerocket-os/twoliter/commit/eab2e7a09fb57761a1c5974424923ef08e92a9c4) (merged after v0.16.0) added source packages to the inventory via `twoliter/embedded/rpm2img`. The test's equality check fails since HEAD includes these new entries while the released v0.16.0 does not.

cezar-r:
**Issue number:**
https://github.com/bottlerocket-os/twoliter/issues/636

Closes #636

**Description of changes:**
Changes the `test_twoliter_application_inventory` test to verify that all packages in the reference inventory exist in the current inventory (subset check), rather than requiring exact equality. This catches regressions (missing packages) while allowing additions.

**Testing done:**
`cargo make test`

**Terms of contribution:**

### DIFF
--- a/tests/integration-tests/src/appinventory.rs
+++ b/tests/integration-tests/src/appinventory.rs
@@ -18,7 +18,7 @@ const EXPECTED_INVENTORY_PATH: &str =
 
 // Last Bottlerocket release that doesn't include source packages in application inventory.
 // For older releases, we check that reference inventory is a subset of current.
-const SOURCE_PACKAGE_INVENTORY_ANCHOR_VERSION: &str = "1.52.0";
+const SOURCE_PACKAGE_INVENTORY_ANCHOR_VERSION: &str = "1.55.0";
 
 #[derive(Serialize, Deserialize)]
 pub struct GithubRelease {


jmt-lab (approved):

tests/integration-tests/src/appinventory.rs
```diff
@@ -161,25 +151,13 @@ async fn test_twoliter_application_inventory() {
...
-                item
-            );
-        }
-    } else {
-        assert_eq!(
-            reference_set, current_set,
-            "twoliter generated different application inventory than last released twoliter"
+    // Verify current inventory contains all packages from reference inventory.
+    // This catches regressions (missing packages) while allowing additions.
+    for item in &reference_set.content {
```
sky1122: This will only catch 
-  Missing package from inventory

However, it will not be caught.
- Wrong package metadata
- extra package

piyush-jena:
I think this should work if we just change `const SOURCE_PACKAGE_INVENTORY_ANCHOR_VERSION: &str = "1.52.0";` to 1.54.0. The anchor version is intentional because we want to make sure the checks are tight and my change adds source packages (additional packages) and once bob uses the twoliter carrying my change the check will enforce equality once again. Making the check loose intentionally is not a good idea imo. @cbgbt and I discussed this specific change at length.

piyush-jena (approved):

sky1122 (approved):

Issue #253, opened 2024-05-31
Default to machine architecture for build variant, build kit, fetch and publish kit


webern:
Currently `twoliter build variant` and `twoliter build kit` default to `x86_64`. Instead they should default to the machine architecture (for example the architecture that `twoliter` was built for).

---
              Can we default to the arch that `twoliter` was built for? It'd be sort of surprising to get x86_64 builds by default on an aarch64 host.

_Originally posted by @bcressey in https://github.com/bottlerocket-os/twoliter/pull/249#discussion_r1622536784_

PR #638, opened 2026-02-10
fix(tests): fetch kits for host architecture

cbgbt:
**Description of changes:**
`make build` was failing on my aarch64 host. It turns out this is because we would fetch for x86_64 but then call twoliter without providing an arch later, which would cause it to default to the host architecture.

**Testing done:**
* `make build` now passes on my aarch64 dev host

**Terms of contribution:**


jmt-lab (approved):

piyush-jena (approved):

yeazelm:
Is this one of the things needed for https://github.com/bottlerocket-os/twoliter/issues/253?

yeazelm (approved):

Issue #641, opened 2026-02-12
Feature Request: Provide way to specify attributes for my variant rather than rely upon naming


yeazelm:
I'd like the ability to specify things like platform, family, or flavor in configuration rather than rely upon the naming scheme. Right now it uses the [bottlerocket-variant](https://github.com/bottlerocket-os/twoliter/blob/50ec9820415ef63dae1ac5e3dd92cd4b0e3a3556/tools/bottlerocket-variant/src/lib.rs#L91) parsing logic based upon name to determine things like platform and flavor but it would be great to have the option to override this in my variant configuration so I could name my variant whatever I like.

PR #647, opened 2026-03-10
Add ability to override variant properties

yeazelm:
**Issue number:**

Closes # https://github.com/bottlerocket-os/twoliter/issues/641

**Description of changes:**
Adds VariantOverrides struct with Option<String> fields for platform, runtime, family, version, and flavor. Includes from_cargo_toml() function to parse overrides from [package.metadata.build-variant] section. This enables the ability to set your variant name to nearly anything you'd like but retain attributes like platform, runtime, family, version, or flavor.

**Testing done:**
Built several different combinations to ensure the end to end workflow is intact:
### Setup
Created settings-defaults symlinks for custom variants:
```
yeazelm-fancy-1.33 -> vmware-k8s-1.33
yeazelm-neat-1.35 -> aws-k8s-1.35
```

### Test Variant builds to prove out existing still build and new ones can be overridden

| Variant | Type | Overrides | Expected Family |
|---------|------|-----------|-----------------|
| aws-ecs-3 | Standard | None | aws-ecs |
| aws-k8s-1.32 | Standard | None | aws-k8s |
| aws-k8s-1.34-nvidia-fips | Standard | None | aws-k8s |
| vmware-k8s-1.32 | Standard | None | vmware-k8s |
| yeazelm-fancy-1.33 | Custom | platform=vmware, runtime=k8s | vmware-k8s |
| yeazelm-neat-1.35 | Custom | platform=aws, runtime=k8s | aws-k8s |

### Build Results

| Variant | Cargo Build | Full Image Build | Notes |
|---------|-------------|------------------|-------|
| aws-ecs-3 | ✅ Success | ✅ | Standard variant |
| aws-k8s-1.32 | ✅ Success | ✅  | Standard variant |
| aws-k8s-1.34-nvidia-fips | ✅ Success | ✅  | Standard variant |
| vmware-k8s-1.32 | ✅ Success | ✅ .vmdk created | Full rebuild, 3m40s |
| yeazelm-fancy-1.33 | ✅ Success | ✅ | Custom variant with vmware overrides |
| yeazelm-neat-1.35 | ✅ Success | ❌ DNF error | Package resolution issue because the rpm for aws-k8s-1.35 doesn't include it|

The symlink in sources/settings-defaults/ only helps with building the settings content, but the RPM spec file (packages/settings-defaults/settings-defaults.spec) needs to have the custom variant name added to an existing subpackage's Requires list.

For yeazelm-neat-1.35 to work, I would need to add it to the aws-k8s-1.35 subpackage in the spec:

```spec
%package aws-k8s-1.35
Requires: (%{shrink:
           %{_cross_os}variant(aws-k8s-1.35) or
           %{_cross_os}variant(aws-k8s-1.35-fips) or
           %{_cross_os}variant(yeazelm-neat-1.35)    # <-- ADD THIS
           %{nil}})
```

**Terms of contribution:**

### DIFF
--- a/docs/design/custom-variant-considerations.md
+++ b/docs/design/custom-variant-considerations.md
@@ -1,0 +1,105 @@
+# Custom Variant Names and Variant Attribute Overrides
+
+### Background
+
+Bottlerocket variants can override attributes like `platform` and `runtime` to inherit build-time behavior from standard variant families. For example, a custom variant named `foo-k8s-1.35` might override:
+
+```toml
+[package.metadata.build-variant]
+platform = "vmware"
+runtime = "k8s"
+```
+
+This results in `family = aws-k8s`, which affects build-time behavior such as:
+- Conditional compilation via `cfg` attributes
+- AMI naming conventions
+- Platform-specific package inclusion
+
+However, these overrides do **not** automatically configure the settings-defaults RPM resolution.
+
+## The Settings-Defaults Problem
+
+The `settings-defaults` RPM package uses exact variant name matching via RPM dependencies. When the variant build runs:
+
+1. The `bottlerocket-metadata` RPM provides `variant(NAME)` based on your literal variant name (e.g., `variant(foo-k8s-1.35)`)
+2. The RPM resolver looks for a `settings-defaults` subpackage with a matching `Requires: variant(foo-k8s-1.35)`
+3. If no match exists, the build fails with: `Unable to find a match: bottlerocket-settings-defaults`
+
+Variant attribute overrides affect build-time behavior but do **not** affect RPM dependency resolution.
+
+## Required Steps for Custom Variants
+
+### 1. Create a Settings-Defaults Directory
+
+You must create a settings-defaults directory named after your **actual variant name**, not the family you're inheriting from.
+
+In the Bottlerocket repository under `sources/settings-defaults/`, create:
+
+```
+foo-k8s-1.35/
+├── Cargo.toml
+└── defaults.d/
+    ├── 10-defaults.toml → ../../shared-defaults/defaults.toml
+    ├── 15-aws-tuf.toml → ../../shared-defaults/aws-tuf.toml
+    ├── 20-aws-host-containers.toml → ../../shared-defaults/aws-host-containers.toml
+    └── 50-kubernetes-aws.toml → ../../shared-defaults/kubernetes-aws.toml
+```
+
+The `Cargo.toml` should define the crate name using underscores (dots are not allowed in crate names):
+
+```toml
+[package]
+name = "settings-defaults-foo-k8s-1_35"
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[build-dependencies]
+gensettings = { path = "../../gensettings" }
+```
+
+### 2. Use Symlinks to Inherit Shared Defaults
+
+The `defaults.d/` directory should contain numbered symlinks to files in `shared-defaults/`. The numbered prefix controls merge order (lower numbers are applied first).
+
+Choose symlinks based on your target family:
+
+| Family | Required Symlinks |
+|--------|------------------|
+| aws-k8s | `defaults.toml`, `aws-tuf.toml`, `aws-host-containers.toml`, `kubernetes-aws.toml` |
+| vmware-k8s | `defaults.toml`, `public-tuf.toml`, `public-host-containers.toml`, `kubernetes-vmware.toml` |
+| metal-k8s | `defaults.toml`, `public-tuf.toml`, `public-host-containers.toml`, `kubernetes-metal.toml` |
+| aws-ecs | `defaults.toml`, `aws-tuf.toml`, `aws-host-containers.toml`, `ecs.toml` |
+
+### 3. Update the RPM Spec
+
+Add your variant to `packages/settings-defaults/settings-defaults.spec`. You can either:
+
+**Option A: Create a new subpackage**:
+
+```spec
+%package foo-k8s-1.35
+Summary: Settings defaults for foo-k8s-1.35 variant
+Requires: %{_cross_os}variant(foo-k8s-1.35)
+Provides: %{_cross_os}settings-defaults(any)
+
+%description foo-k8s-1.35
+%{summary}.
+
+%files foo-k8s-1.35
+%{_cross_factorydir}/%{_cross_os}defaults.d/*
+```
+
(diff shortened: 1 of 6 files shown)


docs/design/custom-variant-considerations.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+**Option B: Add to an existing subpackage** (if settings are compatible):
+
+```spec
+%package aws-k8s-1.35
+...
+Requires: variant(aws-k8s-1.35)
+Requires: variant(aws-k8s-1.35-fips)
+Requires: variant(foo-k8s-1.35)  # Add your variant here
```
bcressey: This syntax doesn't look correct, should be `%{_cross_os}variant`.

docs/design/custom-variant-considerations.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+## Additional Considerations
+
+### FIPS Variants
+
+FIPS variants typically share settings-defaults with their non-FIPS counterparts. If you create `foo-k8s-1.35-fips`, you can add it to the same subpackage as `foo-k8s-1.35`.
```
bcressey: I'm a bit wary of this guidance; it seems to be assuming `bottlerocket-os/bottlerocket` conventions in many places, in a way that seems orthogonal to the override feature.

docs/design/custom-variant-considerations.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Background
+
+Bottlerocket variants can override attributes like `platform` and `runtime` to inherit build-time behavior from standard variant families. For example, a custom variant named `foo-k8s-1.35` might override:
+
+```toml
+[variant]
+platform = "aws"
+runtime = "k8s"
+```
```
bcressey, line 11: This doesn't seem like the right syntax to me.

tools/bottlerocket-variant/src/lib.rs
```diff
@@ -123,6 +124,25 @@ impl Variant {
         Variant::new(value)
     }
 
+    /// Create a new `Variant` from the `VARIANT` environment variable, applying any overrides
+    /// from the specified Cargo.toml manifest file.
+    ///
+    /// If the manifest file doesn't exist or has no `[package.metadata.bottlerocket-variant]`
+    /// section, no overrides are applied. Malformed TOML will return an error.
```
bcressey: `package.metadata.bottlerocket-variant` is owned by `buildsys` - it's weird for it to defer parsing of its own file to a different crate, vs. extending `manifest.rs`.

tools/bottlerocket-variant/src/lib.rs
```diff
@@ -440,3 +496,293 @@ fn parse_err() {
...
+    }
+
+    #[test]
+    fn parse_overrides_partial() {
+        let toml = r#"
+[package]
+name = "test"
+version = "0.1.0"
+
+[package.metadata.bottlerocket-variant]
```
bcressey: I'd somewhat prefer to put this in `build-variant` metadata

tools/buildsys/src/builder.rs
bcressey, line 285: Why don't we need to change `repack` also?

tools/buildsys/src/builder.rs
```diff
@@ -442,9 +444,16 @@ impl DockerBuild {
...
             image_layout.publish_image_sizes_gib();
 
+        let cargo_toml_path = args.common.cargo_manifest_dir.join("Cargo.toml");
         let variant = filename(args.common.cargo_manifest_dir);
 
         let v = Variant::new(&variant).context(error::VariantParseSnafu)?;
+        let cargo_toml_content = std::fs::read_to_string(&cargo_toml_path).unwrap_or_default();
+        let overrides = VariantOverrides::from_cargo_toml(&cargo_toml_content)
+            .context(error::VariantParseSnafu)?;
+        let v = v.with_overrides(&overrides);
```
bcressey, line 453: We've already parsed the manifest at this point, so we could do something like:
```
let v = Variant::new(...)
    .with_platform_override(manifest.variant_platform())
    .with_runtime_override(manifest.variant_runtime())
```

Where those methods could accept an `Option` and only apply the override if a value existed.

docs/design/custom-variant-considerations.md
yeazelm: I can remove, this does seem to be stepping into general guidance not really directed here.

docs/design/custom-variant-considerations.md
yeazelm: Being more specific it should be:
```
[package.metadata.bottlerocket-variant]
platform = "vmware"
runtime = "k8s"
version = "1.33"
```
I'll update to be correct.

docs/design/custom-variant-considerations.md
yeazelm: I'll update, I did this quickly and didn't look closely at the actual example, I'll check the code directly to get the syntax right.

tools/buildsys/src/builder.rs
yeazelm: I'll confirm again, but I believe the only uses are in rpm matching and repack doesn't work at that level, the variant_name is the same as variant as far as the artifacts needed for repack are concerned, but I'll confirm and report back.

tools/buildsys/src/builder.rs
yeazelm: Oh nice, I like this. I'll update to make it a bit cleaner.

tools/bottlerocket-variant/src/lib.rs
yeazelm: Yeah, fair enough, I'll shift to make this more naturally part of `buildsys`.

tools/bottlerocket-variant/src/lib.rs
yeazelm: I can move it to `build-variant`. I had originally tried to keep it separate but it makes sense to keep all these in one place in metadata.

tools/buildsys/src/builder.rs
yeazelm: I did confirm that all the changes we have don't apply in the `repack` flow.

yeazelm:
^ Updated with changes to move the code into buildsys so that the parsing all occurs in the same package.

jpculp (approved):

yeazelm:
^ Updated the changes to streamline the options and stick to one approach. Now it uses `with_overrides` and `VariantOverrides` for this behavior.

bcressey (approved):

PR #649, opened 2026-03-18
pubsys: add Debug-level error chain logging for target download failure

sky1122:
**Description of changes:**

Add Debug-level error chain logging for target download failures in validate_repo.

When target downloads fail, the error output only shows Display-formatted messages (e.g., "error decoding response body"), which hide actionable details like HTTP status codes and timeout flags. This is because `tough` stores inner errors as Box<dyn Error>, and snafu formats the chain using Display.

This change adds a log_error_chain helper that walks the .source() chain and logs each cause with Debug formatting ({:?}). Unlike Display, the &dyn Error reference returned by .source() reaches through the Box<dyn Error> to the
concrete type, exposing internal fields (e.g., reqwest status codes, error classifications) that Display omits.

Called at both error sites in download_target():
- TargetRead — when repo.read_target() fails
- Stream — when reading the target stream fails

**Testing done:**
build success

**Terms of contribution:**

### DIFF
--- a/tools/pubsys/src/repo/validate_repo/mod.rs
+++ b/tools/pubsys/src/repo/validate_repo/mod.rs
@@ -5,7 +5,7 @@ use crate::repo::{error as repo_error, repo_urls};
 use crate::{read_stream, repo, Args};
 use clap::Parser;
 use futures::{stream, StreamExt};
-use log::{info, trace};
+use log::{error, info, trace};
 use pubsys_config::InfraConfig;
 use snafu::{OptionExt, ResultExt};
 use std::io::Cursor;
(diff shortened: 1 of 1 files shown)


tools/pubsys/src/repo/validate_repo/mod.rs
```diff
@@ -56,6 +56,19 @@ async fn retrieve_targets(repo: &Repository) -> Result<(), Error> {
...
 }
 
+/// Walks an error's source chain, logging each cause with Debug formatting.
+/// This surfaces details (like HTTP status codes) that Display formatting hides.
+fn log_error_chain(err: &dyn std::error::Error) {
+    error!("Error details: {:?}", err);
+    let mut source = err.source();
+    let mut depth = 1;
+    while let Some(cause) = source {
+        error!("  Cause #{}: {:?}", depth, cause);
```
dhwaniserai, line 66: Can you remove this extra space at the start to maintain same error logging as others?

sky1122: The leading space is intentional to visually indent the cause chain in logs, making it easier to distinguish root causes from the top-level error.

jmt-lab (approved):

ginglis13 (approved):

PR #661, opened 2026-05-11
twoliter: generate builder-group rpm to provide group(builder)

piyush-jena:
**Description of changes:**
This is a dummy package that provides `group(builder)`. RPM 6.0 introduced automatic user() and group() dependency generation from file ownership changes declared in `%files` sections. In earlier versions the dependencies were weak, so they may be ignored. In RPM 6.0 they are Required.

This is safe because `group(builder)` is responsible for building bottlerocket. Its just that a package has to provide it to satisfy the requirement by the RPM.

**Testing done:**
Built Bottlerocket with a custom bottlerocket-sdk that carries Fedora-43 container.

**Terms of contribution:**

### DIFF
--- a/twoliter/embedded/build.Dockerfile
+++ b/twoliter/embedded/build.Dockerfile
@@ -71,6 +71,13 @@ RUN --mount=target=/host \
         echo "Advisories directory not found, skipping check."; \
     fi
 
+RUN --mount=target=/host \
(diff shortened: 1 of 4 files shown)


jmt-lab (approved):
LGTM. Lets get some backwarts compatibility testing though just so we know how to scope this change.

ginglis13 (approved):

PR #669, opened 2026-06-04
feat(buildsys): embed guest variant images via build-variant guest-images

jmt-lab:
Add a declarative `guest-images` field to a host variant's
`[package.metadata.build-variant]` manifest that embeds the disk-image
artifacts of one or more already-built guest variants into the host's
rootfs. This enables shipping variant-A images as content inside
variant-B (e.g. installer media, recovery partitions, AMI-bundled
secondaries) without introducing a new package type or CLI surface.

Manifest schema:

  [package.metadata.build-variant.guest-images]
  inner-variant = "/usr/share/bottlerocket/guests/inner"

  [build-dependencies]
  inner-variant = { path = "../inner-variant" }

`buildsys build-variant` validates each entry: the key must appear in
`[build-dependencies]`, the target must itself be a `build-variant`,
the value must be an absolute install path, and a variant may not
reference itself. Validated entries are threaded through a
`GUEST_IMAGES` build-arg (newline-delimited
`<guest>:<install_path>:build/images/<arch>-<guest>/<version>`) into the
`imgbuild` Dockerfile stage, where `rpm2img` (via the new
`guest-images-helper`) `cp -a`s each guest's image directory into
`${ROOT_MOUNT}<install_path>` after `rpm -iv` and before SELinux
relabel, so embedded files inherit correct labels.

Tests:
- Unit: `test_build_variant_with_guest_images`,
  `test_build_variant_without_guest_images`,
  `test_build_type_default_package_unchanged`, plus
  `validate_guest_image_entries` and `guest_image_variant_deps`
  coverage for invalid names/paths, self-reference, missing build-dep,
  transitive-only build-dep, and the happy path.
- Fixtures: `tests/projects/guest-images-kit` (host wrapper-variant
  embedding an inner-variant) and `guest-images-graph` (direct,
  transitive, and self-referential guest deps).
- Integration (`#[ignore]`, requires Docker):
  `test_twoliter_build_variant_consumes_guest_images` asserts the
  guest variant builds transitively and the host variant produces
  image artifacts.
**Terms of contribution:**


tools/buildsys/src/builder.rs
```diff
@@ -333,10 +336,41 @@ impl TargetBuildArgs {
...
             TargetBuildArgs::Repack(_) => BuildType::Repack,
+            TargetBuildArgs::ImagePackage(_) => BuildType::ImagePackage,
         }
     }
 }
 
+/// Default install location for the disk-image artifacts inside the produced RPM, when the
+/// user does not specify `install-dir` in the manifest.
+fn default_image_package_install_dir(variant: &str, version_full: &str) -> String {
+    format!("/usr/share/bottlerocket/variants/{variant}/{version_full}")
```
jmt-lab: We might need to adjust this

twoliter/embedded/guest-images-helper
```diff
@@ -0,0 +1,41 @@
...
+  local find_args=( "${src}" -maxdepth 1 \( -type f -o -type l \) \( -false )
+  local suffix
+  for suffix in "${GUEST_IMAGE_SUFFIXES[@]}"; do
+    find_args+=( -o -name "*.${suffix}" )
+  done
+  find_args+=( \) -print0 )
+  local copied=0 file
+  while IFS= read -r -d '' file; do
+    cp -a "${file}" "${dst}/"
+    copied=$((copied + 1))
```
ginglis13: ensure that `cp -a` succeeded before incrementing `copied`

tools/buildsys/src/manifest/error.rs
```diff
@@ -46,4 +46,52 @@ pub(super) enum Error {
         "The cargo package we are building, '{name}', could not be found in the graph"
     ))]
     RootDependencyMissing { name: String },
+
+    #[snafu(display(
+        "Variant '{name}' declares guest-images for itself ('{name}'). A variant cannot use its \
```
ginglis13: nit: 
```suggestion
        "Variant '{name}' declares guest-images for itself. A variant cannot use its \
```

tools/buildsys/src/manifest.rs
```diff
@@ -710,6 +882,9 @@ pub struct BuildVariant {
     pub supported_arches: Option<HashSet<SupportedArch>>,
     pub kernel_parameters: Option<Vec<String>>,
     pub image_features: Option<HashMap<ImageFeature, bool>>,
+    /// Map of guest variant crate name -> absolute install path in this variant's root
+    /// filesystem. See struct-level docs for details.
```
ginglis13:
```suggestion
    /// Map of guest variant crate name -> absolute install path in this variant's root
    /// filesystem.
```

twoliter/embedded/rpm2img
```diff
@@ -166,6 +172,33 @@ fi
...
 rpm -iv --ignorearch --root "${ROOT_MOUNT}" "${PACKAGE_DIR}"/*.rpm
 
+# Embed any declared guest variant images into the host rootfs. Each entry has the form
+# `<guest>:<install_path>:<host_image_dir>` separated by newlines. The host image dir is
+# expressed relative to the project root, which is mounted at /bypass.
+#
+# We deliberately copy only the bootable image artifacts (and their stable name symlinks),
+# not the build metadata that lives alongside them. Files such as `application-inventory
+# .json`, `artifact-metadata.json`, `pcr-predictions.json`, and the `*-sbom-*` SBOM files
+# describe the *guest* build and have no business inside the *host* rootfs; embedding them
```
arnaldo2792: PCR predictions might not, but the sboms yes. So do please include them. Ideally, the SBOM of the host should include it. Basically, treat it as another package that should appear in the host's SBOM.

twoliter/embedded/build.Dockerfile
```diff
@@ -575,3 +577,4 @@ RUN --mount=target=/host \
 FROM scratch AS repack
 ARG NOCACHE
 COPY --from=imgrepack /tmp/.${NOCACHE} /output/
+
```
ginglis13: newline added?

arnaldo2792 (commented):
First quick pass, this is the one line I would really like to keep as downstreams requested.

twoliter/embedded/build.Dockerfile
```diff
@@ -407,7 +408,8 @@ RUN --mount=target=/host \
       ${EROFS_ROOT_PARTITION:+--with-erofs-root-partition=yes} \
       ${UEFI_SECURE_BOOT:+--with-uefi-secure-boot=yes} \
       ${IN_PLACE_UPDATES:+--with-in-place-updates=yes} \
-      ${ENCRYPTED_STORAGE:+--with-encrypted-storage=yes} && \
+      ${ENCRYPTED_STORAGE:+--with-encrypted-storage=yes} \
+      --guest-images="${GUEST_IMAGES}" && \
```
ginglis13: nit: seems like we intentionally are grouping image features down here, this might be better places within lines 399-406

twoliter/embedded/rpm2img
ginglis13: @jmt this comment needs updating now that guest SBOMs are included in the host

twoliter/embedded/rpm2img
ginglis13: I want to flag this for discussion - how we might include guest OS application inventory in the host

One option is to merge the guest package inventory into the host application inventory. We can distinguish between guest OS and host OS packages clearly, such as a `-guest` suffix to all package names appearing in the one application-inventory that are from the guest OS. This will be necessary i.e. if names are shared between guest and host - we can't have overwriting entries.

The other option would be to provide guest inventory in the host as a completely separate file, but this wouldn't integrate out-of-the box with any tooling around inventory that expects just application-inventory.json.

twoliter/embedded/guest-images-helper
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+# Helpers for embedding guest-variant disk images into a host variant's rootfs.
+#
+# This file is sourced by `rpm2img` and by the buildsys integration tests. It is
+# safe to source on its own: it defines variables/functions only and performs no
+# destructive operations. Do not add top-level side effects here.
+
+# Suffixes produced by `compress_image` / `symlink_image` in `imghelper`. Keep
+# this list in sync with the artifacts emitted by an image build.
```
strategicpause: What feedback mechanism exists to tell you when this list is out of sync? For example, warnings?

twoliter/embedded/rpm2img
strategicpause: Playing devil's advocate a bit: Why not just merge them into a single list and not distinguish between the two? At the end of the day if ArtifactA-1.0 is present in both the guest and the host, we will need patch that host, and all artifacts regardless of where they come from. If one of the artifacts is missed, then ArtifactA-1.0 and ArtifactA-1.1 will show up in the resulting SBOM, right?

twoliter/embedded/rpm2img
jmt-lab: Comment updated!

twoliter/embedded/rpm2img
jmt-lab: We are merging into a single list now, the unique by is on the full package identification so strategicpause you are correct, ArtiactA-1.0 and ArtifactA-1.1 will show up

twoliter/embedded/guest-images-helper
jmt-lab: Added a check to this

Sparksssj (approved):

tools/buildsys/src/main.rs
```diff
@@ -270,11 +270,25 @@ fn build_variant(args: BuildVariantArgs) -> Result<()> {
     check_arch_support(manifest.info(), args.common.arch);
     validate_first_party_stack_or_warn(manifest.info())?;
 
+    // Resolve declared guest-images and emit a rerun-if-changed for each guest's image
+    // directory, so rebuilding a guest variant triggers a rebuild of this host variant.
+    let guest_images = manifest
```
sky1122, line 275: An EIF host variant that declares guest-images will build fine but silently embed nothing, since only rpm2img handles the flag and not rpm2eif. Should we reject that combination up front, the way repack already does for EIF?

tools/buildsys/src/manifest.rs
```diff
@@ -656,6 +759,58 @@ impl ManifestInfo {
...
+/// responsible for the graph-level "must be a direct build-dep" check.
+///
+/// The `:` and newline checks exist because the `GUEST_IMAGES` build-arg passed to the image
+/// stage is a newline-delimited list of `<guest>:<install_path>:<host_image_dir>` triples; a
+/// literal `:` or newline in either field would corrupt parsing in `rpm2img`. The `..`
+/// rejection prevents an install path from escaping the host rootfs target.
+fn validate_guest_image_entries(
+    host_name: &str,
+    guest_images: &BTreeMap<String, PathBuf>,
+) -> Result<()> {
```
vigh-m, line 780: Is it worth validating that the host image format is not `eif` here?

vigh-m (approved):

sky1122 (approved):

PR #679, opened 2026-06-24
pubsys: support publishing one AMI into multiple accounts per region

### LINKED ISSUE #678: pubsys: allow registration for multiple accounts in one region
**What I'd like:**

When registering a Bottlerocket AMI across several regions for release, I would like to maintain a static spec file which allows me to register to several accounts in the same region. This will allow teams with preprod and prod accounts to use one registration file / avoid rebuilding the AMI.

e.g.
```
[aws]
regions = ["us-west-2", "us-east-1"]
ssm_prefix = "/br-latest"

[aws.region.us-west-2]
role = "arn:aws:iam::777777777777:role/BottlerocketSourceRole" # pre-prod account
role = "arn:aws:iam::99999999999:role/BottlerocketSourceRole" # prod account

[aws.region.us-east-1]
role = "arn:aws:iam::88888888888:role/BottlerocketSourceRole"
```

Currently we are dynamically writing these pubsys files, and storing them in S3. This is not transparent or intuitive for developers to change, and forces us to rely on bash.

timjhh:
**Issue number:**

Closes #678

**Description of changes:**

pubsys: support publishing one AMI into multiple accounts per region
Previously pubsys assumed one account per region (a single role per
[aws.region.*]), so registering a single built AMI into separate
preprod/prod accounts in the same region required dynamically generated
per-account spec files.

This adds a `roles` list to [aws.region.*] alongside the existing
`role`. We register the AMI once in the base account, then copy + share
it into each additional target account (reusing the existing
modify_snapshots/modify_image + copy_image machinery), and publish a
separate set of SSM parameters per account.

The intermediate --ami-input JSON produced by `pubsys ami` and consumed
by `pubsys ssm` / `pubsys publish-ami` is now nested by account
(region -> account -> image). SSM get/set/validate run once per account
since each account has its own parameter store.

**Testing done:**

- `cargo test` passes for pubsys and pubsys-config packages.
- Added several tests to verify that `role` takes precedence over `roles`, we will allow an empty or missing `roles` vector, duplicate roles across both fields are de-duplicated and accepted, and that the modified `--ami-input` field works across the ssm tool.

**Terms of contribution:**

### DIFF
--- a/tools/pubsys-config/src/lib.rs
+++ b/tools/pubsys-config/src/lib.rs
@@ -138,11 +138,61 @@ pub struct AwsConfig {
     pub s3: Option<HashMap<String, S3Config>>,
 }
 
-/// AWS region-specific configuration
+/// A region's role configuration: either a single role or a list of roles, but not both. These
+/// are mutually exclusive by construction (a flattened, untagged enum), so an `Infra.toml` that
+/// specifies both `role` and `roles` for the same region fails to parse.
 #[derive(Debug, Deserialize, Serialize, PartialEq, Eq, Clone)]
+#[serde(rename_all = "kebab-case")]
+pub enum RoleConfig {
+    /// A single role to assume for this region. Assumed after the "global" `aws.role`, if that is
+    /// also specified. Serializes to/from the `role` key in Infra.toml.
+    Role(String),
+
+    /// Multiple roles to assume for this region, one per target account. Each role is assumed
+    /// independently (after the "global" `aws.role`), producing a separate set of AMIs / SSM
+    /// parameters per account. Serializes to/from the `roles` key in Infra.toml.
+    Roles(Vec<String>),
+}
+
+/// AWS region-specific configuration
+#[derive(Debug, Default, Deserialize, Serialize, PartialEq, Eq, Clone)]
 #[serde(deny_unknown_fields)]
 pub struct AwsRegionConfig {
-    pub role: Option<String>,
+    #[serde(flatten)]
+    pub role_config: Option<RoleConfig>,
+}
+
+impl AwsRegionConfig {
+    /// Returns the roles configured for this region as an ordered, de-duplicated list, or `None`
+    /// if no roles are configured (meaning "use the base/global credentials without assuming a
+    /// region-specific role").
+    pub fn all_roles(&self) -> Option<Vec<String>> {
+        let roles: Vec<&String> = match &self.role_config {
+            None => return None,
+            Some(RoleConfig::Role(role)) => vec![role],
+            Some(RoleConfig::Roles(roles)) => roles.iter().collect(),
+        };
+
+        let mut seen = std::collections::HashSet::new();
+        Some(
+            roles
+                .into_iter()
+                .filter(|role| seen.insert((*role).clone()))
+                .cloned()
+                .collect(),
+        )
+    }
+
+    /// Returns the configured role whose ARN targets the given account, by matching the account ID
+    /// embedded in a role ARN (`arn:aws:iam::<account>:role/...`). Returns `None` if no configured
+    /// role matches; for example when the AMI was published using the base/global credentials.
+    pub fn role_for_account(&self, account_id: &str) -> Option<String> {
+        let needle = format!(":{account_id}:");
+        self.all_roles()
+            .into_iter()
+            .flatten()
+            .find(|role| role.contains(&needle))
+    }
 }
 
 /// Location of signing keys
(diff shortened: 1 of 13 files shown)


tools/pubsys-config/src/lib.rs
```diff
@@ -142,7 +142,50 @@ pub struct AwsConfig {
...
+    #[serde(default)]
+    pub roles: Vec<String>,
+}
+
+impl AwsRegionConfig {
+    /// Returns every role configured for this region, combining the singular `role` and the
+    /// `roles` list, de-duplicated while preserving order (`role` first). If no roles are
+    /// configured, returns a single `None`, meaning "use the base/global credentials without
+    /// assuming a region-specific role".
+    pub fn all_roles(&self) -> Vec<Option<String>> {
```
sam-berning: This is a nit, but I think it's a bit more idiomatic for this to return an `Option<Vec<String>>`

tools/pubsys-config/src/lib.rs
```diff
@@ -142,7 +142,50 @@ pub struct AwsConfig {
 #[derive(Debug, Deserialize, Serialize, PartialEq, Eq, Clone)]
 #[serde(deny_unknown_fields)]
 pub struct AwsRegionConfig {
```
sam-berning: I'd prefer if role and roles were mutually exclusive. We should be able to achieve this with a flattened enum, for example:

```rust
#[derive(Debug, Deserialize, Serialize, Clone, Eq, PartialEq)]
enum RoleConfig {
    role(String),
    roles(Vec<String>),
}

#[derive(Debug, Deserialize, Serialize, PartialEq, Eq, Clone)]
#[serde(deny_unknown_fields)]
pub struct AwsRegionConfig {
    #[serde(flatten)]
    pub role_config: Option<RoleConfig>,
}
```

tools/pubsys/src/aws/ami/mod.rs
```diff
@@ -496,56 +571,79 @@ impl Image {
     }
 }
 
-/// Returns the set of account IDs associated with the roles configured for the given regions.
+/// Resolves the account ID reachable through the given STS client (in the given region, used only
+/// for error messages).
+async fn get_account_id(sts_client: &StsClient, region: &Region) -> Result<String> {
```
sam-berning, line 780: Can we parse the account ID out of the role ARN instead? Seems faster + doesn't require the STS client.

tools/pubsys/src/aws/ssm/mod.rs
```diff
@@ -197,61 +215,100 @@ pub(crate) async fn run(args: &Args, ssm_args: &SsmArgs) -> Result<()> {
...
+    for parameter in new_parameters {
+        parameters_by_account
+            .entry(parameter.account_id.clone())
+            .or_default()
+            .push(parameter);
+    }
 
-    trace!("Current SSM parameters: {current_parameters:#?}");
+    let mut any_changes = false;
+    for (account_id, account_parameters) in &parameters_by_account {
```
sam-berning: If using a different AWS account per region, this will end up breaking parallelism while publishing SSM parameters to multiple regions. Instead, each region will now run in serial since they each use their own accounts.

Can we implement the multi-account support within `get_parameters` / `set_parameters` / `validate_parameters` instead, to preserve parallelism?

tools/pubsys-config/src/lib.rs
timjhh: sure I can do that!

tools/pubsys/src/aws/ami/mod.rs
jpculp: +1

timjhh: will do

tools/pubsys-config/src/lib.rs
```diff
@@ -138,11 +138,67 @@ pub struct AwsConfig {
...
+/// are mutually exclusive by construction (a flattened, untagged enum), so an `Infra.toml` that
+/// specifies both `role` and `roles` for the same region fails to parse.
+// The variant names are lowercase because they have to match the keys in Infra.toml (`role` /
+// `roles`), and it's more common for TOML config to be lowercase.
+#[allow(non_camel_case_types)]
 #[derive(Debug, Deserialize, Serialize, PartialEq, Eq, Clone)]
+pub enum RoleConfig {
+    /// A single role to assume for this region. Assumed after the "global" `aws.role`, if that is
+    /// also specified.
+    role(String),
```
sam-berning: let's use `#[serde(rename = "role")]` (or better yet `#[serde(rename_all = "kebab-case")]`) here and keep the enum variants in pascal case

tools/pubsys/src/aws/validate_ssm/mod.rs
```diff
@@ -61,53 +63,76 @@ pub async fn validate(
...
+                    error!("Failed to retrieve images in region {}: {e}", key.region);
                     error::Error::UnreachableRegion {
-                        region: region.to_string(),
+                        region: key.region.to_string(),
                     }
                 }),
             )
         })
-        .collect::<HashMap<&Region, Result<_>>>();
+        .collect::<HashMap<Region, Result<_>>>();
```
sam-berning: this needs to be a `HashMap<RegionAccount, Result>>` or else errors may be swallowed if there are multiple accounts in a region

tools/pubsys-config/src/lib.rs
sam-berning: I still think this would be a good improvement if you get around to it!

tools/pubsys/src/aws/ami/mod.rs
timjhh: This has been completed in the new rev

tools/pubsys/src/aws/ssm/mod.rs
timjhh: This has been done in the new rev

tools/pubsys-config/src/lib.rs
timjhh: looked into it, now that this is an intermediary form of the accounts we are publishing to, we are using the Nonetype of the vector to fall back to use the globally configured roles. 

If we used `Option<Vec<String>>` we would only be able to specify 'no roles arns can be used' or 'some role arns can be used' instead of the current semantic meaning of 'targets to publish to' of the data type

tools/pubsys-config/src/lib.rs
timjhh: This has been done in a new rev

tools/pubsys-config/src/lib.rs
timjhh: actually nvm, I was able to do this!

tools/pubsys/src/aws/validate_ssm/mod.rs
timjhh: so true, threaded RegionAccount all the way through the workflow to address the issue

tools/pubsys-config/src/lib.rs
timjhh: done

tools/pubsys-config/src/lib.rs
```diff
@@ -138,11 +138,61 @@ pub struct AwsConfig {
...
+/// AWS region-specific configuration
+#[derive(Debug, Default, Deserialize, Serialize, PartialEq, Eq, Clone)]
 #[serde(deny_unknown_fields)]
 pub struct AwsRegionConfig {
-    pub role: Option<String>,
+    #[serde(flatten)]
+    pub role_config: Option<RoleConfig>,
+}
+
+impl AwsRegionConfig {
```
jmt-lab: **[P1] `roles = []` silently skips a region from publishing.**

When `Infra.toml` contains `roles = []`, `role_config` deserializes to `Some(RoleConfig::Roles(vec![]))` and `all_roles()` returns `Some(vec![])`. Downstream, `roles_for_region()` in `ami/mod.rs:740-749` matches on `Some(roles)` and emits an empty vector, so the region silently produces zero publish targets — no AMI registered, no SSM parameters published, no error reported.

The docstring on `all_roles` says it returns `None` when no roles are configured; `Some(vec![])` violates that contract. It also diverges from `role_config: None` (no `role`/`roles` at all), which correctly falls back to base credentials via the `None => vec![None]` arm.

**Fix — pick one:**
1. Reject `roles = []` at deserialization so config load fails with a clear error (best diagnostic).
2. Have `all_roles()` collapse `Some(vec![])` to `None` to honor its docstring.
3. Have `roles_for_region()` treat `Some(vec![])` the same as `None`.

Please add a `region_config_empty_roles_list` test locking in whichever behavior is chosen.

tools/pubsys-config/src/lib.rs
```diff
@@ -138,11 +138,61 @@ pub struct AwsConfig {
...
+    pub fn all_roles(&self) -> Option<Vec<String>> {
+        let roles: Vec<&String> = match &self.role_config {
+            None => return None,
+            Some(RoleConfig::Role(role)) => vec![role],
+            Some(RoleConfig::Roles(roles)) => roles.iter().collect(),
+        };
+
+        let mut seen = std::collections::HashSet::new();
+        Some(
+            roles
```
sam-berning, line 178: it's possible under the current configuration to specify two roles in the same account, which would lead to a confusing error when pubsys runs into a name collision - can we throw an error on deserialization if there are multiple roles using the same account in a region? might need a custom deserializer with `#[serde(deserialize_with ...)]`

tools/pubsys/src/aws/ssm/mod.rs
```diff
@@ -158,28 +162,36 @@ pub(crate) async fn run(args: &Args, ssm_args: &SsmArgs) -> Result<()> {
...
         // The HashMap `entry` API doesn't play well with `async`, so we use a match here instead.
-        let client_config = match aws_sdk_configs.get(region) {
+        let client_config = match aws_sdk_configs.get(&key) {
             Some(client_config) => client_config.clone(),
             None => {
-                let client_config = build_client_config(region, &base_region, &aws).await;
-                aws_sdk_configs.insert(region.clone(), client_config.clone());
+                let role = role_for_account(&aws, &key.region, &key.account_id);
+                let client_config =
+                    build_client_config_for_role(&key.region, &base_region, &aws, role.as_deref())
```
sam-berning, line 185: there is a sneaky breaking change here: `pubsys` previously allowed you to publish SSM parameters in one account that reference AMIs in a different account, with the SSM account's role specified in the `Infra.toml` used to publish the SSM parameters. (this was a happy accident since there was a 1:1 region-AMI mapping, so there was no confusion on which AMI should be referenced by the parameter)

if I'm understanding correctly, as this is written, it will only search for a role that matches the account ID of the AMI. the SSM account role in the Infra.toml will be ignored and it will fall back to the base creds.

I'll think a bit about how we can handle this. we might need to introduce the concept of a "primary" image in a region or something like that.

PR #680, opened 2026-06-26
repack-variant improvements

jpculp:
**Description of changes:**

Hardens the `img2img` repack script by:

- Replacing bare `dd` calls with a `write_partition()` helper that adds `fsync` and sha256 read-back comparison to catch silent corruption or truncation.
- Consolidating scattered inline debugfs heredoc blocks into a `debugfs_replace()` helper that verifies each file write and SELinux label immediately rather than batching error checks.
- Documenting the `repack-variant` workflow in the design doc.

**Tradeoffs:**

- Adds ~10-15s of repack time for hash verification across partition writes.

**Testing done:**

- [x] `repack-variant` on a secure-boot variant _(exercises all `debugfs_replace` and `write_partition` paths including EFI resign)_

**Terms of contribution:**

### DIFF
--- a/docs/design/README.md
+++ b/docs/design/README.md
@@ -540,6 +540,40 @@ After installing packages, the result of `dnf list installed` will be output to 
 This will allow maintainers to audit which kit each package was taken from.
 The rest of the variant image creation steps proceed the same way that they do now.
 
+## Twoliter Repack
+
+In the event that a maintainer wants to replace specific artifacts on an already-built variant image, such as the CA certificate bundle, TUF `root.json`, or Secure Boot keys, they can use the `repack-variant` target rather than rebuilding from source.
+
+The `repack-variant` target operates on whatever images exist in the local build artifacts.
+There are two ways to get them there:
+
+- `build-variant`: builds images from source.
+  Use this when the images haven't been published yet or you have local source changes.
+- `fetch-variant`: downloads previously published images from a TUF repository.
+  Use this when repacking a released set of images without rebuilding all packages.
+
+```sh
+# Repack published images:
+twoliter make fetch-variant
+twoliter make repack-variant
+
+# Or repack locally built images:
+twoliter make build-variant
+twoliter make repack-variant
+```
+
+The `repack-variant` target takes the input image, replaces the configured artifacts, re-signs where necessary, regenerates dm-verity, and produces the final image.
+
+For example, a maintainer behind a corporate proxy can add their internal CA to an existing image by pointing `BUILDSYS_CACERTS_BUNDLE_OVERRIDE` at a custom bundle:
+
+```sh
+twoliter make \
+  -e BUILDSYS_CACERTS_BUNDLE_OVERRIDE=/path/to/corporate-ca-bundle.crt \
+  repack-variant
+```
+
+Because `repack-variant` skips package builds entirely, it is significantly faster than a full `build-variant`.
+
 ## Publishing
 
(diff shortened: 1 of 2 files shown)


twoliter/embedded/img2img
```diff
@@ -41,15 +41,61 @@ cleanup() {
...
+  errlog="$(mktemp)"
+
+  cat <<EOF | debugfs -w -f - "${image}" 2>"${errlog}"
+rm ${target}
+write ${source} ${target}
+ea_set ${target} security.selinux system_u:object_r:os_t:s0
+EOF
+
+  local unexpected
+  unexpected="$(grep -Ev '^debugfs [0-9]+\.[0-9]+\.[0-9]+ \([^)]+\)$' "${errlog}" || true)"
```
jmt-lab, line 58: The new filter ^debugfs [0-9]+\.[0-9]+\.[0-9]+ \([^)]+\)$ matches today's e2fsprogs banner (debugfs 1.47.0 (5-Feb-2023)) but is tighter than the previous ^debugfs  filter. A future SDK bump that produces any of these would silently break every repack-variant with a misleading "debugfs produced unexpected output" error:
• 4-component versions (1.47.0.1)
• RC/WIP suffixes (1.47.1-WIP)
• Extra startup lines

Not sure though if we should be concerned by any of these corner cases though

twoliter/embedded/img2img
```diff
@@ -41,15 +41,61 @@ cleanup() {
...
-  # debugfs prints it's version info to STDERR, so we need to filter it out.
-  # shellcheck disable=SC2312  # grep returns an error code if no match.
-  if [[ "$(grep -vc '^debugfs ' "${stderr}")" -ne 0 ]]; then
-    grep -v '^debugfs ' "${stderr}" >&2
+# Replace a file in an ext4 image and verify the result.
+# Usage: debugfs_replace <image> <source> <target>
+debugfs_replace() {
+  local image="$1" source="$2" target="$3"
+  local errlog
+  errlog="$(mktemp)"
```
jmt-lab: put errlog under ${WORKDIR}

errlog="$(mktemp -p "${WORKDIR}")"

This way the existing cleanup trap removes it on abnormal exit, instead of leaking to /tmp.

docs/design/README.md
```diff
@@ -540,6 +540,29 @@ After installing packages, the result of `dnf list installed` will be output to
 This will allow maintainers to audit which kit each package was taken from.
 The rest of the variant image creation steps proceed the same way that they do now.
 
+## Twoliter Repack
```
jmt-lab, line 543: Two small improvements:
1. Add a one-liner distinguishing when to use fetch-variant vs. build-variant as the repack input.
2. Enumerate the replaceable artifacts (CA bundle, root.json, Secure Boot keys, grub.cfg) — they're already mentioned generically in the intro.

jmt-lab (approved):

docs/design/README.md
jpculp: 1. Sure thing.
2. I was hesitant to enumerate just because I'd hate for the doc to fall out of date if more customization are added down the line.

twoliter/embedded/img2img
jpculp: Actually that is the intended behavior. If there is any deviation in debugfs we need it to fail so that we can validate that things still work as we expect. I suspect this will result in mild annoyances in the future, but it's a small price to pay given the risk of debugfs doing something unexpected.

sam-berning (approved):

PR #690, opened 2026-07-27
fix(buildsys): include EIF artifacts in copy_guest_image_artifacts

ginglis13:
**Issue number:**

n/a

**Description of changes:**

rpm2eif produces three artifacts that use naming conventions
incompatible with the existing dot-suffix allowlist: {prefix}.eif,
{prefix}-disk.img, and {prefix}-kernel.

Without this fix, `copy_guest_image_artifacts` silently drops all
EIF outputs when embedding a guest variant into a host image, because
the find glob `*.${suffix}` never matches the hyphen-separated
names.

Add `eif` to IMAGE_ARTIFACT_SUFFIXES and introduce a new
IMAGE_ARTIFACT_GLOBS array for patterns that cannot be expressed as
dot-suffixes. Update guest-images-helper to iterate both arrays when
building the find command. Update shell and Rust integration tests to
cover the new artifact types.

**Testing done:**

Built a variant with these changes, `image-format = "eif"`, `standalone-image = true`, and https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/944

After building the host variant, extracted its EROFS root partition and checked the guest embed paths:

```
$ dump.erofs --ls --path /…/usr/share/bottlerocket/guests/enclave /tmp/host-root.img

NID TYPE  FILENAME
26123648    1  bottlerocket-aws-dev-enclave-x86_64-1.63.0-7e3a523f-disk.img
26140035    1  bottlerocket-aws-dev-enclave-x86_64-1.63.0-7e3a523f-kernel
26146876    1  bottlerocket-aws-dev-enclave-x86_64-1.63.0-7e3a523f.eif
26153717    7  latest-disk.img
26153721    7  latest-kernel
 7463527    7  latest.eif
```

**Terms of contribution:**

### DIFF
--- a/tests/integration-tests/src/guest_images_helper.rs
+++ b/tests/integration-tests/src/guest_images_helper.rs
@@ -14,7 +14,7 @@ fn helper_path() -> PathBuf {
 
 /// Build a synthetic guest image directory mimicking the layout produced by a real variant
 /// build. Includes:
-///   - the bootable image artifacts that should be copied
+///   - the bootable image artifacts that should be copied (standard + EIF)
 ///   - the build-metadata files that must NOT be copied (the bug we're guarding against)
 ///   - a stable-name symlink (`os_image.img.lz4` -> `bottlerocket-…img.lz4`)
 fn populate_synthetic_guest_dir(dir: &Path) {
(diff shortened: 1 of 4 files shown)


jmt-lab (approved):
LGTM

sam-berning (approved):

ginglis13:
Added testing details

PR #694, opened 2026-07-31
build(krane-bundle): use prebuilt krane binary v0.21.8

jmt-lab:
Switch krane-bundle from building krane from Go source to embedding the official prebuilt release binary from
https://github.com/google/go-containerregistry/releases/tag/v0.21.8, pinned by per-platform SHA512.

The build script now selects the appropriate goreleaser archive (Linux_x86_64, Linux_arm64, Darwin_x86_64, Darwin_arm64) based on the Cargo target, fetches it via build-cache-fetch, and extracts the krane binary. Go and patch are no longer build-time dependencies, so this also drops:

- The setup-go step and 'patch' apt install from .github/workflows/rust.yml
- The 'apt install golang-1.22 patch' [workspace.metadata.cross.build] pre-build hook in Cargo.toml
- The 'which' build-dependency and the local ecr-login source patch (the v0.21.8 release already tracks a current ecr-login)

Attribution still pulls the pinned source tarball to vendor Go deps for license scanning. CI continues to enforce cache-only fetches; local developers can set UPSTREAM_SOURCE_FALLBACK=true to bootstrap before the archives are mirrored, but the fallback is intentionally not enabled in CI to preserve the sha-pinned-cache trust model.

**Terms of contribution:**


.github/workflows/rust.yml
```diff
@@ -17,11 +17,9 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
       - run: cargo install cargo-deny --locked
       - run: cargo install cargo-make --locked
-      - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c
-        with:
-          go-version: "^1.18"
-      # Install `patch`, needed to build `krane-bundle`, and tools for linking against musl
-      - run: sudo apt-get install -y patch musl-tools musl-dev
+      # Install tools for linking against musl. `krane-bundle` now bundles a
+      # prebuilt `krane` binary, so `go` and `patch` are no longer required.
```
ginglis13:
```suggestion
      # Install tools for linking against musl.
```

I think the second half of this comment only has relevance to this PR and not to long term maintenance of this

tools/krane/hashes/Darwin_x86_64
```diff
@@ -0,0 +1,2 @@
+# https://github.com/google/go-containerregistry/releases/download/v0.21.8/go-containerregistry_Darwin_x86_64.tar.gz
```
ginglis13: we probably don't need these given afaik we aren't building twoliter for mac

tools/krane/build.rs
```diff
@@ -1,127 +1,103 @@
...
+    // Download and checksum-verify the release archive. `build-cache-fetch`
+    // looks up files by the SHA512 sum in the hash file from Bottlerocket's
+    // build cache. When bumping `CRANE_VERSION`, the new archives must be
+    // uploaded to the cache before the change lands in CI (see README).
+    //
+    // As a local-only escape hatch (never enabled in CI), setting
+    // `UPSTREAM_SOURCE_FALLBACK=true` allows the fetch to fall back to the
+    // GitHub Releases URL encoded as a comment at the top of the hash file.
+    // The archive is still SHA512-verified against the pinned hash regardless
+    // of which source served the bytes.
```
ginglis13: this seems overly verbose and better covered in the README

ginglis13 (approved):

sam-berning (approved):

PR #697, opened 2026-08-05
deprecate(testsys): hard deprecation of testsys

jmt-lab:
Testsys has not been updated or used in this repository for almost a year now, as such we are deprecating the testsys binary included with twoliter. 

**Terms of contribution:**


deny.toml
```diff
@@ -71,13 +71,30 @@ skip = [
...
     { name = "darling_core" },
     { name = "darling_macro" },
-]
-
-skip-tree = [
-    # we are considering removing testsys from twoliter's source tree
-    # ignoring the dependency closure for now to ensure twoliter stays
-    # up-to-date
-    { name = "testsys" },
+    # Older transitive versions pulled in by build/runtime deps (fatfs, rustls'
```
sam-berning, line 74: looks like there are a number of new dependencies that are getting flagged compared to when we moved `testsys` to `skip-tree` [last year](https://github.com/bottlerocket-os/twoliter/commit/a2bc469b188dd6ee19b555f105f9252a09ea6805#diff-1040309c64844eb1b6b63d8fd67938adbf9461f1b3c61f12cf738f064a02d3deR64-L66). we should follow up to get these back up to date. can you create a tracking issue?

sam-berning (approved):
LGTM, apart from the new skips in `deny.toml`. don't want to increase the scope of this PR too much, so approving assuming we follow up on those skips.

deny.toml
jmt-lab: Good callout: https://github.com/bottlerocket-os/twoliter/issues/699

qianxjcraig (approved):

Sparksssj (approved):

PR #701, opened 2026-08-11
Add support for Unified Kernel Images and Ephemeral Encryption Keys

arnaldo2792:
**Description of changes:**

This adds support to build Unified Kernel Images for Bottlerocket. It also includes changes to support ephemeral encryption keys to encrypt the private and data partitions at runtime with LUKS.

Unlike GRUB based images, UKIs must be installed in a VFAT filesystem as systemd-boot lacks ext4 support. The UKIs are installed in the XBOOTLDR partition rather than the EFI partition. This is to be as close as possible to the GRUB based support, where the kernel is shipped as a separate partition.

In-place upgrades are not supported. And there is a pending change to update the re-pack workflow and pcrsys measurements. They will come as separate PRs.

**Testing done:**

- UKIs were build with: https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/523 and https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/1007
- Pending testing on encryption keys, they were tested independently by @piyush-jena , but we need to test both features together.

**Terms of contribution:**


arnaldo2792:
(Forced push simplifies commit history and adds missing changes lost in rebase)

tools/buildsys/src/manifest.rs
```diff
@@ -1201,7 +1228,32 @@ pub fn validate_image_features(
...
+    {
+        return error::IncompatibleImageFeaturesSnafu {
+            context: "`ephemeral-encryption-keys = true`",
+            reason: "ephemeral encryption keys require `encrypted-storage = true`".to_string(),
+        }
+        .fail()?;
+    }
+
     let is_eif = matches!(image_format, Some(ImageFormat::Eif));
+
```
jmt-lab, line 1243: **[P1] UKI + default `partition-plan = "split"` slips through validation, then breaks `fetch-variant`.** _(commenting here; the fix belongs in `resolved_image_layout` around L1195-1197)_

`resolved_image_layout` coerces `partition_plan` to `Unified` only when `is_eif` is true (L1195-1197). `ImageFormat::Uki` with the manifest default (`Split`) validates fine and `validate_image_features` only rejects `Split` under `standalone-image` or EIF.

Concrete failure path: `rpm2img` (~L914) only emits `os_image.img.lz4` for the UKI branch — no `-data.img.lz4`. `pubsys/src/repo/fetch_variant/mod.rs` L163-171 falls into the default arm, sees `Split`, and asks `save_target` for `{prefix}-data.img.lz4`, which does not exist → opaque `tough` error at publish/fetch time.

The doc comment at L1244 asserts "UKI images have no B partition set" but nothing enforces the corresponding layout.

**Suggested fix at L1195-1197:**
```rust
if is_eif || matches!(image_format, Some(ImageFormat::Uki)) {
    resolved.partition_plan = PartitionPlan::Unified;
}
```
Or add UKI to the split-rejection block in `validate_image_features` for a clear build-time error. Please add a unit test analogous to `uki_format_rejects_in_place_updates`.

tools/buildsys/src/manifest.rs
```diff
@@ -1201,7 +1228,32 @@ pub fn validate_image_features(
...
+    if matches!(image_format, Some(ImageFormat::Uki))
+        && features.contains(&ImageFeature::InPlaceUpdates)
+    {
+        return error::IncompatibleImageFeaturesSnafu {
+            context: "`image-format = \"uki\"`",
+            reason: "\n  - `in-place-updates`: in-place updates require two banks of OS \
+                     partitions (A/B); a UKI image has no B partition set"
+                .to_string(),
+        }
+        .fail()?;
```
jmt-lab, line 1254: **[P1] UKI/ephemeral-encryption-keys validators short-circuit — breaks the module's "report all conflicts once" contract.**

Both new checks use `return … .fail()?` on the first hit:
- L1235-1239 (ephemeral-encryption-keys requires encrypted-storage)
- L1248-1254 (UKI + in-place-updates)

But the module contract (documented L1214-1216) is: *"All discovered conflicts are reported in a single error so that users see the complete list in one build cycle."*

Concrete impact: a user with `image-format = "uki"` + `in-place-updates = true` + `standalone-image = true` + `partition-plan = "split"` gets only the first conflict, fixes it, rebuilds (full container), sees the next, repeats — instead of the single unified error that EIF users get today.

**Suggested fix:** push these into `conflict_messages` alongside the existing standalone-image / EIF logic and let the final `if !conflict_messages.is_empty()` block emit the unified error.

tools/buildsys/src/manifest.rs
```diff
@@ -1201,7 +1228,32 @@ pub fn validate_image_features(
...
 ) -> Result<()> {
+    // `ephemeral-encryption-keys` only makes sense on top of encrypted storage.
+    if features.contains(&ImageFeature::EphemeralEncryptionKeys)
+        && !features.contains(&ImageFeature::EncryptedStorage)
+    {
+        return error::IncompatibleImageFeaturesSnafu {
+            context: "`ephemeral-encryption-keys = true`",
+            reason: "ephemeral encryption keys require `encrypted-storage = true`".to_string(),
+        }
+        .fail()?;
```
jmt-lab, line 1239: **[P2] Add unit tests for the `ephemeral-encryption-keys` precondition.**

The new precondition has no unit test, unlike the sibling UKI check (covered at L2132-2158). A future refactor of `validate_image_features` that removes/reorders this guard would pass CI silently.

**Suggested test** (mirrors `uki_format_rejects_in_place_updates`):
```rust
#[test]
fn ephemeral_encryption_keys_requires_encrypted_storage() {
    let layout = ImageLayout::default();
    let features = HashSet::from([ImageFeature::EphemeralEncryptionKeys]);
    let err = validate_image_features(&features, &layout, None)
        .expect_err("must fail without encrypted-storage");
    let msg = format!("{err}");
    assert!(msg.contains("ephemeral-encryption-keys"));
    assert!(msg.contains("encrypted-storage"));
}
```
Plus a positive test with both features present.

tools/pubsys/src/repo/fetch_variant/mod.rs
```diff
@@ -142,7 +142,7 @@ async fn fetch_variant(
         .context(error::MissingImageLayoutSnafu { variant })?;
     let image_format = manifest_info.image_format();
     let image_ext = match image_format {
-        Some(ImageFormat::Raw) | None => "img.lz4",
+        Some(ImageFormat::Raw) | Some(ImageFormat::Uki) | None => "img.lz4",
```
jmt-lab, line 145: **[P3] Missing `#[snafu(display(…))]` on the new `ManifestParse` variant (~L261-263).** _(commenting here since the file only has one diff hunk)_

Every other variant in the error enum has an explicit `display` attribute; this one falls back to Snafu's default and prints inconsistently vs. siblings like `MissingImageLayout` / `SaveTarget`.

```rust
#[snafu(display("Failed to parse variant manifest: {}", source))]
ManifestParse {
    source: buildsys::manifest::Error,
},
```

If you intend for the message to bubble up set it to `#[snafu(transparent)]`

twoliter/embedded/rpm2img
```diff
@@ -532,66 +630,122 @@ fi
...
+  fi
+  stub_efi="${stub_files[0]}"
+
+  # Get kernel version from /lib/modules
+  kver_dir=$(find "${ROOT_MOUNT}/lib/modules" -mindepth 1 -maxdepth 1 -type d -printf '%P
+' | head -1)
+  if [[ -z "${kver_dir}" ]]; then
+    echo "kernel version directory not found" >&2
+    exit 1
+  fi
```
jmt-lab, line 658: **[P2] Dead code: `kver_dir` is computed but never used.**

`kver_dir` is populated (L652-654) and validated non-empty (L655-658) but never referenced afterward. The `ukify build` call on L677 uses `${BOOT_MOUNT}/vmlinuz` directly.

Either drop the block, or (more useful) pass it into ukify via `--uname "${kver_dir}"` so the UKI's `.uname` PE section reflects the actual kernel version — otherwise ukify falls back to the value in `vmlinuz`'s `.notes` section, which can be missing on some SDK builds and leaves `bootctl list` reporting `"unknown"`.

twoliter/embedded/rpm2img
```diff
@@ -532,66 +630,122 @@ fi
...
+  # be part of the command line itself (this mirrors grub.cfg).
+  # shellcheck disable=SC2016
+  uki_cmdline="${KERNEL_PARAMETERS} ${uki_bootconfig} root=/dev/dm-0 rootwait ro raid=noautodetect random.trust_cpu=on selinux=1 enforcing=1 dm-mod.create=\"${DM_VERITY_ROOT[*]}\" -- systemd.log_target=journal-or-kmsg systemd.log_color=0 systemd.show_status=true"
+
+  # Create EFI/Linux directory for UKI
+  mkdir -p "${BOOT_MOUNT}/EFI/Linux"
+
+  # Build UKI with ukify
+  # Note: bootconfig initrd is handled at runtime via PRIVATE partition
+  ukify build     --stub "${stub_efi}"     --linux "${BOOT_MOUNT}/vmlinuz"     --cmdline "${uki_cmdline}"     --os-release @"${ROOT_MOUNT}/${SYS_ROOT}/usr/lib/os-release"     --output "${BOOT_MOUNT}/EFI/Linux/bottlerocket.efi"
```
jmt-lab, line 677: **[P2] `ukify build` should be split across lines and pass `--uname`.**

The invocation is a single very long line with multiple spaces, obscuring the arguments. Also missing `--uname "${kver_dir}"` (see adjacent comment on the dead `kver_dir` block).

```bash
ukify build \
  --stub "${stub_efi}" \
  --linux "${BOOT_MOUNT}/vmlinuz" \
  --uname "${kver_dir}" \
  --cmdline "${uki_cmdline}" \
  --os-release "@${ROOT_MOUNT}/${SYS_ROOT}/usr/lib/os-release" \
  --output "${BOOT_MOUNT}/EFI/Linux/bottlerocket.efi"
```

twoliter/embedded/rpm2img
```diff
@@ -615,27 +769,60 @@ if [[ ! -s "${BOOTCONFIG_INPUT}" ]]; then
...
+  mkfs.vfat -I -S 512 "${BOOT_IMAGE}" $((partsize["BOOT-A"] * 1024))
+  # Create EFI/Linux directory and copy UKI
+  mmd -i "${BOOT_IMAGE}" ::/EFI
+  mmd -i "${BOOT_IMAGE}" ::/EFI/Linux
+  mcopy -i "${BOOT_IMAGE}" "${BOOT_MOUNT}/EFI/Linux/bottlerocket.efi" ::/EFI/Linux/
+  # Copy HMAC file if present
+  if [[ -f "${BOOT_MOUNT}/.vmlinuz.hmac" ]]; then
+    mcopy -i "${BOOT_IMAGE}" "${BOOT_MOUNT}/.vmlinuz.hmac" ::/
+  fi
+  dd if="${BOOT_IMAGE}" of="${OS_IMAGE}" conv=notrunc bs=1M seek="${partoff["BOOT-A"]}"
```
jmt-lab, line 785: **[P1] UKI BOOT-A has no capacity check — `mcopy` fails with a cryptic error when the UKI overflows the FAT partition.**

The UKI (systemd-stub + vmlinuz + cmdline + os-release) can be >20 MiB on kernel 6.x, especially FIPS / debug builds. `partsize["BOOT-A"]` scales via `BOOT_SCALE_FACTOR=20` in `partyplanner` (≈20 MiB per GiB of OS image, doubled off IPU). For an `os-image-size-gib = 1` UKI variant, BOOT-A is only ~40 MiB and can be exceeded by real kernels.

Failure mode today: `mcopy` on L780 errors with "No space left" deep into stage 3.

**Suggested up-front check** (mirrors `check_image_size`):
```bash
uki_size=$(stat -c %s "${BOOT_MOUNT}/EFI/Linux/bottlerocket.efi")
boot_bytes=$(( partsize["BOOT-A"] * 1024 * 1024 ))
if (( uki_size > boot_bytes - 65536 )); then
  echo "UKI (${uki_size} B) does not fit in BOOT-A (${boot_bytes} B); "\
       "increase os-image-size-gib or BOOT_SCALE_FACTOR" >&2
  exit 1
fi
```
Consider raising `BOOT_SCALE_FACTOR` on the UKI path and/or documenting the minimum image size for UKI variants.

twoliter/embedded/rpm2img
```diff
@@ -615,27 +769,60 @@ if [[ ! -s "${BOOTCONFIG_INPUT}" ]]; then
...
 
 # BOTTLEROCKET-PRIVATE
 if [[ "${STANDALONE_IMAGE}" == "no" ]]; then
-  # Generate bootconfig file for the image.
-  touch "${PRIVATE_MOUNT}/bootconfig.data"
-  bootconfig -a "${BOOTCONFIG_INPUT}" "${PRIVATE_MOUNT}/bootconfig.data"
+  # Generate bootconfig file for the image. When ephemeral encryption keys are
+  # enabled, bootconfig.data was baked onto the BOOT partition above (PRIVATE is
+  # LUKS-encrypted and grub-unreadable), so skip seeding it here.
+  if [[ "${EPHEMERAL_ENCRYPTION_KEYS}" != "yes" ]]; then
```
jmt-lab, line 822: **[P0] `EPHEMERAL_ENCRYPTION_KEYS=yes` + non-UKI (GRUB) is almost certainly unbootable — please confirm the intended combination matrix.**

L821-822 skips seeding `bootconfig.data` on PRIVATE when `EPHEMERAL_ENCRYPTION_KEYS=yes` (comment: *"PRIVATE is LUKS-encrypted and grub-unreadable"*).

But in the non-UKI branch above (L692-702), the guard for `INITRD="initrd (\$private)/bootconfig.data"` is `STANDALONE_IMAGE == "no"` only — it does **not** exclude `EPHEMERAL_ENCRYPTION_KEYS=yes`. So a GRUB-based image with ephemeral keys enabled emits a `grub.cfg` that tells GRUB to read `bootconfig.data` from a LUKS-encrypted PRIVATE partition that GRUB cannot decrypt — unbootable.

Two possible resolutions (please pick one and enforce it):

1. If ephemeral-encryption-keys is **UKI-only**, reject the GRUB+ephemeral combo up-front in `sanity_checks` / `validate_image_features`, matching how UKI+IPU is rejected.
2. If both bootloaders are supposed to support it, extend the L696 guard to also disable the `initrd (\$private)/bootconfig.data` line when ephemeral keys are enabled and put `bootconfig.data` somewhere GRUB can read (or generate it at runtime from initramfs).

twoliter/embedded/imghelper
```diff
@@ -493,8 +527,8 @@ generate_verity_root() {
     "${verity_data_512b_blocks}"
     "verity"
     "${VERITY_VERSION}"
-    "PARTUUID=\$boot_uuid/PARTNROFF=1"
-    "PARTUUID=\$boot_uuid/PARTNROFF=2"
+    "PARTUUID=${boot_partuuid}/PARTNROFF=1"
```
jmt-lab, line 530: **[P2] UKI dm-verity uses `PARTNROFF=1/2` — hardcodes BOOT-A / ROOT-A / HASH-A partition adjacency.**

The `PARTUUID=${boot_partuuid}/PARTNROFF=1` (ROOT-A) and `.../PARTNROFF=2` (HASH-A) references on L530-531 rely on the current on-disk order (BIOS, EFI-A, BOOT-A, ROOT-A, HASH-A). This is fine today, but `sgdisk --sort` (rpm2img:205) reorders by start sector, so any future layout change that inserts a partition between BOOT-A and ROOT-A (or between ROOT-A and HASH-A) — e.g. a LUKS keyslot partition as this feature series grows — silently breaks the UKI verity table. The failure appears at boot as a dm-verity setup error.

**Robust fix:** resolve the real PARTUUIDs of ROOT-A and HASH-A via `get_partition_uuid` and pass them into `generate_verity_root` directly, replacing the offset arithmetic entirely.

twoliter/embedded/imghelper
```diff
@@ -457,13 +464,40 @@ get_erofs_usage() {
...
   local -n dm_verity_root
   root_image="${1:?}"
   verity_image="${2:?}"
   veritypart_mib="${3:?}"
   dm_verity_root="${4:?}"
+  # Optional: the real PARTUUID of the BOOT partition. When omitted we emit the
+  # literal `$boot_uuid` placeholder, which GRUB expands at boot time via
+  # `gptprio.next`. Callers that build images without GRUB (e.g. UKI) must pass
+  # the real UUID, since nothing will expand the placeholder for them.
+  boot_partuuid="${5:-\$boot_uuid}"
```
jmt-lab, line 500: **[P2] `generate_verity_root`: validate `boot_partuuid` format when not the GRUB placeholder.**

The default `${5:-\$boot_uuid}` correctly preserves the literal `$boot_uuid` for the GRUB path. But when a real UUID is passed (the new UKI code path), the value is embedded unchecked into `PARTUUID=…/PARTNROFF=…`. A malformed UUID (whitespace, `/`, empty) silently produces a corrupt verity table.

**Suggested defense-in-depth:**
```bash
if [[ "${boot_partuuid}" != '$boot_uuid' && \
      ! "${boot_partuuid}" =~ ^[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}$ ]]; then
  echo "generate_verity_root: invalid boot_partuuid '${boot_partuuid}'" >&2
  return 1
fi
```

twoliter/embedded/partyplanner
```diff
@@ -554,9 +557,13 @@ set_partition_types() {
   local typecode
   for part in BOOT ROOT HASH RESERVED; do
     for bank in A B; do
-      typecode="BOTTLEROCKET_${part}_TYPECODE"
-      typecode="${!typecode}"
-      pp_type["${part}-${bank}"]="${typecode}"
+      if [[ "${part}" == "BOOT" && "${uki_image}" == "yes" ]]; then
```
jmt-lab, line 560: **[P2] `set_partition_uuids` (L573-594) leaks `partition_plan` into the caller's scope — missing `local`.** _(commenting near the UKI typecode change since L579 is not in the diff)_

```bash
set_partition_uuids() {
  local -n pp_uuid
  pp_uuid="${1:?}"
  partition_plan="${2:?}"   # <-- L579, missing `local`
```
The sibling `set_partition_sizes` correctly declares `local ... partition_plan` (L235). Copy-paste drift.

Today the only callers use uppercase `PARTITION_PLAN`, so no collision, but any future caller (e.g. a helper inside img2img) with its own lowercase `local partition_plan` will find its value silently clobbered by this function via bash's dynamic scoping.

**Fix:** `local partition_plan="${2:?}"` on L579.

jmt-lab (changes requested):
If some of these are inaccurate feel free to comment with explanation. I think if they aren't accurate it will help me get through my second pass if i know why since this is a rather large change.

tools/buildsys/src/builder.rs
```diff
@@ -599,11 +600,21 @@ impl DockerBuild {
...
+            // `img2img` has no EIF support. Fail fast here so users get a clear
+            // error instead of a cryptic failure deep inside the Dockerfile
+            // build.
+            Some(ImageFormat::Eif) => return error::EifRepackUnsupportedSnafu.fail(),
+            // UKI repack is supported. `img2img` extracts the FAT boot
+            // partition with mtools, rebuilds `EFI/Linux/bottlerocket.efi` from
+            // the sections of the existing UKI so that the kernel command line
+            // carries the new dm-verity root hash, and recreates the partition.
+            // See `repack_uki` in `twoliter/embedded/img2img`.
+            Some(ImageFormat::Uki) => {}
```
sam-berning, line 616: nit: why is this on its own line? can be lumped in with the other supported formats

twoliter/embedded/build.Dockerfile
```diff
@@ -606,6 +614,7 @@ RUN --mount=target=/host \
       ${EROFS_ROOT_PARTITION:+--with-erofs-root-partition=yes} \
       ${UEFI_SECURE_BOOT:+--with-uefi-secure-boot=yes} \
       ${IN_PLACE_UPDATES:+--with-in-place-updates=yes} \
+      $([ "${IMAGE_FORMAT}" = "uki" ] && echo "--with-uki-image=yes") \
```
sam-berning, line 617: this looks like it's ignored in `img2img`, and `--output-fmt "uki"` can be used instead. I don't see a reason to break with the standard `image-format` argument style here.

twoliter/embedded/rpm2eif
```diff
@@ -42,6 +42,7 @@ Optional:
     --with-uefi-secure-boot=yes      Rejected: EIF does not sign shim/grub.
     --with-in-place-updates=yes      Rejected: EIF is single-bank.
     --with-encrypted-storage=yes     Rejected: EIF has no PRIVATE/DATA.
```
sam-berning, line 44: nit: please match the indentation of the new `--with-ephemeral-encryption-keys=yes` argument

twoliter/embedded/rpm2img
```diff
@@ -532,66 +630,122 @@ fi
...
-       ${BOOTCONFIG} \\
-       root=/dev/dm-0 rootwait ro \\
-       raid=noautodetect \\
-       random.trust_cpu=on \\
-       selinux=1 enforcing=1 \\
-       dm-mod.create="\$dm_verity_root" \\
-       -- \\
-       systemd.log_target=journal-or-kmsg \\
-       systemd.log_color=0 \\
+   linux (\$root)/vmlinuz \
```
sam-berning, line 728: this change seems strange. since IIUC this only affects grub, I'd prefer not to touch this unless it fixes some known bug

tools/buildsys/src/manifest.rs
```diff
@@ -975,12 +975,35 @@ pub struct BuildVariant {
...
+/// separation is preserved at runtime, where the built image carries
+/// `/usr/share/bottlerocket/image-format.env` alongside (and separate from)
+/// `/usr/share/bottlerocket/image-features.env`. Like `image-features.env`,
+/// `image-format.env` is written unconditionally for every image format, so a
+/// consumer never has to treat file absence as meaningful:
+///
+/// | Key            | Values                                | Notes                                    |
+/// |----------------|---------------------------------------|------------------------------------------|
+/// | `IMAGE_FORMAT` | `raw`, `qcow2`, `vmdk`, `eif`, `uki`  | The lowercase manifest value.            |
+/// | `UKI_IMAGE`    | `true` / `false`                      | Convenience boolean for `IMAGE_FORMAT=uki`. |
```
sam-berning, line 991: is this variable with the `true` or `false` format actually used anywhere?

arnaldo2792: It is used in the core kit, to prevent the execution of systemd services that should run for other variants.

tools/pubsys/src/repo/fetch_variant/mod.rs
arnaldo2792: This is not a change introduced by this PR.

twoliter/embedded/build.Dockerfile
arnaldo2792: It won't be in a change to support re-pack!

twoliter/embedded/imghelper
arnaldo2792: Isn't this true for `in-place-updates = false`? This seems like a larger change on how we arrange part sizes/layouts and we could enforce that in a separate PR.

twoliter/embedded/partyplanner
arnaldo2792: As far as I can tell, this is not a problem introduced by this PR.  We should have an issue to fix this.

twoliter/embedded/rpm2img
arnaldo2792: This is a great catch!. If we are missing the .notes section in some builds, I would rather fix that.

twoliter/embedded/rpm2img
arnaldo2792: Yup, the AI was fixing shellchecks, I'll drop it.

piyush-jena:
The force push above addresses missing unit tests for ephemeral-encryption-keys and a formatting issue.

tools/buildsys/src/manifest.rs
```diff
@@ -1204,7 +1228,32 @@ pub fn validate_image_features(
     layout: &ImageLayout,
     image_format: Option<&ImageFormat>,
 ) -> Result<()> {
+    // `ephemeral-encryption-keys` only makes sense on top of encrypted storage.
+    if features.contains(&ImageFeature::EphemeralEncryptionKeys)
+        && !features.contains(&ImageFeature::EncryptedStorage)
```
piyush-jena: Can you please add an additional condition here where Ephemeral encryption keys work only with UKI?

twoliter/embedded/imghelper
jmt-lab: Can you file an issue for us to keep track of this?

jmt-lab: nvm i got it

tools/buildsys/src/manifest.rs
arnaldo2792: The missing B partition doesn't inherently mean one single disk must be produced. Having rootfs + data as two separate images is a valid use case.

jmt-lab (approved):

arnaldo2792:
(Forced push includes a change to generate the correct hmac for the UKIs)

arnaldo2792:
(Forced push to fix conflict)

tools/buildsys/src/manifest.rs
```diff
@@ -1148,11 +1173,12 @@ const DEPRECATED_IMAGE_FEATURES: &[&ImageFeature] = &[
...
 /// - If the image format is `eif`, the partition plan is forced to
 ///   `unified`. EIF images are inherently single-disk (no separate DATA
-///   image), and `rpm2eif` ignores the partition-plan flag anyway; forcing
+///   image), and `rpm2eif` ignores the partition-plan flag anyway. Forcing
 ///   `unified` here means variants don't have to write out
 ///   `partition-plan = "unified"` boilerplate purely to satisfy the
 ///   validator, and the historical default (`split`) doesn't turn into an
-///   error the moment the user opts into EIF.
+///   error the moment the user
+///   opts into EIF.
```
sam-berning: nit: this comment should remain on one line

twoliter/embedded/imghelper
```diff
@@ -692,6 +732,34 @@ provide_certs() {
   mcopy -i "${efi_image}" -ov "${SBKEYS}"/db.{crt,cer} ::/EFI/BOOT
 }
 
+# Copy the systemd-boot configuration onto the EFI system partition.
+#
+# systemd-boot reads exactly one configuration file, `\loader\loader.conf`, and
```
sam-berning: nit: 
```suggestion
# systemd-boot reads exactly one configuration file, `/loader/loader.conf`, and
```

arnaldo2792:
Forced push includes:

- Address feedback on menuentry generation shellcheck fixes
- re-format comment
- Use `/` instead of `\` in comment

sam-berning (approved):

PR #705, opened 2026-08-12
feat(eif): add [eif] signing config schema in pubsys-config

jmt-lab:
Introduce a dedicated [eif] section in Infra.toml carrying EIF signing
config, decoupled from `signing_keys` (TUF metadata) and `sbkeys` (UEFI
Secure Boot). Two backends: `file` (local PEM EC key) and `kms` (AWS
KMS key id/ARN/alias with optional region).

Adds `EifConfig` and `EifSigningKey` types to pubsys-config with three
serde round-trip unit tests. Documents the section in Infra.toml.example
alongside the existing signing profiles.

No consumers yet — this is the schema foundation. Buildsys plumbing and
the eif-builder CLI land in follow-up commits in this stack.

### DIFF
--- a/tools/pubsys-config/src/lib.rs
+++ b/tools/pubsys-config/src/lib.rs
@@ -29,6 +29,9 @@ pub struct InfraConfig {
 
     // Config for container registries
     pub vendor: Option<HashMap<String, Vendor>>,
+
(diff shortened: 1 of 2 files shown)


jmt-lab:

PR #707, opened 2026-08-12
feat(eif-builder): add signing flags, resign, and describe subcommands

jmt-lab:
Extend the eif-builder CLI to consume the new signing library:

* Top-level build gains --signing-cert + (--signing-key | --kms-key-id)
  + optional --region, gated by a clap ArgGroup. Passing --signing-cert
  alone is an error. Omitting all three produces an unsigned EIF, so
  existing callers keep working.
* New `resign` subcommand: rewrites only the SIGNATURE section (and
  header CRC) of an existing EIF, byte-preserving kernel/cmdline/
  ramdisk. PCR0 is recomputed from the input bytes so the new signature
  covers the same measurement the enclave sees.
* New `describe` subcommand: prints a JSON summary (including the
  decoded METADATA section) to stdout so `eif2eif` can carry
  KernelVersion / BuildToolVersion forward across a repack.

The implicit top-level "build" flow is preserved for backward
compatibility with rpm2eif/eif2eif.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -46,9 +46,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.4"
+version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


jmt-lab:

PR #709, opened 2026-08-12
feat(eif): add in-tree EIF repack path (eif2eif)

jmt-lab:
Add `eif2eif`, a new Dockerfile-invoked tool that repackages an existing
EIF: rewrites the CA bundle / root.json inside the EIF's rootfs, rebuilds
erofs + dm-verity, pins the cmdline to the new verity root hash, and
produces a freshly signed .eif using the [eif] signing profile from the
prior commit.

Wires the imgrepack Dockerfile stage to dispatch to `eif2eif` when
IMAGE_FORMAT=eif and `img2img` otherwise. Plumbs KERNEL_PARAMETERS and
TWOLITER_VERSION as build args so the rebuilt EIF's cmdline and
BuildToolVersion metadata match what rpm2eif emitted at first build.

Removes the `EifRepackUnsupported` early-fail in `buildsys::repack_variant`
now that a real repack path exists.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -46,9 +46,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.4"
+version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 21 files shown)


jmt-lab:

PR #710, opened 2026-08-12
feat(eif): resign guest EIFs during host image repack

jmt-lab:
Extend img2img so that when a host variant embeds guest EIFs (declared
under `[package.metadata.build-variant.guest-images]`), the repack walks
each guest's install path and re-signs every `*.eif` in place via
`eif-builder resign`. The kernel, cmdline, ramdisk, and metadata sections
are byte-preserved; only the SIGNATURE section (and header CRC) are
rewritten. Uses the same [eif] signing profile as the host build.

Adds an end-to-end integration test that builds a host variant embedding
a guest EIF and verifies every guest EIF under the resulting output has
been resigned with the expected certificate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2622,6 +2622,7 @@ dependencies = [
  "tokio",
  "toml",
  "twoliter",
+ "walkdir",
  "which",
 ]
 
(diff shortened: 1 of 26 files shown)


jmt-lab:

PR #711, opened 2026-08-12
feat(eif): embed os-release into rpm2eif rootfs

jmt-lab:
Mirror the os-release append that rpm2img performs, so EIF-built rootfs
images expose the same identifying fields (VERSION, PRETTY_NAME,
VARIANT_ID, VERSION_ID, BUILD_ID, VENDOR_NAME, and the canonical URLs)
to userspace. Unlike rpm2img, INSTALL_ROOT already points at the merged
sysroot, so the append writes directly to ${INSTALL_ROOT}/usr/lib/os-release
without the ${ARCH}-bottlerocket-linux-gnu/sys-root prefix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2622,6 +2622,7 @@ dependencies = [
  "tokio",
  "toml",
  "twoliter",
+ "walkdir",
  "which",
 ]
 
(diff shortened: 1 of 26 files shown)


jmt-lab:

PR #712, opened 2026-08-12
feat(eif): expose EIF header PCIE flags as variant metadata

jmt-lab:
Add `[package.metadata.build-variant] eif-pcie-flags` so `image-format =
"eif"` variants can author the PCIE flag word written into the EIF
header. The value flows buildsys -> Docker ARG (`EIF_PCIE_FLAGS`) ->
rpm2eif / eif2eif -> `eif-builder --pcie-flags <hex>`; when unset,
`eif-builder`'s built-in `DEFAULT_PCIE_FLAGS` (0x240) is applied, so
existing variants are unaffected.

The header value must match the PCIE flags the sidecar shim passes to
the hypervisor at launch (`header == launch-flags`, hypervisor-enforced);
today the shim's launch flags are authored independently, so a
per-variant knob is what lets a build keep the two sides in sync. Once
the shim derives its launch flags from the header directly (host echoes,
hypervisor enforces), the coupling goes away — but this knob is still
the authoring surface for the header value itself.

Accepted forms: TOML integer literal (decimal, per TOML spec, e.g. `832`)
or a `0x`-prefixed hex string (e.g. `"0x340"`). Unprefixed string forms
like `"340"` are rejected at parse time: eif-builder's --pcie-flags CLI
treats `340` as hex, so silently reading the unprefixed string as
decimal would produce exactly the header/launch-flags mismatch this
knob exists to prevent.

The header==launch-flags contract is documented on
`DEFAULT_PCIE_FLAGS` in eif-builder, on the new `BuildVariant`
field, and briefly in the eif-builder `--pcie-flags` CLI help.

Tests: 6 new unit tests in tools/buildsys covering omitted field,
integer literal, hex string (`0x` and `0X`), unprefixed-string
rejection, and u16 overflow.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2622,6 +2622,7 @@ dependencies = [
  "tokio",
  "toml",
  "twoliter",
+ "walkdir",
  "which",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -46,9 +46,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.4"
+version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
+checksum = "c982642fa9e8606056828ee9a8505737230110bb1099153c79efe865c59d12ba"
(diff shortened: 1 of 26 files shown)


jmt-lab:

PR #713, opened 2026-08-12
feat(eif): select x86_64 EIF kernel format via variant metadata

jmt-lab:
The sidecar Nitro Enclaves in-memory image loader validates
`BZIMAGE_HEADER_MAGIC` in the EIF kernel section and boots via the
bzImage protocol (same as every `nitro-cli`-built EIF). rpm2eif was
embedding an uncompressed ELF `vmlinux` "for PVH boot" -- correct only
for a bare-metal Firecracker PVH loader, wrong for the sidecar path.
The enclave started, the guest never booted, zero console output, and
the failed boot wedged the host.

Fix by making the format selectable and defaulting to bzImage:

* Add `[package.metadata.build-variant] eif-kernel-format` accepting
  `"bzimage"` (default) or `"vmlinux"`. Ignored on aarch64 (only
  PE-wrapped `Image` is valid) and by non-EIF variants.
* Plumb via buildsys -> Docker ARG (`EIF_KERNEL_FORMAT`) -> rpm2eif
  `--eif-kernel-format=<fmt>`. Empty ARG means "apply rpm2eif's default"
  (bzImage), so existing variants pick up the fix automatically.
* rpm2eif x86_64 branch switches on the flag: `bzimage` uses the
  RPM-shipped `vmlinuz` verbatim (no ELF extraction); `vmlinux` retains
  the existing prefer-bare-then-carve-ELF-out-of-vmlinuz path for
  bare-metal PVH consumers.
* eif-builder x86_64 `prepare_kernel` is already a byte pass-through;
  both formats round-trip cleanly. Only doc comments needed updating.

Kept selectable rather than hardcoded because Firecracker upstream has
shipped conflicting guidance on compressed vs. uncompressed kernels
more than once, and forking rpm2eif to swap formats is worse than a
one-line variant knob. The `extract_vmlinux` machinery is reachable only
via `--eif-kernel-format=vmlinux` but is not dead code.

Tests: 4 new buildsys unit tests (omitted/bzimage/vmlinux/unknown),
3 new rpm2eif bash tests (bzimage/vmlinux accepted, unknown rejected).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2622,6 +2622,7 @@ dependencies = [
  "tokio",
  "toml",
  "twoliter",
+ "walkdir",
  "which",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -46,9 +46,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.4"
+version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
+checksum = "c982642fa9e8606056828ee9a8505737230110bb1099153c79efe865c59d12ba"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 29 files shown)


jmt-lab:

PR #714, opened 2026-08-12
feat(eif): add [eif] signing config schema in pubsys-config

jmt-lab:
Introduce a dedicated [eif] section in Infra.toml carrying EIF signing
config, decoupled from `signing_keys` (TUF metadata) and `sbkeys` (UEFI
Secure Boot). Two backends: `file` (local PEM EC key) and `kms` (AWS
KMS key id/ARN/alias with optional region).

Adds `EifConfig` and `EifSigningKey` types to pubsys-config with three
serde round-trip unit tests. Documents the section in Infra.toml.example
alongside the existing signing profiles.

No consumers yet — this is the schema foundation. Buildsys plumbing and
the eif-builder CLI land in follow-up commits in this stack.

### DIFF
--- a/tools/pubsys-config/src/lib.rs
+++ b/tools/pubsys-config/src/lib.rs
@@ -29,6 +29,9 @@ pub struct InfraConfig {
 
     // Config for container registries
     pub vendor: Option<HashMap<String, Vendor>>,
+
+    // EIF (Nitro Enclave Image Format) signing config
+    pub eif: Option<EifConfig>,
 }
 
 impl InfraConfig {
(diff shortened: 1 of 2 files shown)


ginglis13 (approved):

Sparksssj (approved):

tools/pubsys-config/src/lib.rs
```diff
@@ -212,6 +215,47 @@ impl TryFrom<SigningKeyConfig> for Url {
...
+    pub signing_key: EifSigningKey,
+}
+
+/// Backend for the EIF signing key. Only `file` and `kms` are supported;
+/// these map to `eif-builder`'s `--signing-key` and `--kms-key-id` flags.
+#[allow(non_camel_case_types)]
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+pub enum EifSigningKey {
+    file {
```
sam-berning, line 238: nit: I'd prefer we stick with Pascal case enum variant names, and `serde(rename)` to the lower-case version. It's unintuitive to me when reading this code that `file` and `kms` are enums

jmt-lab: So this is consistent with SigningKey above, however I think in the future we should probably clean up this crate with proper serde rename

sam-berning: Ok, I won't block on this then.

sam-berning (approved):

PR #715, opened 2026-08-12
feat(eif-builder): add PCR0-covering signature section

jmt-lab:
Extend eif-builder library with EifSectionSignature (0x04) covering PCR0
via COSE_Sign1 ECDSA (ES256/ES384). Introduces a Signer trait with two
implementations:

* LocalSigner (aws-lc-rs) — reads a PEM ECDSA key from disk.
* KmsSigner (aws-sdk-kms) — calls KMS Sign on P-256/P-384 keys with
  typed errors for KmsAccessDenied / KmsKeyDisabled / KmsInvalidKeyState.

Wire format matches aws-nitro-enclaves-image-format: CBOR Array(1) of
Map { signing_certificate: PEM, signature: COSE_Sign1 } with a PcrInfo
payload for PCR0 = SHA-384(48 zero bytes || SHA-384(kernel || cmdline
|| ramdisks)). This makes `nitro-cli describe-eif` populate PCR8.

Also adds byte-preserving resign_eif() and describe_eif() library
entry points for the CLI subcommands landing in the next commit.

New deps: aws-lc-rs, coset, ciborium, pem, x509-cert.

Tests: PCR0 spec conformance, COSE payload shape, end-to-end signature
verification, DER parser edge cases (SEQUENCE length + trailing byte
rejection), resign round-trips.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,6 +103,7 @@ async-trait = "0.1"
 async-walkdir = "1"
 authenticode = "0.5"
 aws-config = { version = "1", default-features = false, features = ["credentials-process", "default-https-client", "rt-tokio" ] }
+aws-lc-rs = { version = "1.17", default-features = false, features = ["aws-lc-sys"] }
 aws-credential-types = "1"
 aws-sdk-ebs = { version = "1", default-features = false, features = ["default-https-client", "rt-tokio"] }
 aws-sdk-ec2 = { version = "1", default-features = false, features = ["default-https-client", "rt-tokio"] }
@@ -119,6 +120,7 @@ bounded-integer = "0.5"
 bytes = "1"
 clap = "4"
 coldsnap = { version = "0.11", default-features = false }
+coset = { version = "0.3", features = ["std"] }
 crc32fast = "1"
 daemonize = "0.5"
 darling = "0.23"
@@ -145,6 +147,7 @@ libc = "0.2"
 log = "0.4"
 lzma-rs = "0.3"
 maplit = "1"
+minicbor-serde = { version = "0.7", features = ["std"] }
 mockall = "0.13"
 nix = "0.31"
 nonzero_ext = "0.3"
@@ -153,6 +156,7 @@ num-traits = "0.2"
 object = "0.37"
 once_cell = "1.21"
 olpc-cjson = "0.1"
+pem = "3"
 pest = "2.8"
 pest_derive = "2.8"
 proc-macro2 = "1"
@@ -192,6 +196,7 @@ url = "2"
 uuid = "1"
 walkdir = "2"
 which = "8"
+x509-cert = { version = "0.2", default-features = false, features = ["std", "pem"] }
 zstd = "0.13"
 
 [profile.dev.package]
--- a/deny.toml
+++ b/deny.toml
@@ -10,6 +10,7 @@ allow = [
     "BSD-2-Clause",
     "BSD-3-Clause",
     "BSL-1.0",
+    "BlueOak-1.0.0",
     # "CC0-1.0",
     "ISC",
     "MIT",
@@ -60,8 +61,6 @@ multiple-versions = "deny"
 wildcards = "deny"
 
 skip = [
-    # several dependencies are using multiple versions of base64
-    { name = "base64" },
     # governor and guppy collide bringing two versions of hashbrown and foldhash
     { name = "hashbrown" },
     { name = "foldhash" },
--- a/tools/eif-builder/src/lib.rs
+++ b/tools/eif-builder/src/lib.rs
@@ -7,23 +7,42 @@
 //! All multi-byte fields are big-endian.
 
 pub mod kernel;
+pub mod signer;
 
 pub use kernel::KernelPrepError;
+pub use signer::{Signer, SignerError};
 
+use std::borrow::Cow;
 use std::fs;
 use std::io::Write;
 use std::path::{Path, PathBuf};
 
 use crc32fast::Hasher as Crc32Hasher;
+use serde::{Deserialize, Serialize};
 use snafu::{ensure, ResultExt, Snafu};
 
 const EIF_MAGIC: [u8; 4] = [0x2e, 0x65, 0x69, 0x66]; // ".eif"
+/// EIF header version. v4 adds the METADATA section (0x05) and retains the
+/// v3 SIGNATURE section (0x04).
 const EIF_HDR_VERSION: u16 = 4;
 
 const EIF_SECTION_KERNEL: u16 = 1;
 const EIF_SECTION_CMDLINE: u16 = 2;
 const EIF_SECTION_RAMDISK: u16 = 3;
+const EIF_SECTION_SIGNATURE: u16 = 4;
 const EIF_SECTION_METADATA: u16 = 5;
+
+/// Human-readable EIF section type name; `"UNKNOWN"` for unregistered types.
+fn section_type_name(ty: u16) -> &'static str {
+    match ty {
+        EIF_SECTION_KERNEL => "KERNEL",
+        EIF_SECTION_CMDLINE => "CMDLINE",
+        EIF_SECTION_RAMDISK => "RAMDISK",
+        EIF_SECTION_SIGNATURE => "SIGNATURE",
+        EIF_SECTION_METADATA => "METADATA",
+        _ => "UNKNOWN",
+    }
+}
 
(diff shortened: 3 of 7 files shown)


Sparksssj (approved):

tools/eif-builder/src/signer.rs
```diff
@@ -0,0 +1,673 @@
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//! EIF signature production for the `EifSectionSignature` (0x04) section.
+//!
+//! [`LocalSigner`] signs in-process from a PEM ECDSA key.
+//! [`KmsSigner`] delegates to AWS KMS. Both emit raw ECDSA `r || s`
+//! (RFC 8152 §8.1); P-256 → ES256, P-384 → ES384; other curves rejected.
```
ginglis13: Can you link to the actual RFC if you're going to mention it? This particular RFC is obsoleted: https://www.rfc-editor.org/info/rfc8152/

tools/eif-builder/src/signer.rs
```diff
@@ -0,0 +1,673 @@
...
+    /// COSE algorithm to advertise.
+    fn algorithm(&self) -> SignAlg;
+}
+
+/// In-process ECDSA signer backed by `aws-lc-rs`.
+pub struct LocalSigner {
+    key_pair: EcdsaKeyPair,
+    cert_pem: Vec<u8>,
+    #[cfg(test)]
+    public_key_spki_der: Vec<u8>,
```
ginglis13: odd to have test only field in the pub struct?

tools/eif-builder/src/signer.rs
```diff
@@ -0,0 +1,673 @@
...
+        if key_curve != cert_alg {
+            return CurveMismatchSnafu {
+                cert_curve: name_of(cert_alg),
+                key_curve: name_of(key_curve),
+            }
+            .fail();
+        }
+
+        #[cfg(test)]
+        let public_key_spki_der = spki_der_for_ecdsa(&raw_pub, key_curve);
```
ginglis13: again why are we embedding test only data on the actual pub struct?

Can we just have a 

```rust
/// SPKI DER for the signing key. Test-only; derived on demand.
#[cfg(test)]
pub fn public_key_der(&self) -> Vec<u8> {
      spki_der_for_ecdsa(self.key_pair.public_key().as_ref(), self.alg)
}
```

for use in lib.rs?

tools/eif-builder/src/signer.rs
```diff
@@ -0,0 +1,673 @@
...
+        if parsed_cert.tag() != "CERTIFICATE" {
+            return CertPemLabelSnafu {
+                label: parsed_cert.tag().to_string(),
+            }
+            .fail();
+        }
+        let alg = detect_cert_curve(parsed_cert.contents())?;
+
+        // Current-thread runtime so sync callers can invoke the async KMS SDK.
+        let runtime = tokio::runtime::Builder::new_current_thread()
```
ginglis13: you could panic if already inside a tokio runtime, via either `block_on` or dropping the runtime. You could gate this with a new snafu and an early check in this method:

```rust
// This type owns a current-thread runtime and drives KMS via block_on.
// Both block_on and dropping the runtime panic if we're already inside a
// runtime, so refuse up front and turn a late panic into a typed error.
if tokio::runtime::Handle::try_current().is_ok() {
    return KmsInAsyncContextSnafu.fail();
}
```

tools/eif-builder/src/signer.rs
jmt-lab: good catch this is really weird.

tools/eif-builder/src/signer.rs
jmt-lab: I'm also questioning why this method isn't async to start with,

deny.toml
```diff
@@ -54,14 +55,24 @@ license-files = [
     { path = "LICENSE", hash = 0x001c7e6c },
 ]
 
+[[licenses.clarify]]
+name = "half:1.8.3"
```
ginglis13: Is this still needed? Now that we've got rid of `serde_cbor`

clarify.toml
```diff
@@ -5,6 +5,14 @@ ignore-licenses = [
     "Pixar"
 ]
 
+[clarify.half]
```
ginglis13: shouldn't be needed anymore

Cargo.toml
```diff
@@ -113,8 +114,10 @@ bit_field = "0.10"
 bon = "3"
 bounded-integer = "0.5"
 bytes = "1"
+ciborium = "0.2"
```
ginglis13: this is now only a transitive dep
```suggestion
```

deny.toml
```diff
@@ -54,14 +55,24 @@ license-files = [
...
+
 [bans]
 # Deny multiple versions or wildcard dependencies.
 multiple-versions = "deny"
 wildcards = "deny"
 
 skip = [
     # several dependencies are using multiple versions of base64
-    { name = "base64" },
+    { name = "half" },
```
ginglis13: skip no longer needed, this is a no-op

tools/eif-builder/src/lib.rs
```diff
@@ -286,26 +320,556 @@ pub fn build_eif(
...
+///   "sections": [
+///     { "type": "KERNEL",   "type_id": 1, "size": 50331648 },
+///     { "type": "CMDLINE",  "type_id": 2, "size":      431 },
+///     { "type": "METADATA", "type_id": 5, "size":      254 },
+///     { "type": "SIGNATURE","type_id": 4, "size":     2115 }
+///   ],
+///   "metadata": { "ImageName": "...", "BuildMetadata": { ... } }
+/// }
+/// ```
+///
```
ginglis13:
```suggestion
/// Summarize an EIF's contents as JSON.
```

tools/eif-builder/src/signer.rs
```diff
@@ -0,0 +1,668 @@
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//! EIF signature production for the `EifSectionSignature` (0x04) section.
+//!
+//! [`LocalSigner`] signs in-process from a PEM ECDSA key.
+//! [`KmsSigner`] delegates to AWS KMS. Both emit the COSE raw ECDSA
+//! encoding `I2OSP(R, n) | I2OSP(S, n)` with `n = ceil(key_length / 8)`
+//! (RFC 9053 §2.1, which along with RFC 9052 obsoletes RFC 8152);
```
ginglis13: Please just link to the RFC, like
```suggestion
//! (RFC 9053 §2.1: https://www.rfc-editor.org/rfc/rfc9053#section-2.1);
```

tools/eif-builder/src/signer.rs
```diff
@@ -0,0 +1,668 @@
...
+//! P-256 → ES256, P-384 → ES384; other curves rejected. The TBS bytes
+//! passed to [`Signer::sign_cose`] are the COSE `Sig_structure1`
+//! serialization defined in RFC 9052 §4.4, built by the caller.
```
ginglis13: Please link to the RFC

tools/eif-builder/src/signer.rs
```diff
@@ -0,0 +1,668 @@
...
+            Self::Es384 => 48,
+        }
+    }
+}
+
+/// Backend for producing the COSE_Sign1 signature over the pre-hash TBS bytes.
+///
+/// `sign_cose` is `async` so KMS-backed implementations can await the SDK
+/// directly instead of blocking on an owned runtime. In-process backends do
+/// only CPU work and complete without yielding.
```
ginglis13:
```suggestion
```

tools/eif-builder/src/lib.rs
```diff
@@ -286,26 +320,539 @@ pub fn build_eif(
...
+            continue; // drop the old signature
+        }
+        entries.push(SectionEntry {
+            ty: s.ty,
+            data: Cow::Borrowed(&s.data),
+        });
+    }
+    // Check MAX_NUM_SECTIONS here so the error attributes overflow to resign.
+    ensure!(
+        entries.len() < MAX_NUM_SECTIONS,
```
sam-berning, line 718: nit: I think this is slightly more readable:


```suggestion
        entries.len() <= MAX_NUM_SECTIONS - 1,
```

IMO, it just makes it clearer that we intentionally dropped one section.

sam-berning (approved):

ginglis13 (approved):

PR #716, opened 2026-08-12
feat(eif-builder): add signing flags, resign, and describe subcommands

jmt-lab:
Extend the eif-builder CLI to consume the new signing library:

* Top-level build gains --signing-cert + (--signing-key | --kms-key-id)
  + optional --region, gated by a clap ArgGroup. Passing --signing-cert
  alone is an error. Omitting all three produces an unsigned EIF, so
  existing callers keep working.
* New `resign` subcommand: rewrites only the SIGNATURE section (and
  header CRC) of an existing EIF, byte-preserving kernel/cmdline/
  ramdisk. PCR0 is recomputed from the input bytes so the new signature
  covers the same measurement the enclave sees.
* New `describe` subcommand: prints a JSON summary (including the
  decoded METADATA section) to stdout so `eif2eif` can carry
  KernelVersion / BuildToolVersion forward across a repack.

The implicit top-level "build" flow is preserved for backward
compatibility with rpm2eif/eif2eif.

### DIFF
--- a/tools/eif-builder/src/main.rs
+++ b/tools/eif-builder/src/main.rs
@@ -1,9 +1,11 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-use clap::Parser;
+use clap::{Parser, Subcommand};
 use eif_builder::kernel::prepare_kernel;
+use eif_builder::signer::{KmsSigner, LocalSigner};
 use eif_builder::{
-    build_eif, EifError, MetadataFields, PrepareKernelSnafu, ReadKernelSnafu, TargetArch,
+    build_eif, build_signed_eif, describe_eif, resign_eif, EifError, MetadataFields,
+    PrepareKernelSnafu, ReadCertSnafu, ReadKernelSnafu, ReadKeySnafu, SignerBuildSnafu, TargetArch,
     WritePreparedKernelSnafu, DEFAULT_CMDLINE, DEFAULT_PCIE_FLAGS,
 };
 use snafu::{report, ResultExt};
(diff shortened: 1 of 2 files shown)


qianxjcraig (approved):

Sparksssj (approved):

tools/eif-builder/src/main.rs
```diff
@@ -77,6 +81,88 @@ struct Args {
...
+
+    /// PEM-encoded ECDSA private key (P-256 or P-384) used to sign PCR0.
+    /// Mutually exclusive with `--kms-key-id`; requires `--signing-cert`.
+    #[arg(long, requires = "signing_cert", group = "signing_backend")]
+    signing_key: Option<PathBuf>,
+
+    /// AWS KMS key ID or ARN used to sign PCR0. Mutually exclusive with
+    /// `--signing-key`; requires `--signing-cert`. Credentials come from
+    /// the ambient AWS environment; pass `--region` when off-EC2 (the
+    /// buildkit sandbox has no ambient region source).
```
arnaldo2792: nit: the comment within parentheses is extra:

```suggestion
    /// the ambient AWS environment; pass `--region` when off-EC2
```

ginglis13:
```suggestion
    /// `--signing-key`; requires `--signing-cert`. 
```

ginglis13 (approved):

ginglis13 (approved):

tools/eif-builder/src/main.rs
```diff
@@ -101,41 +187,147 @@ fn write_prepared_kernel(
...
+        (None, None, None) => {
+            build_eif(
+                &kernel,
+                &args.cmdline,
+                &output,
+                args.mem,
+                args.cpus,
+                args.pcie_flags,
+                arch,
+                &metadata,
```
arnaldo2792, line 236: Not for this PR, but it would have been nice to create a struct for `build_eif` and reduce some of the verbosity here.

PR #724, opened 2026-08-13
Re-pack support for UKIs

arnaldo2792:
**Description of changes:**

This PR complements: https://github.com/bottlerocket-os/twoliter/pull/701 and closes the gap to support re-pack workflows with UKIs.

A new binary (ukisys) is provided, to extract the `stub` from the compiled UKI. This is needed to offer the same experience like in other re-pack workflows: all the artifacts used to build the final image come from the cracked-open rootfs and UKIs.

**Testing done:**

- Ran the `re-pack` workflow against a local build, using a custom certificate and new signing keys
- Publish the AMIs with the re-packed artifacts
- AMI booted, and cryptographic material matched that of the new keys

**Terms of contribution:**

### DIFF
--- a/tools/ukisys/src/main.rs
+++ b/tools/ukisys/src/main.rs
@@ -1,0 +1,78 @@
+//! ukisys: derives the unsigned systemd-stub PE a Bottlerocket UKI was
+//! built from, by stripping its Authenticode signature and truncating the
+//! trailing payload sections. Those sections are `.osrel`, `.cmdline`,
+//! `.uname`, and `.linux`.
+
+mod error;
+mod pe;
+
+use clap::{Parser, Subcommand};
+use pe::PeImage;
+use snafu::ResultExt;
+use std::path::{Path, PathBuf};
+
+use crate::error::Result;
+
+/// All four sections must be removed, not just the three payload sections:
+/// `ukify build` overwrites any same-named section already in the stub in
+/// place instead of appending fresh content, rather than skipping it.
+const TRAILING_SECTIONS_TO_REMOVE: &[&str] = &[".osrel", ".cmdline", ".uname", ".linux"];
+
+/// Command-line arguments for ukisys.
+#[derive(Parser)]
+#[command(
+    version,
+    about = "PE section removal for Bottlerocket Unified Kernel Images"
+)]
+struct Args {
+    #[command(subcommand)]
+    command: Command,
+}
+
+/// Subcommands for UKI repack section removal.
+#[derive(Subcommand)]
+enum Command {
+    /// Derive an unsigned systemd-stub PE from a finished, signed UKI.
+    DeriveStub {
+        /// Path to the finished, signed UKI to strip.
+        uki: PathBuf,
+
+        /// Path to write the derived, unsigned stub to.
+        stub: PathBuf,
+    },
+}
+
+#[snafu::report]
+fn main() -> Result<()> {
+    let args = Args::parse();
+    match &args.command {
+        Command::DeriveStub { uki, stub } => derive_stub(uki, stub),
+    }
+}
+
+fn derive_stub(uki_path: &Path, stub_path: &Path) -> Result<()> {
+    let mut image = PeImage::load(uki_path)
+        .with_whatever_context(|_| format!("Failed to parse UKI '{}'", uki_path.display()))?;
+
+    image
+        .remove_signature()
+        .with_whatever_context(|_| "Failed to remove signature".to_string())?;
+
+    image
+        .derive_stub_by_truncating_trailing_sections(TRAILING_SECTIONS_TO_REMOVE)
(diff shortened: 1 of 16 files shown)


jmt-lab (approved):

arnaldo2792:
(forced push includes a change to generate the correct HMAC, based on top off: https://github.com/bottlerocket-os/twoliter/pull/701)

arnaldo2792:
(forced push to include the change in https://github.com/bottlerocket-os/twoliter/pull/701, which was needed due to the dependency on the new `fat` file extension).

tools/ukisys/src/pe.rs
```diff
@@ -0,0 +1,577 @@
...
+        buf[offset],
+        buf[offset + 1],
+        buf[offset + 2],
+        buf[offset + 3],
+    ]))
+}
+
+/// Writes a little-endian u16 into `buf` at `offset` in place.
+fn write_u16(buf: &mut [u8], offset: usize, value: u16) {
+    buf[offset..offset + 2].copy_from_slice(&value.to_le_bytes());
```
sam-berning: do we need bounds checks on the writes?

arnaldo2792:
(Forced push to rebase changes merged to develop)

arnaldo2792:
Forced push includes:
- Clean up comments
- Format README to match other markdowns
- Implement bound checks on PE operations

piyush-jena (approved):

tools/ukisys/src/main.rs
```diff
@@ -0,0 +1,80 @@
...
+    image
+        .derive_stub_by_truncating_trailing_sections(TRAILING_SECTIONS_TO_REMOVE)
+        .with_whatever_context(|_| "Failed to derive stub".to_string())?;
+
+    image
+        .write_to(stub_path)
+        .with_whatever_context(|_| format!("Failed to write stub '{}'", stub_path.display()))?;
+
+    eprintln!(
+        "ukisys: removed {} from '{}' (signature + trailing sections {:?}), wrote {} bytes to '{}' (from {size_before} bytes)",
```
sam-berning: Not sure I understand this wording. Should it be:
```suggestion
        "ukisys: removed {} sections from '{}' (signature + trailing sections {:?}), wrote {} bytes to '{}' (from {size_before} bytes)",
```
?

Or should the number be `TRAILING_SECTIONS_TO_REMOVE.len() + 1`?

tools/ukisys/src/pe.rs
```diff
@@ -0,0 +1,707 @@
...
+    }
+
+    /// Parses a PE32+ image already resident in memory.
+    pub fn parse(bytes: Vec<u8>) -> Result<Self, PeError> {
+        if bytes.len() < 0x40 || &bytes[0..2] != b"MZ" {
+            return Err(PeError::BadDosSignature);
+        }
+
+        let pe_offset = read_u32(&bytes, OFF_E_LFANEW, "e_lfanew")? as usize;
+        let pe_offset_end = pe_offset.checked_add(4).ok_or(PeError::TruncatedFile {
```
sam-berning: This error case is inverted. We should introduce a `FileTooLarge` error for when an offset exceeds the max value of `usize`. This case also occurs in all of the `read` and `write` utility functions.

tools/ukisys/src/pe.rs
```diff
@@ -0,0 +1,707 @@
...
+        let coff_offset = pe_offset + 4;
+        let num_sections = read_u16(&bytes, coff_offset + COFF_NUM_SECTIONS, "NumberOfSections")?;
+        let size_of_optional_header = read_u16(
+            &bytes,
+            coff_offset + COFF_SIZE_OPT_HDR,
+            "SizeOfOptionalHeader",
+        )?;
+
+        let opt_offset = pe_offset + 24;
+        let magic = read_u16(&bytes, opt_offset + OPT_MAGIC, "OptionalHeader.Magic")?;
```
sam-berning, line 163: nit: Technically, if we're worried about `usize` overflow, these should be `checked_add`s as well. In practice I'm not sure we need `checked_add` in this function, the length of the Vec is a `usize`, so if the file can be loaded into a `Vec` then there shouldn't be any valid offset that exceeds `usize::MAX`

arnaldo2792:
Last two forced pushes include:

- Address feedback
- Fix `cargo fmt` findings

tools/ukisys/src/pe.rs
arnaldo2792: I questioned the AI about it. Its a harmless/cheap check worth keeping:

> removing this check doesn't simplify anything - you'd just replace checked_add with a plain +, which panics instead of erroring cleanly.

Since we are doing byte math here, I would prefer if we keep it

sam-berning (approved):

PR #734, opened 2026-09-01
Preserve variant identity string in `with_overrides`

### LINKED ISSUE #641: Feature Request: Provide way to specify attributes for my variant rather than rely upon naming

ginglis13:
**Issue number:**

Closes #641 

**Description of changes:**

`with_overrides()` was reconstructing the variant identity string from the overridden attributes (`platform-runtime[-version][-flavor]`), which broke the `bottlerocket-variant(name)` RPM Provides/Requires contract for custom variants that use overrides.

For example, a variant named `"aws-dev-gpu"` with overrides `platform=aws`, `runtime=custom-runtime`, `flavor=gpu` produced the identity `"aws-custom-runtime-gpu-gpu"` instead of `"aws-dev-gpu"`. This caused unresolvable RPM dependencies since settings-defaults Requires `bottlerocket-variant(aws-dev-gpu)` but metadata Provides the reconstructed name instead.

The fix preserves `self.variant` (the original name) as the identity while still allowing overridden attributes to flow through to the individual accessors (`platform()`, `runtime()`, `family()`, `version()`, `variant_flavor()`) which drive the `%_cross_variant_platform/runtime/family/flavor` RPM macros and bconds.

Standard Bottlerocket variants are unaffected since none use overrides.

**Testing done:**

* unit tests addded to cover these scenarios
* manual testing with my branch https://github.com/ginglis13/bottlerocket/tree/test-custom-variant - variants build successfully, custom named `aws-dev-gpu` and `prod-1`, also built `aws-dev` as a sanity check

**Terms of contribution:**

### DIFF
--- a/tools/bottlerocket-variant/src/lib.rs
+++ b/tools/bottlerocket-variant/src/lib.rs
@@ -195,44 +195,78 @@ impl Variant {
     ///
     /// If an override is `Some`, it replaces the original value. If `None`, the original is kept.
     /// Family is recomputed from final platform/runtime UNLESS family was explicitly overridden.
-    pub fn with_overrides(&self, overrides: &VariantOverrides) -> Self {
+    ///
+    /// The variant identity string (i.e. its name as used by `Display`, `AsRef<str>`, etc.) is
+    /// always preserved from the original variant.  Overrides only affect the individual attribute
+    /// accessors (`platform()`, `runtime()`, `family()`, `version()`, `variant_flavor()`).  This
+    /// is important because the identity string is used for the `bottlerocket-variant(name)` RPM
+    /// Provides/Requires contract; rewriting it from overridden attributes breaks that contract.
+    pub fn with_overrides(&self, overrides: &VariantOverrides) -> Result<Self> {
         let platform = overrides
             .platform
             .clone()
             .unwrap_or_else(|| self.platform.clone());
+        ensure!(
+            !platform.is_empty(),
+            error::VariantPartEmptySnafu {
+                part_name: "platform",
+                variant: self.variant.clone()
+            }
+        );
         let runtime = overrides
             .runtime
             .clone()
             .unwrap_or_else(|| self.runtime.clone());
+        ensure!(
+            !runtime.is_empty(),
+            error::VariantPartEmptySnafu {
+                part_name: "runtime",
+                variant: self.variant.clone()
+            }
+        );
         let family = overrides
             .family
             .clone()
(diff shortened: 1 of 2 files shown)


tools/bottlerocket-variant/src/lib.rs
```diff
@@ -195,6 +195,12 @@ impl Variant {
     ///
     /// If an override is `Some`, it replaces the original value. If `None`, the original is kept.
     /// Family is recomputed from final platform/runtime UNLESS family was explicitly overridden.
+    ///
+    /// The variant identity string (i.e. its name as used by `Display`, `AsRef<str>`, etc.) is
+    /// always preserved from the original variant.  Overrides only affect the individual attribute
+    /// accessors (`platform()`, `runtime()`, `family()`, `version()`, `variant_flavor()`).  This
+    /// is important because the identity string is used for the `bottlerocket-variant(name)` RPM
+    /// Provides/Requires contract; rewriting it from overridden attributes breaks that contract.
     pub fn with_overrides(&self, overrides: &VariantOverrides) -> Self {
```
jmt-lab: P2 — No emptiness validation on overridden platform/runtime

  ▎ parse() rejects empty platform/runtime via ensure!(!platform.is_empty(), ...), but with_overrides() has no equivalent check. VariantOverrides { 
  ▎ platform: Some(String::new()), .. } silently produces family == "-dev" and an empty variant_platform cfg. Given VariantOverrides is Deserialized from 
  ▎ Cargo.toml metadata, a typo'd empty override would sail through undetected. Suggest reusing the same ensure! guards.

This finding makes me a bit concerned that we can produce something invalid here that could cause cascading issues? Have we validated if that is the case or not?

ginglis13: Added emptiness checks in latest force push https://github.com/bottlerocket-os/twoliter/compare/adc191b2c8e2ff9a17e18300061269729cd29bd2..7a3c0f213c70e8d45930637221b5862f11aa7e09

jmt-lab (approved):

sam-berning (approved):

